PR #6, opened 2019-11-21
Add DeltaStorageHandler for Hive

windpiger:
Implement HiveOnDelta with StorageHandler
**DDL:**
```
create external table deltaTbl(a string, b int)
stored by 'io.delta.hive.DeltaStorageHandler'
location '/test/delta'
```
- must be external table
- must not be a Hive partition table
- if DeltaTable is a partitionTable, then the partition column should be after data column when creating Hive table
- Hive's schema should be match with the under delta'schema ,including column number &column name
- the delta.table.path should be existed

**Read:**
`set hive.input.format = io.delta.hive.HiveInputFormat`
- support read a non-partition or a partition table
- support push down filter with delta's partition column, currently support predicate (=,!=,>,>=,<,<=,in,like)
- auto-detected delta's partition change

**Unit Tests:**
 - Added(`build/sbt clean test`)
 - `build/sbt clean package` test ok in real Hive Cluster using delta-core-shaded-assembly-0.4.0.jar and hive-delta_2.12-0.4.0.jar


build.sbt
```diff
@@ -133,24 +139,25 @@ lazy val hive = (project in file("hive")) settings (
...
       ExclusionRule(organization = "org.apache.parquet"),
       ExclusionRule("ch.qos.logback", "logback-classic"),
       ExclusionRule("org.pentaho", "pentaho-aggdesigner-algorithm")
     ),
     // TODO Figure out how this fixes some bad dependency
-    "org.apache.spark" %% "spark-core" % "2.4.2" % "test" classifier "tests",
-    "org.scalatest" %% "scalatest" % "3.0.5" % "test"
+    "org.apache.spark" %% "spark-core" % sparkVersion % "test" classifier "tests",
+    "org.scalatest" %% "scalatest" % "3.0.5" % "test",
+    "io.delta" %% "delta-core" % "0.4.0" % "test" excludeAll ExclusionRule("org.apache.hadoop")
```
zsxwing: Why do we need this? The shaded jar has included it.

hive/pom.xml
```diff
@@ -9,21 +9,25 @@
     <version>1.0-SNAPSHOT</version>
 
     <properties>
-        <scala.binary.version>2.11</scala.binary.version>
+        <scala.binary.version>2.12</scala.binary.version>
```
zsxwing: Do we still need this pom.xml? Have you seen any issue when using the jars built by SBT?

hive/src/main/java/io/delta/hive/DeltaInputFormat.java
```diff
@@ -0,0 +1,60 @@
...
+import org.apache.hadoop.io.ArrayWritable;
+import org.apache.hadoop.io.NullWritable;
+import org.apache.hadoop.mapred.*;
+import org.apache.hadoop.mapreduce.security.TokenCache;
+import org.apache.parquet.hadoop.ParquetInputFormat;
+import org.apache.spark.sql.delta.DeltaHelper;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class DeltaInputFormat extends FileInputFormat<NullWritable, ArrayWritable> {
```
zsxwing: Why not just extend `MapredParquetInputFormat`?

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
```diff
@@ -1,38 +1,132 @@
...
+    val convertedFilterExpr = DeltaPushFilter.partitionFilterConverter(filterExprSerialized)
+
+    // selected files to Hive to be processed
+    DeltaLog.filterFileList(
+      snapshotToUse.metadata.partitionColumns, snapshotToUse.allFiles.toDF(), convertedFilterExpr)
+      .as[AddFile](SingleAction.addFileEncoder)
+      .collect()
+      .map { f =>
+        logInfo(s"selected delta file ${f.path} under $rootPath" )
+        fs.getFileStatus(new Path(rootPath, f.path))
```
zsxwing: We can convert `AddFile` to `FileStatus`. The only missing piece is block size, but we can pick up the default one for now. We can add the block size to `AddFile` later.

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
```diff
@@ -1,38 +1,132 @@
...
+import org.apache.hadoop.hive.metastore.api.{FieldSchema, MetaException}
+import org.apache.hadoop.hive.ql.plan.TableScanDesc
+import org.apache.hadoop.mapred.JobConf
+import org.apache.spark.internal.Logging
+import org.apache.spark.sql.SparkSession
+import org.apache.spark.sql.delta.actions.{AddFile, SingleAction}
+
+object DeltaHelper extends Logging {
+  // cache the deltalog instance
+  private var deltaLogMap = scala.collection.mutable.Map[Path, DeltaLog]()
```
zsxwing: `DeltaLog ` has a cache internally. We don't need to cache it twice.

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
```diff
@@ -1,38 +1,132 @@
...
+
+  def parsePathPartition(path: Path, partitionCols: Seq[String]): Map[String, String] = {
```
zsxwing: Ideally we should use the same approach in https://github.com/delta-io/delta/blob/master/src/main/scala/org/apache/spark/sql/delta/files/TahoeFileIndex.scala#L59 to parse the partition values. This can be done later.

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
```diff
@@ -1,38 +1,132 @@
...
-      throw DeltaErrors.pathNotExistsException(partitionFragments.mkString(","))
+  def listDeltaFiles(rootPath: Path, job: JobConf): java.util.List[FileStatus] = {
+    val deltaLog = getDeltaLog(rootPath)
+    // get the snapshot of the version
+    val snapshotToUse = deltaLog.snapshot
+
+    val fs = rootPath.getFileSystem(job)
+
+    // get the partition prune exprs
+    val filterExprSerialized = job.get(TableScanDesc.FILTER_EXPR_CONF_STR)
```
zsxwing: I'm not familiar with Hive. If a query has two Delta tables, e.g., union two tables, will each `InputFormat` have its own `JobConf`, or Hive resets it when calling an `InputFormat`?

hive/src/main/java/io/delta/hive/HiveInputFormat.java
```diff
@@ -0,0 +1,15 @@
+package io.delta.hive;
+
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.mapred.JobConf;
+
+public class HiveInputFormat extends org.apache.hadoop.hive.ql.io.HiveInputFormat {
```
zsxwing: Could you explain why we need this?

hive/src/main/java/io/delta/hive/DeltaStorageHandler.java
```diff
@@ -0,0 +1,202 @@
...
+                throw new MetaException("table location should be set when creating table");
+            } else {
+                Path deltaPath = new Path(deltaRootString);
+                FileSystem fs = deltaPath.getFileSystem(getConf());
+                if (!fs.exists(deltaPath)) {
+                    throw new MetaException("delta.table.path(" + deltaRootString + ") does not exist...");
+                } else {
+                    Map<String, String> partitionProps = JavaConverters.mapAsJavaMap(
+                            DeltaHelper.checkHiveColsInDelta(deltaPath, tbl.getSd().getCols()));
+                    tbl.getSd().getSerdeInfo().getParameters().putAll(partitionProps);
```
zsxwing: Why do we need this? I think we are hiding the partitions from Hive. Right?

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
zsxwing: `new Path(rootPath, f.path)` => `new Path(rootPath, new URI(f.path))`

build.sbt
windpiger: If we don't have this, we will get erro in https://github.com/delta-io/connectors/blob/f29c11ef939d29a42a7a726d790e50df6148c042/hive/src/test/scala/io/delta/hive/HiveConnectorSuite.scala#L9 where `build/sbt clean test`

```
connectors/hive/src/test/scala/io/delta/hive/HiveConnectorSuite.scala:6: object tables is not a member of package io.delta
[error] import io.delta.tables.DeltaTable
[error]                 ^
```

hive/src/main/java/io/delta/hive/DeltaInputFormat.java
windpiger: I want to use https://github.com/apache/hive/blob/f37c5de6c32b9395d1b34fa3c02ed06d1bfbf6eb/ql/src/java/org/apache/hadoop/hive/ql/io/parquet/MapredParquetInputFormat.java#L62 , but it is private.

I think DeltaInputFormat do the similar things with MapredParquetInputFormat, and can repalce it.

hive/src/main/java/io/delta/hive/HiveInputFormat.java
windpiger: Currently HiveOnDelta implement by StorageHandler, but actually it is read by the way of Parquet.
So some logical should be the same with parquet tbl, or we will get some problem.

Here nonNative is decided by https://github.com/apache/hive/blob/827f4f9c01dab706f247fb3d1a9099a405c67b2e/ql/src/java/org/apache/hadoop/hive/ql/plan/TableDesc.java#L185, the HiveOnDelta is `nonNative=true`(while parquet tbl is false), then it can not push down projections(https://github.com/apache/hive/blob/c104e8bfab957702dc012a8b4ac6a944669f8a3a/ql/src/java/org/apache/hadoop/hive/ql/io/HiveInputFormat.java#L937), then NULL column values will be read from parquet(because the read schema is like _col1 _col2, not name cnt)

correct:
```
INFO [main] org.apache.hadoop.hive.ql.io.parquet.read.DataWritableReadSupport: message spark_schema {
  optional binary name (UTF8);
  optional binary cnt (UTF8);
}
```
wrong:
```
INFO [main] org.apache.hadoop.hive.ql.io.parquet.read.DataWritableReadSupport: message spark_schema {
  optional binary _col1 (UTF8);
  optional binary _col2 (UTF8);
}```

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
windpiger: here is to resolve partitionvalues from file path, like `/xxxx/a=1/b=2` -> `Map(a->1, b ->2)`.
It does not the same approach in TahoeFileIndex.scala.

I copied this from https://github.com/delta-io/delta/blob/08c6e75b84b1dc18d663c6cc97118fd34298f3c8/src/main/scala/org/apache/spark/sql/delta/util/PartitionUtils.scala#L161

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
windpiger: unset & reset it

https://github.com/apache/hive/blob/c104e8bfab957702dc012a8b4ac6a944669f8a3a/ql/src/java/org/apache/hadoop/hive/ql/io/HiveInputFormat.java#L816

I test two different tables union, both the two can do partitionprune correctly

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
windpiger: > `new Path(rootPath, f.path)` => `new Path(rootPath, new URI(f.path))`

there is no this constructor function with URI?

windpiger: > We can convert `AddFile` to `FileStatus`. The only missing piece is block size, but we can pick up the default one for now. We can add the block size to `AddFile` later.

file.getLen() & file.getBlockSize() is used in doing the splits later.
I think it shoube the right value of blocksize(default one is not properly), maybe we can refactor this after
we already finished add blocksize to AddFile?

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
windpiger: 屏幕快照 2019-11-21 下午10 42 45

push predicate to StorageHandler logical based on HBase

PR #7, opened 2019-11-26
Rewrite Java files using Scala

zsxwing:
This PR rewrites Java files using Scala to make future development easier since Delta itself is written using Scala. `IndexPredicateAnalyzer` is not rewritten because it's a fork from Hive's `IndexPredicateAnalyzer` and it's better to not change it so that we can compare them when `IndexPredicateAnalyzer` is changed in Hive.

This PR doesn't change any logic in codes.


windpiger:
I have test it:
1. `build/sbt clean test`
2. `build/sbt clean package` to test it in real Hive Cluster

LGTM

zsxwing:
Thanks! Merging to master.

PR #8, opened 2019-11-27
minor modify README.md

windpiger:
Make some minor change on READ.md

No need to test.


zsxwing:
Thanks! Merging this. The failure is because the test somehow exceeded the memory limit of circleci. I will check how we can reduce the memory usage.

PR #9, opened 2019-11-27
Improve error messages for unsupported features in Hive

zsxwing:
This PR makes some minor improvements in the error messages for unsupported features. It also adds the table property `spark.sql.sources.provider` so that a Delta table created by Hive can be read by Spark 3.0.0+ when they share the same metastore.

### DIFF
--- a/hive/src/main/scala/io/delta/hive/DeltaStorageHandler.scala
+++ b/hive/src/main/scala/io/delta/hive/DeltaStorageHandler.scala
@@ -1,10 +1,7 @@
 package io.delta.hive
 
-import com.google.common.base.Joiner
 import scala.collection.JavaConverters._
 import scala.collection.mutable
-
-import org.apache.hadoop.fs.FileSystem
 import org.apache.hadoop.fs.Path
 import org.apache.hadoop.hive.metastore.HiveMetaHook
 import org.apache.hadoop.hive.metastore.MetaStoreUtils
@@ -22,9 +19,7 @@ import org.apache.hadoop.hive.ql.plan.TableDesc
 import org.apache.hadoop.hive.serde2.AbstractSerDe
 import org.apache.hadoop.hive.serde2.Deserializer
 import org.apache.hadoop.hive.serde2.typeinfo.TypeInfoFactory
-import org.apache.hadoop.mapred.InputFormat
-import org.apache.hadoop.mapred.JobConf
-import org.apache.hadoop.util.StringUtils
+import org.apache.hadoop.mapred.{InputFormat, JobConf, OutputFormat}
 import org.apache.spark.sql.delta.DeltaHelper
 import org.apache.spark.sql.delta.DeltaPushFilter
 import org.slf4j.LoggerFactory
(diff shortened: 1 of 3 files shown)


hive/src/main/scala/io/delta/hive/DeltaStorageHandler.scala
```diff
@@ -127,13 +128,19 @@ class DeltaStorageHandler extends DefaultStorageHandler with HiveMetaHook with H
...
-    } else if (tbl.getPartitionKeysSize() > 0) {
-      throw new MetaException("HiveOnDelta does not support to create a partition hive table")
+    if (!MetaStoreUtils.isExternalTable(tbl)) {
+      throw new UnsupportedOperationException(
+        s"The type of table ${tbl.getDbName}:${tbl.getTableName} is ${tbl.getTableType}." +
+          "Only external Delta tables can be read in Hive right now")
     }
 
+    if (tbl.getPartitionKeysSize > 0) {
+      throw new MetaException(
```
zsxwing: This is a `MetaException` because it will never be supported.

hive/src/test/scala/io/delta/hive/HiveConnectorSuite.scala
```diff
@@ -19,23 +19,18 @@ class HiveConnectorSuite extends HiveTest with BeforeAndAfterEach {
...
+        s"""
+           |create table deltaTbl(a string, b int)
+           |stored by 'io.delta.hive.DeltaStorageHandler'""".stripMargin
+      )
     }
+    assert(e.getMessage != null && e.getMessage.contains("Only external Delta tables"))
   }
 
-  test("DDL: location should be set when creating table") {
+  test("location should be set when creating table") {
```
zsxwing: I removed `DDL: ` from the tests since it's too verbose.

hive/src/main/scala/io/delta/hive/DeltaStorageHandler.scala
```diff
@@ -146,6 +153,7 @@ class DeltaStorageHandler extends DefaultStorageHandler with HiveMetaHook with H
         val partitionProps = DeltaHelper.checkHiveColsInDelta(deltaPath, tbl.getSd().getCols())
         tbl.getSd().getSerdeInfo().getParameters().putAll(partitionProps.asJava)
         tbl.getSd().getSerdeInfo().getParameters().put(DELTA_TABLE_PATH, deltaRootString)
+        tbl.getParameters.put("spark.sql.sources.provider", "DELTA")
```
windpiger: we should also add  path property for spark-sql to read HiveOnDelta tbl as Delta DataSource

tbl.getSd().getSerdeInfo().getParameters().put("path", deltaRootString)

I have test it is ok.

windpiger: maybe we also Add a test case for spark read HiveOnDelta tbl as a Delta DataSource

zsxwing: > tbl.getSd().getSerdeInfo().getParameters().put("path", deltaRootString)

This is not necessary. Spark uses `tbl.getSd().getLocation()` which has been set. See the following codes:

https://github.com/apache/spark/blob/e99a9f78ea2e18eabca722c5714280eb7a737499/sql/hive/src/main/scala/org/apache/spark/sql/hive/client/HiveClientImpl.scala#L523
https://github.com/apache/hive/blob/fafa46a3cfeb4c95efcf7ea86dd227965163587b/ql/src/java/org/apache/hadoop/hive/ql/metadata/Table.java#L279
https://github.com/apache/hive/blob/fafa46a3cfeb4c95efcf7ea86dd227965163587b/ql/src/java/org/apache/hadoop/hive/ql/metadata/Table.java#L267

> maybe we also Add a test case for spark read HiveOnDelta tbl as a Delta DataSource

Spark 2.4.x doesn't support this. We need to wait for Spark 3.0.0 in order to write a test.

windpiger: ok, LGTM thanks~

zsxwing:
Thanks! Merging to master.

PR #10, opened 2019-12-03
Convert AddFile to FileStatus directly to save FileSystem RPC calls

zsxwing:
This PR creates `FileStatus` directly from `AddFile` to save FileSystem RPC calls. As `AddFile` doesn't have the block locations, we lose the locality. But this is fine today since most of storage systems are on Cloud and the computation is running separately.

I also fixes a bug that we return incorrect file paths to Hive when the partition values have some special values that are escaped in the file path.

### DIFF
--- a/hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
+++ b/hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
@@ -69,9 +69,12 @@ object DeltaHelper extends Logging {
 
   def listDeltaFiles(nonNormalizedPath: Path, job: JobConf): Array[FileStatus] = {
     val fs = nonNormalizedPath.getFileSystem(job)
-    // TODO It seems Hive doesn't work without "makeQualified". We should understand this deeply to
-    // make sure we call "makeQualified" in all places properly, because missing "makeQualified"
-    // returns an incorrect answer rather than failing the query, which is pretty bad.
+    // We need to normalize the table path so that all paths we return to Hive will be normalized
+    // This is necessary because `HiveInputFormat.pushProjectionsAndFilters` will try to figure out
+    // which table a split path belongs to by comparing the split path with the normalized (? I have
+    // not yet confirmed this) table paths.
+    // TODO The assumption about Path in Hive is too strong, we should try to see if we can fail if
+    // `pushProjectionsAndFilters` doesn't find a table for a Delta split path.
     val rootPath = fs.makeQualified(nonNormalizedPath)
     val deltaLog = DeltaLog.forTable(spark, rootPath)
     // get the snapshot of the version
--- a/hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
+++ b/hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
@@ -1,10 +1,12 @@
 package org.apache.spark.sql.delta
 
-import io.delta.hive.DeltaStorageHandler
+import java.net.URI
+
 import scala.collection.JavaConverters._
 import scala.collection.mutable.ArrayBuffer
 
-import org.apache.hadoop.fs.{FileStatus, Path}
+import io.delta.hive.DeltaStorageHandler
+import org.apache.hadoop.fs._
 import org.apache.hadoop.hive.metastore.api.{FieldSchema, MetaException}
 import org.apache.hadoop.hive.ql.plan.TableScanDesc
 import org.apache.hadoop.mapred.JobConf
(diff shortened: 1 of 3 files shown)


hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
```diff
@@ -65,26 +67,84 @@ object DeltaHelper extends Logging {
     }
   }
 
-  def listDeltaFiles(rootPath: Path, job: JobConf): java.util.List[FileStatus] = {
+  def listDeltaFiles(nonNormalizedPath: Path, job: JobConf): Array[FileStatus] = {
+    val fs = nonNormalizedPath.getFileSystem(job)
+    // TODO It seems Hive doesn't work without "makeQualified". We should understand this deeply to
```
zsxwing: @windpiger do you know about this requirement in Hive? I took a quick look but didn't find anything. But if removing `makeQualified`, the query result would become wrong.

hive/src/test/scala/io/delta/hive/HiveConnectorSuite.scala
```diff
@@ -408,4 +408,29 @@ class HiveConnectorSuite extends HiveTest with BeforeAndAfterEach {
...
+        }
+
+        runQuery(
+          s"""
+             |create external table deltaPartitionTbl(c1 int, c2 string)
+             |stored by 'io.delta.hive.DeltaStorageHandler' location '${dir.getCanonicalPath}'
+         """.stripMargin
+        )
+
+        // TODO Read partition values from `AddFile.partitionValues` to fix incorrect escaped
```
zsxwing: @windpiger since you are working on this, I don't fix this in my PR. This test right now is just verifying that we return correct file paths. Otherwise, it will fail because of `FileNotFoundException`.

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
windpiger: I dig it out that we should makeQualified, because the pushdown 
projection(https://github.com/apache/hive/blob/c104e8bfab957702dc012a8b4ac6a944669f8a3a/ql/src/java/org/apache/hadoop/hive/ql/io/HiveInputFormat.java#L919) will be missed(https://github.com/apache/hive/blob/c104e8bfab957702dc012a8b4ac6a944669f8a3a/common/src/java/org/apache/hadoop/hive/common/FileUtils.java#L968) if we do not do this (file:/xxx != /xxx)

Before this pr, it is ok because liststatus rpc request return normalized path, this pr change to use AddFile ,so we should do it by ourself. Or we can also put this logic to `def absolutePath`

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
```diff
@@ -65,26 +67,84 @@ object DeltaHelper extends Logging {
...
+    // "spark.sql.files.maxPartitionBytes" in Spark. It's also the default parquet row group size
+    // which is usually the best split size for parquet files.
+    val blockSize = job.getLong("parquet.block.size", 128L * 1024 * 1024)
+
     // selected files to Hive to be processed
     DeltaLog.filterFileList(
       snapshotToUse.metadata.partitionColumns, snapshotToUse.allFiles.toDF(), convertedFilterExpr)
       .as[AddFile](SingleAction.addFileEncoder)
-      .collect().par.map { f =>
+      .collect().map { f =>
```
windpiger: par is not ok?

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
```diff
@@ -65,26 +67,84 @@ object DeltaHelper extends Logging {
...
+   *
+   * @param child an escaped string read from Delta's [[AddFile]] directly which requires to
+   *              unescape before creating the [[Path]] object.
+   */
+  private def absolutePath(fs: FileSystem, root: Path, child: String): Path = {
+    val p = new Path(new URI(child))
+    if (p.isAbsolute) {
+      fs.makeQualified(p)
+    } else {
+      new Path(root, p)
```
windpiger: it is better to makeQualified in this function?

hive/src/test/scala/io/delta/hive/HiveConnectorSuite.scala
windpiger: ok ,I will fix this soon

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
zsxwing: The step should be fast now. `par` is not necessary.

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
zsxwing: I'm calling `makeQualified` outside so that we don't need to do it for every path.

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
zsxwing: Thanks for digging into the root cause. It seems Hive tried to avoid this by adding `splitPathWithNoSchema` but they forget to do the same thing for `key`.

zsxwing: I updated the comment to describe my current understanding.

windpiger:
LGTM

zsxwing:
Thanks! Merging to master.

PR #11, opened 2019-12-09
Consistent schema in Hive and Delta metadata

zsxwing:
Right now we require the partition columns should be after the data columns.

This PR adds a new DeltaInputSplit to remove the above limitation and also adds validation to ensure Hive's schema is always consistent with Delta's metadata regarding column types and order.

### DIFF
--- a/hive/src/main/scala/io/delta/hive/DeltaInputFormat.scala
+++ b/hive/src/main/scala/io/delta/hive/DeltaInputFormat.scala
@@ -6,15 +6,35 @@ import org.apache.hadoop.fs.FileStatus
 import org.apache.hadoop.fs.Path
 import org.apache.hadoop.hive.ql.exec.Utilities
 import org.apache.hadoop.hive.ql.io.parquet.read.DataWritableReadSupport
-import org.apache.hadoop.hive.serde.serdeConstants
-import org.apache.hadoop.io.ArrayWritable
-import org.apache.hadoop.io.NullWritable
+import org.apache.hadoop.io.{ArrayWritable, NullWritable, Writable}
 import org.apache.hadoop.mapred._
 import org.apache.hadoop.mapreduce.security.TokenCache
 import org.apache.parquet.hadoop.ParquetInputFormat
 import org.apache.spark.sql.delta.DeltaHelper
 import org.slf4j.LoggerFactory
 
+/**
+ * A special [[InputFormat]] to wrap [[ParquetInputFormat]] to read a Delta table.
+ *
+ * The underlying files in a Delta table are in Parquet format. However, we cannot use the existing
+ * [[ParquetInputFormat]] to read them directly because they only store data for data columns.
+ * The values of partition columns are in Delta's metadata. Hence, we need to read them from Delta's
+ * metadata and re-assemble rows to include partition values and data values from the raw Parquet
+ * files.
+ *
+ * Note: We cannot use the file name to infer partition values because Delta Transaction Log
+ * Protocol requires "Actual partition values for a file must be read from the transaction log".
+ *
+ * In the current implementation, when listing files, we also read the partition values and put them
+ * into an `Array[PartitionColumnInfo]`. Then create a temp `Map` to store the mapping from the file
+ * path to `PartitionColumnInfo`s. When creating an [[InputSplit]], we will create a special
+ * [[FileSplit]] called [[DeltaInputSplit]] to carry over `PartitionColumnInfo`s.
+ *
+ * For each reader created from a [[DeltaInputSplit]], we can get all partition column types, the
+ * locations of a partition column in the schema, and their string values. The reader can build
+ * [[Writable]] for all partition values, and insert them to the raw row returned by
+ * [[org.apache.parquet.hadoop.ParquetRecordReader]].
+ */
 class DeltaInputFormat(realInput: ParquetInputFormat[ArrayWritable]) extends FileInputFormat[NullWritable, ArrayWritable] {
 
   private val LOG = LoggerFactory.getLogger(classOf[DeltaInputFormat])
--- a/hive/src/main/scala/io/delta/hive/DeltaInputFormat.scala
+++ b/hive/src/main/scala/io/delta/hive/DeltaInputFormat.scala
@@ -1,9 +1,11 @@
 package io.delta.hive
 
+import java.io.IOException
 import java.net.URI
 
 import org.apache.hadoop.fs.FileStatus
 import org.apache.hadoop.fs.Path
+import org.apache.hadoop.hive.metastore.api.MetaException
 import org.apache.hadoop.hive.ql.exec.Utilities
 import org.apache.hadoop.hive.ql.io.parquet.read.DataWritableReadSupport
 import org.apache.hadoop.io.{ArrayWritable, NullWritable, Writable}
@@ -63,10 +65,20 @@ class DeltaInputFormat(realInput: ParquetInputFormat[ArrayWritable]) extends Fil
     }
   }
 
+  @throws(classOf[IOException])
   override def listStatus(job: JobConf): Array[FileStatus] = {
     val deltaRootPath = new Path(job.get(DeltaStorageHandler.DELTA_TABLE_PATH))
     TokenCache.obtainTokensForNamenodes(job.getCredentials(), Array(deltaRootPath), job)
-    val (files, partitions) = DeltaHelper.listDeltaFiles(deltaRootPath, job)
+    val (files, partitions) =
+      try {
+        DeltaHelper.listDeltaFiles(deltaRootPath, job)
+      } catch {
+        // Hive is using Java Reflection to call `listStatus`. Because `listStatus` doesn't declare
+        // `MetaException`, the Reflection API would throw `UndeclaredThrowableException` without an
+        // error message if `MetaException` was thrown directly. To improve the user experience, we
+        // wrap `MetaException` with `IOException` which will provide a better error message.
+        case e: MetaException => throw new IOException(e)
+      }
     fileToPartition = partitions.filter(_._2.nonEmpty)
     files
   }
--- a/hive/src/main/scala/io/delta/hive/DeltaInputFormat.scala
+++ b/hive/src/main/scala/io/delta/hive/DeltaInputFormat.scala
@@ -1,11 +1,12 @@
 package io.delta.hive
 
-import java.io.IOException
+import java.net.URI
 
 import org.apache.hadoop.fs.FileStatus
 import org.apache.hadoop.fs.Path
 import org.apache.hadoop.hive.ql.exec.Utilities
 import org.apache.hadoop.hive.ql.io.parquet.read.DataWritableReadSupport
+import org.apache.hadoop.hive.serde.serdeConstants
 import org.apache.hadoop.io.ArrayWritable
 import org.apache.hadoop.io.NullWritable
 import org.apache.hadoop.mapred._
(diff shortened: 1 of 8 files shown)


hive/src/test/scala/io/delta/hive/HiveConnectorSuite.scala
```diff
@@ -181,9 +169,7 @@ class HiveConnectorSuite extends HiveTest with BeforeAndAfterEach {
          """.stripMargin
         )
 
-        assert(runQuery(
-          "select * from deltaNonPartitionTbl").sorted ===
-          testData.map(r => s"${r._1}\t${r._2}").sorted)
+        checkAnswer("select * from deltaNonPartitionTbl", testData)
```
zsxwing: I added this new method `checkAnswer` to simplify the codes.

hive/src/main/scala/io/delta/hive/DeltaStorageHandler.scala
```diff
@@ -141,48 +139,68 @@ class DeltaStorageHandler extends DefaultStorageHandler with HiveMetaHook with H
...
+    val cols = tbl.getSd.getCols
+    val columnNames = new java.util.ArrayList[String](cols.size)
+    val columnTypes = new java.util.ArrayList[TypeInfo](cols.size)
+    cols.asScala.foreach { col =>
+      columnNames.add(col.getName)
+      columnTypes.add(TypeInfoUtils.getTypeInfoFromTypeString(col.getType))
     }
+    val hiveSchema = TypeInfoFactory.getStructTypeInfo(columnNames, columnTypes)
+      .asInstanceOf[StructTypeInfo]
+    DeltaHelper.checkTableSchema(snapshot.metadata.schema, hiveSchema)
```
windpiger: we check Hive tbl schema with delta's schema, including the schema column order. This can guarantee the `insertPartitionValues` function in `DeltaRecordReaderWrapper` correctly, which push partition values to the correct position index in `valueArray`.  This is important logic, I think we should add test, which create a hive table with different column order with delta's schema

hive/src/main/scala/io/delta/hive/DeltaRecordReaderWrapper.scala
```diff
@@ -1,81 +1,61 @@
...
+   * `next` is called. We should insert partition values manually for each row.
+   */
+  private def insertPartitionValues(value: ArrayWritable): Unit = {
+    val valueArray = value.get()
+    var i = 0
+    val n = partitionValues.length
+    // Using while loop for better performance since this method is called for each row.
+    while (i < n) {
+      val partition = partitionValues(i)
+      valueArray(partition._1) = partition._2
```
windpiger: It is better to add comment here to explain how to guarantee the position in valueArray correctly.  such as by the checkTableSchema in preCreate

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
```diff
@@ -77,29 +87,49 @@ object DeltaHelper extends Logging {
...
-    // TODO Verify the table schema is consistent with `snapshotToUse.metadata`.
+    val hiveSchema = TypeInfoUtils.getTypeInfoFromTypeString(
+      job.get(DeltaStorageHandler.DELTA_TABLE_SCHEMA)).asInstanceOf[StructTypeInfo]
+    DeltaHelper.checkTableSchema(snapshotToUse.metadata.schema, hiveSchema)
 
     // get the partition prune exprs
     val filterExprSerialized = job.get(TableScanDesc.FILTER_EXPR_CONF_STR)
 
+    // TODO Check whether valid types of a partition column in Spark can still be used in Hive.
+    // TODO Verify the partition columns are not changed so that the pushed filters are still
```
windpiger: The pushed filters is generated by https://github.com/delta-io/connectors/blob/master/hive/src/main/scala/io/delta/hive/DeltaStorageHandler.scala#L55
`val partitionColumns = DeltaHelper.getPartitionCols(new Path(deltaRootPath))`
I think it is no need to verify the partition columns

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
```diff
@@ -77,29 +87,62 @@ object DeltaHelper extends Logging {
     // `pushProjectionsAndFilters` doesn't find a table for a Delta split path.
     val rootPath = fs.makeQualified(nonNormalizedPath)
     val deltaLog = DeltaLog.forTable(spark, rootPath)
-    // get the snapshot of the version
     val snapshotToUse = deltaLog.snapshot
```
windpiger: should we add deltaLog.update here to get the latest snapshot?

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
windpiger: https://github.com/delta-io/connectors/issues/12 this issuse related

hive/src/main/scala/io/delta/hive/DeltaStorageHandler.scala
```diff
@@ -141,48 +139,68 @@ class DeltaStorageHandler extends DefaultStorageHandler with HiveMetaHook with H
...
+    }
+
+    val snapshot = DeltaHelper.loadDeltaLog(new Path(deltaRootString)).snapshot
+    if (snapshot.version < 0) {
+      throw new MetaException(s"$deltaRootString does not exist or it's not a Delta table")
+    }
+
+    // Extract the table schema in Hive and put it into the table property. Then we can compare it
+    // with the latest table schema in Delta logs and fail the query if it was changed.
+    // TODO Investigate if we can get the table schema without manually storing it in the table
```
windpiger: alter table maybe change the schema, DELTA_TABLE_SCHEMA properties will be wrong. releate issuse https://github.com/delta-io/connectors/issues/14

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
zsxwing: Good catch. We do need to call `update`.

hive/src/main/scala/io/delta/hive/DeltaStorageHandler.scala
zsxwing: I checked `ALTER TABLE` and found out Hive doesn't allow it on a non-native table:
```
org.apache.hadoop.hive.ql.parse.SemanticException: ALTER TABLE cannot be used for a non-native table deltatbl
        at org.apache.hadoop.hive.ql.parse.DDLSemanticAnalyzer.validateAlterTableType(DDLSemanticAnalyzer.java:1354)
        at org.apache.hadoop.hive.ql.parse.DDLSemanticAnalyzer.addInputsOutputsAlterTable(DDLSemanticAnalyzer.java:1533)
        at org.apache.hadoop.hive.ql.parse.DDLSemanticAnalyzer.addInputsOutputsAlterTable(DDLSemanticAnalyzer.java:1479)
        at org.apache.hadoop.hive.ql.parse.DDLSemanticAnalyzer.analyzeAlterTableRenameCol(DDLSemanticAnalyzer.java:2661)
        at org.apache.hadoop.hive.ql.parse.DDLSemanticAnalyzer.analyzeInternal(DDLSemanticAnalyzer.java:283)
        at org.apache.hadoop.hive.ql.parse.BaseSemanticAnalyzer.analyze(BaseSemanticAnalyzer.java:258)
        at org.apache.hadoop.hive.ql.Driver.compile(Driver.java:512)
        at org.apache.hadoop.hive.ql.Driver.compileInternal(Driver.java:1317)
        at org.apache.hadoop.hive.ql.Driver.runInternal(Driver.java:1457)
        at org.apache.hadoop.hive.ql.Driver.run(Driver.java:1237)
        at org.apache.hadoop.hive.ql.Driver.run(Driver.java:1227)
```

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
zsxwing: > I think it is no need to verify the partition columns

Yep, it's no need to verify because I forgot to call `update`. After calling `update()`, these two places may see different `Snapshot`s, so we need to check the partition columns. I added a commit to do a simple check: make sure all filters are based on the partition columns we get from the latest `Snapshot`. This should be enough.

zsxwing:
@windpiger I have addressed all of your comments. PTAL.

hive/src/main/scala/io/delta/hive/DeltaStorageHandler.scala
zsxwing: I think we should investigate how a user can change a schema for a table using storage handler, and make sure we can persist the Hive schema correctly. Hive itself has stored the schema in metastore, but I cannot find any API to get this schema. If we can get it directly from Hive without persisting it by ourselves, it would be better.

hive/src/main/scala/io/delta/hive/DeltaRecordReaderWrapper.scala
```diff
@@ -1,81 +1,70 @@
...
-          + ", partitionWritable length:" + partitionWritable.length)
-        partitionWritable
-      }
+  /** The positions of partition columns in Delta schema and their corresponding values. */
+  private val partitionValues: Array[(Int, Writable)] =
+    split.getPartitionColumns.map { partition =>
+      val oi = PrimitiveObjectInspectorFactory
+        .getPrimitiveWritableObjectInspector(TypeInfoFactory
+          .getPrimitiveTypeInfo(partition.tpe))
+      partition.index -> ObjectInspectorConverters.getConverter(
```
brkyvz: for readability can you declare the second tuple as a variable as well?

hive/src/main/scala/io/delta/hive/DeltaRecordReaderWrapper.scala
```diff
@@ -1,81 +1,70 @@
...
     val hasNext = super.next(key, value)
-    if (partitionWritable != null && partitionWritable.length != 0) {
-      // append partition values to data values
-      for (i <- 0 until partitionWritable.length) {
-        value.get()(value.get().length - partitionWritable.length + i) = partitionWritable(i)
-      }
-    }
+    // TODO Figure out when the parent reader resets partition columns to null so that we may come
+    // out a better solution to not insert partition values for each row.
+    insertPartitionValues(value)
```
brkyvz: do you not need a `if (hasNext) insertPartitionValues(value)`?

hive/src/main/scala/io/delta/hive/DeltaStorageHandler.scala
```diff
@@ -141,48 +139,68 @@ class DeltaStorageHandler extends DefaultStorageHandler with HiveMetaHook with H
...
+   */
   val DELTA_TABLE_PATH = "delta.table.path"
-  val DELTA_PARTITION_COLS_NAMES = "delta.partition.columns"
-  val DELTA_PARTITION_COLS_TYPES = "delta.partition.columns.types"
+
+  /**
+   * A config we use to remember the table schema in the table properties.
+   *
+   * TODO Maybe Hive can tell us this in the `configureInputJobProperties` method. Then we don't
+   * * need to store this extra information.
```
brkyvz: nit: second `*`

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
```diff
@@ -150,35 +191,102 @@ object DeltaHelper extends Logging {
...
-      throw new MetaException(s"The partition cols of Delta should be after data cols " +
-        s"when creating hive table. Delta dataschema is " +
-        s"${deltaMeta.dataSchema.json} and partitionschema is ${deltaMeta.partitionSchema.json}")
+      false
     }
   }
 
-  def getPartitionCols(rootPath: Path): Seq[String] = {
-    DeltaLog.forTable(spark, rootPath).snapshot.metadata.partitionColumns
+  private def isSameType(sparkType: DataType, hiveType: TypeInfo): Boolean = {
```
brkyvz: Is this copied from somewhere?

hive/src/test/scala/io/delta/hive/HiveConnectorSuite.scala
```diff
@@ -82,83 +82,134 @@ class HiveConnectorSuite extends HiveTest with BeforeAndAfterEach {
...
+        }
+
+        // Should detect the underlying schema change and fail the query
+        val e = intercept[Exception] {
+          runQuery("SELECT * FROM deltaTbl")
+        }
+        assert(e.getMessage.contains(s"schema is not the same"))
+
+        // Re-create the table because Hive doesn't allow `ALTER TABLE` on a non-native table.
+        // TODO Investigate whether there is a more convenient way to update the table schema.
```
brkyvz: MSCK REPAIR TABLE?

hive/src/main/scala/io/delta/hive/DeltaRecordReaderWrapper.scala
windpiger: `super.next(key, value)` has been fill data in value, and `hasNext` is used  as a return value to parent function caller, and the parent will use it to determine it is need to call this `next` again. I think it is no need to do this.

hive/src/main/scala/io/delta/hive/DeltaStorageHandler.scala
windpiger: yes, I can do some investigate about this

windpiger:
> @windpiger I have addressed all of your comments. PTAL.

LGTM except the alter table related issuse. Greate Job!

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
zsxwing: No. This is a total new codes. There is a test to cover all types here and a TODO to support more types.

hive/src/main/scala/io/delta/hive/DeltaInputFormat.scala
```diff
@@ -1,44 +1,109 @@
...
+            "Reading Delta tables using Parquet's VectorizedReader is not supported")
+        } else {
+          new DeltaRecordReaderWrapper(this.realInput, deltaSplit, job, reporter)
+        }
+      case _ =>
+        throw new IllegalArgumentException("Expected DeltaInputSplit but it was: " + split)
     }
   }
 
+  @throws(classOf[IOException])
```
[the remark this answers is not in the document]
zsxwing: Yep :) This is Scala syntax.

hive/src/main/scala/io/delta/hive/DeltaRecordReaderWrapper.scala
zsxwing: No need to do this. But I agreed that it's better to add a check to make the codes clearer.

hive/src/test/scala/io/delta/hive/HiveConnectorSuite.scala
zsxwing: `MSCK REPAIR TABLE` == `Recover Partitions` in Hive :( https://cwiki.apache.org/confluence/display/Hive/LanguageManual+DDL#LanguageManualDDL-RecoverPartitions(MSCKREPAIRTABLE)

zsxwing:
Thanks! I'm going to merge this PR. The `Alter Table` issue is tracked in #14.

PR #16, opened 2020-01-14
Set 'path' in the storage properties

zsxwing:
Spark reads the `path` option from the storage properties. We should set it when creating a table in Hive so that the table can be read by Spark as well.

I also removed `delta.table.path` from the table properties since we can get it from the `location` property.


zsxwing:
@windpiger I'm merging this PR. Feel free to take a look and do a post-hoc review.

PR #22, opened 2020-03-05
Cross build Scala 2.11 and 2.12

### LINKED ISSUE #18: Support both Scala 2.11 and 2.12
We run Zeppelin 0.8.2 in our infrastructure and the Spark interpreter in that version doesn't support Scala 2.12, there is a fix in place that won't be released until Zeppelin 0.9.0 (https://jira.apache.org/jira/browse/ZEPPELIN-3552)

To support Zeppelin we build out everything, including spark from source, to use Scala 2.11

Currently there is no documented way to build the Delta connectors using Scala 2.11 outside of potentially modifying the variable set here: https://github.com/delta-io/connectors/blob/master/build.sbt#L31

zsxwing:
Add build support for Scala 2.11.

Closes #18

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,4 +8,5 @@ jobs:
       - run:
           name: Run tests 
           command: |
-            build/sbt coreTest/test hive/test
+            build/sbt "+ coreTest/test"
+            build/sbt "+ hive/test"
--- a/README.md
+++ b/README.md
@@ -29,11 +29,18 @@ This project contains all the code needed to make Hive read Delta Lake tables.
 The above commands will generate two JARs in the following paths.
 
 ```
-core/target/scala-2.12/delta-core-shaded-assembly-0.5.0.jar
+core/target/scala-2.12/delta-core-shaded-assembly_2.12-0.5.0.jar
 hive/target/scala-2.12/hive-delta_2.12-0.5.0.jar
 ```
 
 These two JARs include the Hive connector and all its dependencies. They need to be put in Hive’s classpath.
+
+Note: if you would like to build jars using Scala 2.11, you can run the SBT command `build/sbt "++ 2.11.12 hive/package"` and the generated JARS will be in the following paths.
+
+```
+core/target/scala-2.11/delta-core-shaded-assembly_2.11-0.5.0.jar
+hive/target/scala-2.12/hive-delta_2.11-0.5.0.jar
+```
 
 ### Setting up Hive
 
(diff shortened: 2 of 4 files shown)


.circleci/config.yml
```diff
@@ -8,4 +8,5 @@ jobs:
       - run:
           name: Run tests 
           command: |
-            build/sbt coreTest/test hive/test
+            build/sbt "+ coreTest/test"
```
zsxwing: Tried but didn't find any way to merge these two commands into one SBT command :(

README.md
```diff
@@ -48,7 +55,7 @@ Before starting your Hive CLI or running your Hive script, add the following spe
 </property>
 ```
 
-The second step is to upload the above two JARs to the machine that runs Hive. Finally, add the paths of the JARs toHive’s environment variable, `HIVE_AUX_JARS_PATH`. You can find this environment variable in the `hive-env.sh` file, whose location is ``/etc/hive/conf/hive-env.sh` on an EMR cluster. This setting will tell Hive where to find the connector JARs.
+The second step is to upload the above two JARs to the machine that runs Hive. Finally, add the paths of the JARs toHive’s environment variable, `HIVE_AUX_JARS_PATH`. You can find this environment variable in the `hive-env.sh` file, whose location is `/etc/hive/conf/hive-env.sh` on an EMR cluster. This setting will tell Hive where to find the connector JARs.
```
zsxwing: Remove an extra `` ` ``

build.sbt
```diff
@@ -92,8 +94,13 @@ lazy val assemblySettings = Seq(
     case "reference.conf" => MergeStrategy.concat
     case _ => MergeStrategy.first
   },
+  assemblyJarName in assembly := s"${name.value}-assembly_${scalaBinaryVersion.value}-${version.value}.jar",
```
zsxwing: Add the scala version to the assembly jar

build.sbt
```diff
@@ -92,8 +94,13 @@ lazy val assemblySettings = Seq(
...
 
-  assemblyShadeRules in assembly := Seq(
+  assemblyShadeRules in assembly :=
+    (if (scalaBinaryVersion.value == "2.11") Seq(
```
zsxwing: Scala 2.11 compiler doesn't happen when sharding json4s :(

coreTest/src/main/scala/Test.scala
```diff
@@ -18,12 +18,14 @@ package test
 
 import shadedelta.io.delta.tables._
 
+import java.nio.file.Files
+
+import scala.collection.JavaConverters._
+
 import org.apache.spark.sql._
 import org.apache.spark.sql.functions._
-// import shadedelta.org.apache.commons.io.FileUtils
```
zsxwing: This one is not in the shaded jar. So I changed to use `Files` to clean up files. I just realized this test leaks tons of files in my laptop :)

build.sbt
tdas: what is the error?

zsxwing:
```
hive/src/main/scala/io/delta/hive/DeltaStorageHandler.scala:189: Symbol 'term org.json4s' is missing from the classpath.
[error] This symbol is required by 'method org.apache.spark.sql.types.DataType.jsonValue'.
```
The above error when compiling codes. Looks like the compiler reads the original `org.json4s` name from the metadata of `method org.apache.spark.sql.types.DataType.jsonValue` and tries to load it.

tdas:
LGTM.

PR #24, opened 2020-03-07
Optimize the connector memory usage

zsxwing:
- Disable Spark UI and remove `SparkListener`s to reduce the memory usage of Spark. They are unnecessary for the Spark cluster in connector.
- Set the `DeltaLog` cache size to 1 if the user doesn't specify it to avoid caching lots of `DeltaLog` objects.

### DIFF
--- a/hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
+++ b/hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
@@ -268,11 +268,14 @@ object DeltaHelper extends Logging {
   /**
    * Start a special Spark cluster using local mode to process Delta's metadata. The Spark UI has
    * been disabled and `SparkListener`s have been removed to reduce the memory usage of Spark.
-   * `DeltaLog` cache size is also set to "1" to cache only the recent accessed `DeltaLog`.
+   * `DeltaLog` cache size is also set to "1" if the user doesn't specify it, to cache only the
+   * recent accessed `DeltaLog`.
    */
   def spark: SparkSession = {
     // TODO Configure `spark` to pick up the right Hadoop configuration.
-    System.setProperty("delta.log.cacheSize", "1")
+    if (System.getProperty("delta.log.cacheSize") == null) {
+      System.setProperty("delta.log.cacheSize", "1")
+    }
     val sparkSession = SparkSession.builder()
       .master("local[*]")
       .appName("Delta Connector")
--- a/hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
+++ b/hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
@@ -97,6 +97,8 @@ object DeltaHelper extends Logging {
     val files = DeltaLog.filterFileList(
       snapshotToUse.metadata.partitionColumns, snapshotToUse.allFiles.toDF(), convertedFilterExpr)
       .as[AddFile](SingleAction.addFileEncoder)
+      // Drop unused potential huge fields
+      .map(add => add.copy(stats = null, tags = null))(SingleAction.addFileEncoder)
       .collect().map { f =>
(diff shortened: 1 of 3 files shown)


hive/src/test/scala/io/delta/hive/HiveConnectorSuite.scala
```diff
@@ -22,19 +22,10 @@ import io.delta.hive.test.HiveTest
 import io.delta.tables.DeltaTable
 
 import org.apache.spark.network.util.JavaUtils
-import org.apache.spark.sql.delta.DeltaLog
 import org.scalatest.BeforeAndAfterEach
 
 class HiveConnectorSuite extends HiveTest with BeforeAndAfterEach {
 
-  override def beforeEach(): Unit = {
-    DeltaLog.clearCache()
```
zsxwing: Not needed because the cache size is 1 now.

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
```diff
@@ -97,6 +97,8 @@ object DeltaHelper extends Logging {
     val files = DeltaLog.filterFileList(
       snapshotToUse.metadata.partitionColumns, snapshotToUse.allFiles.toDF(), convertedFilterExpr)
       .as[AddFile](SingleAction.addFileEncoder)
+      // Drop unused potential huge fields
+      .map(add => add.copy(stats = null, tags = null))(SingleAction.addFileEncoder)
```
windpiger: snapshot in connector has lots of unused actions, such as RemoveFile/SetTransaction etc.
It is better to do more special process to optimize the memory usage

hive/src/test/scala/io/delta/hive/HiveConnectorSuite.scala
windpiger: > Not needed because the cache size is 1 now.

If we query sql using beeline request to HiveServer, HiveServer will just hold one delta log cache, Actually it will receive lots of sql query(for example one table will be query lots of times), so it can not use the cache to improve the perfermance

windpiger: > Not needed because the cache size is 1 now.

If we query sql using beeline request to HiveServer, HiveServer will just hold one delta log cache, Actually it will receive lots of sql query(for example one table will be query lots of times), so it can not use the cache to improve the perfermance

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
windpiger: snapshot in connector has lots of unused actions, such as RemoveFile/SetTransaction etc.
It is better to do more special process to optimize the memory usage

zsxwing: Good catch. However, that's hard to fix unless forking the Snapshot class to rewrite it.

hive/src/test/scala/io/delta/hive/HiveConnectorSuite.scala
zsxwing: Good point. How about making it configurable? Such as, if `delta.log.cacheSize` is not set, we use `1`. Otherwise, we use the user specified value.

zsxwing: I updated the PR to set `delta.log.cacheSize` to 1 only if the user doesn't specify it.

hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
windpiger: > Snapshot
Yes,maybe we can seperate the delta log parser from Delta ,specially used for other engine's connector

windpiger:
LGTM

PR #29, opened 2020-03-23
Run unit tests for Tez

### LINKED ISSUE #21: Run unit tests in Tez mode
Tez is the default execution engine in Amazon EMR. It would be great that we can run our unit tests in Tez mode so that we can make sure it's working out of the box.

zsxwing:
- Move the unit tests for Hive MR to HiveMR project
- Add a new project HiveTez to run unit tests in Tez mode.
- Document that `hive.tez.input.format` should also be set to `io.delta.hive.HiveInputFormat`.
- Fix the dependency issue for Tez
  - Remove dependencies used by Spark UI because Tez has its own UI and its dependencies conflict with Spark's.
  - Exclude `org.xerial.snappy` and don't shade it. Tez uses a different version and doesn't work with the Spark one.

Closes #21


hive/src/test/scala/io/delta/hive/test/HiveTestSuite.scala
```diff
@@ -1,32 +0,0 @@
-/*
```
zsxwing: Remove this file as it's not needed. This was added to debug Hive cluster config.

.circleci/config.yml
```diff
@@ -8,5 +8,9 @@ jobs:
       - run:
           name: Run tests 
           command: |
-            build/sbt "+ coreTest/test"
-            build/sbt "+ hive/test"
+            # build/sbt "++ 2.12.8 coreTest/test"
```
windpiger: these two have been commented out, it is expected?

windpiger:
LGTM except above one little question

.circleci/config.yml
zsxwing: I was commented out them to debug. Have added these tests back.

zsxwing:
Thanks!

PR #30, opened 2020-03-25
Fail incorrect format and throw an error rather then returning NULL values

zsxwing:
When a user doesn't set the input format to `io.delta.hive.HiveInputFormat`, we will return NULL values right now. This is pretty bad as the user may not notice it. It's better to throw an error when the input format is not set.

### DIFF
--- a/hive/src/main/scala/io/delta/hive/DeltaInputFormat.scala
+++ b/hive/src/main/scala/io/delta/hive/DeltaInputFormat.scala
@@ -21,6 +21,7 @@ import java.net.URI
 
 import org.apache.hadoop.fs.FileStatus
 import org.apache.hadoop.fs.Path
+import org.apache.hadoop.hive.conf.HiveConf
 import org.apache.hadoop.hive.metastore.api.MetaException
 import org.apache.hadoop.hive.ql.exec.Utilities
 import org.apache.hadoop.hive.ql.io.parquet.read.DataWritableReadSupport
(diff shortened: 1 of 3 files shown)


hive/src/main/scala/io/delta/hive/DeltaInputFormat.scala
```diff
@@ -103,6 +105,45 @@ class DeltaInputFormat(realInput: ParquetInputFormat[ArrayWritable])
...
+      case "mr" =>
+        if (HiveConf.getVar(job, HiveConf.ConfVars.HIVEINPUTFORMAT) != deltaFormat) {
+          throw deltaFormatError(engine, HiveConf.ConfVars.HIVEINPUTFORMAT.varname, deltaFormat)
+        }
+      case "tez" =>
+        if (HiveConf.getVar(job, HiveConf.ConfVars.HIVETEZINPUTFORMAT) != deltaFormat) {
+          throw deltaFormatError(engine, HiveConf.ConfVars.HIVETEZINPUTFORMAT.varname, deltaFormat)
+        }
+      case other =>
+        throw new UnsupportedOperationException(s"The execution engine '$other' is not supported." +
```
zsxwing: Cannot test this branch (a.k.a, use `spark` execution engine) because the query will fail before this when loading Spark classes (We ship some of them in our jar and that's not the same version as Hive on Spark).

windpiger:
LGTM

zsxwing:
Thanks!

PR #31, opened 2020-03-25
Get table schema from table properties directly

zsxwing:
so that we don't need to store the property `delta.table.schema` in Hive Metastore.


windpiger:
LGTM

zsxwing:
Thanks!

PR #32, opened 2020-03-29
Log Delta operation duration and the table stats

zsxwing:
Log Delta operation duration and the table stats so that we have a better understanding on how long it takes to load Delta metadata.

### DIFF
--- a/hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
+++ b/hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
@@ -21,6 +21,7 @@ import java.util.Locale
 
 import scala.collection.mutable
 import scala.collection.JavaConverters._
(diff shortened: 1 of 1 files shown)


hive/src/main/scala/org/apache/spark/sql/delta/DeltaHelper.scala
```diff
@@ -102,7 +104,6 @@ object DeltaHelper extends Logging {
       // Drop unused potential huge fields
       .map(add => add.copy(stats = null, tags = null))(SingleAction.addFileEncoder)
       .collect().map { f =>
-        logInfo(s"selected delta file ${f.path} under $rootPath")
```
zsxwing: Remove this log as it's too verbose.

windpiger:
LGTM

zsxwing:
Thanks!

PR #44, opened 2020-07-01
Fix a small typo in the README, and add a create table command example

areese:


README.md
```diff
@@ -86,9 +86,18 @@ Here is an example of a CREATE TABLE command that defines an external Hive table
 ```SQL
 CREATE EXTERNAL TABLE deltaTable(col1 INT, col2 STRING)
 STORED BY 'io.delta.hive.DeltaStorageHandler'
-LOCATION '/delta/table/path'
+LOCATION 's3a://foo-bucket/bar-dir'
```
zsxwing, line 89: I think the Hive users should be familiar with the path syntax. It's probably not worth to add multiple examples.

areese: @zsxwing I added the location as that is what I had to do to get things to work.
I'm fine reverting the path.

README.md
```diff
@@ -89,6 +89,15 @@ STORED BY 'io.delta.hive.DeltaStorageHandler'
 LOCATION '/delta/table/path'
 ```
 
+Here is an  example of a CREATE TABLE command that defines an external Hive table pointing to a Delta table on the local filesystem at `/data/books.delta`.
```
zsxwing, line 92: Can you add a statement to explain how to use `LOCATION` rather than repeating the whole command. Such as,


`/delta/table/path` in `LOCATION` is a normal path. If there is no scheme in the path, it will use the default file system specified in your Hadoop configuration. You can add an explicit scheme to specify which file system you would like to use, such as `file://delta/table/path`, `s3://your-s3-bucket/delta/table/path`.

areese: Sure, let me take a shot at that.

areese: @zsxwing please let me know if that reads better?

README.md
```diff
@@ -91,6 +91,10 @@ LOCATION '/delta/table/path'
 
 `io.delta.hive.DeltaStorageHandler` is the class that implements Hive data source APIs. It will know how to load a Delta table and extract its metadata. The table schema in the `CREATE TABLE` statement must be consistent with the underlying Delta metadata. Otherwise, the connector will throw an error to tell you about the inconsistency.
 
+#### Specifying paths in LOCATION
+`/delta/table/path` in LOCATION is a normal path.  If there is no scheme in the path, it will use the default file system specified in your Hadoop configuration.
```
zsxwing, line 95:
```suggestion
`/delta/table/path` in LOCATION is a normal path. If there is no scheme in the path, it will use the default file system specified in your Hadoop configuration.
```

README.md
```diff
@@ -91,6 +91,10 @@ LOCATION '/delta/table/path'
...
+You can add an explicit scheme to specify which file system you would like to use, such as `file://delta/table/path`, `s3://your-s3-bucket/delta/table/path`.
```
zsxwing, line 96:
```suggestion
You can add an explicit scheme to specify which file system you would like to use, such as `file:///delta/table/path`, `s3://your-s3-bucket/delta/table/path`.
```
Just realized we need `///` here.

areese:
@zsxwing sorry it took me a bit, but those are good catches

PR #47, opened 2020-09-14
Non spark api 01 init

scottsand-db:


alpine/src/main/scala/DeltaLog.scala
```diff
@@ -0,0 +1,87 @@
...
+    try {
+      body
+    } finally {
+      deltaLogLock.unlock()
+    }
+  }
+}
+
+object DeltaLog {
+  val hadoopConf = new Configuration()
```
scottsand-db, line 63: pass this in to the DeltaLog constructor

alpine/src/main/scala/actions/actions.scala
```diff
@@ -0,0 +1,241 @@
...
+  @JsonIgnore
+  def simpleString: String = s"($minReaderVersion,$minWriterVersion)"
+}
+
+object Protocol {
+  // TODO: conf: apache.hadoop.conf.Configuration, requiredProtocol: Option[Protocol] ?
+  def apply(): Protocol = {
+    // TODO: use DELTA_PROTOCOL_DEFAULT_READER_VERSION) ?
+    // TODO: use DELTA_PROTOCOL_DEFAULT_WRITER_VERSION ?
+    Protocol(0, 0)
```
zsxwing, line 59: We should use `Action.readerVersion, Action.writerVersion`

scottsand-db: todo; think about this compared to delta-oss actions.scala :: Protocol class / object

scottsand-db: since we don't care about writing we can actually just sue Action.readerVersion; action.writerVersion

alpine/src/main/scala/exception/DeltaErrors.scala
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  def emptyDirectoryException(directory: String): Throwable = {
+    new FileNotFoundException(s"No file found in the directory: $directory.")
+  }
+
+  def logFileNotFoundException(
+    path: Path,
+    version: Long): Throwable = {
+// TODO
```
scottsand-db, line 36: remove the comments; leave a better TODO describing what we can do in the future (improve error msg)

alpine/src/main/scala/storage/HDFSReadOnlyLogStore.scala
```diff
@@ -0,0 +1,53 @@
...
+import java.io.{BufferedReader, FileNotFoundException, InputStreamReader}
+import java.nio.charset.StandardCharsets.UTF_8
+
+import scala.collection.JavaConverters._
+
+import org.apache.commons.io.IOUtils
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.{FileStatus, Path}
+
+class HDFSReadOnlyLogStore(hadoopConf: Configuration) extends ReadOnlyLogStore {
```
scottsand-db, line 28: TODO later: turn into ABC so that AzureReadOnlyLogStore can also extend that ABC...

scottsand-db:
Closing. This shouldn't be requesting to merge into delta-io/connectors, but instead into scottsand-db/connectors dev branch.

PR #51, opened 2020-10-27
Delta Standalone Reader

scottsand-db:
In this PR, we add a new project `standalone` which implements a Delta reader without using Spark. The reader provides APIs to access Delta transaction logs programmingly and a simple interface to read the data files in the table as well.

We also update the Hive connector project to use this new project and get rid of the embedded Spark from the Hive connector.


.circleci/config.yml
```diff
@@ -8,9 +8,9 @@ jobs:
       - run:
           name: Run tests 
           command: |
-            build/sbt "++ 2.12.8 coreTest/test"
+            build/sbt "++ 2.12.8 standalone/test"
             build/sbt "++ 2.12.8 hiveMR/test"
             build/sbt "++ 2.12.8 hiveTez/test"
```
zsxwing, line 13: Could you comment out Hive Tez test? The `hive-exec` har packages an old parquet version and makes the tests fail. It may take a while to resolve it. We don't need to block this large PR by this.

zsxwing (approved):
LGTM 🎉

PR #58, opened 2020-11-16
update README.md

scottsand-db:


README.md
```diff
@@ -42,7 +42,7 @@ Scala 2.11:
 
 #### SBT
 ```
-libraryDependencies += "io.delta" %% "delta-standalone" % "0.2.0"
+libraryDependencies += "io.delta" %% "delta-standalone" % "0.2.0-SNAPSHOT"
```
zsxwing, line 45: We prefer to show a release version rather than a snapshot version in README so that people can try it.

README.md
```diff
@@ -96,7 +98,20 @@ SET hive.input.format=io.delta.hive.HiveInputFormat;
 SET hive.tez.input.format=io.delta.hive.HiveInputFormat;
 ```
 
-The second step is to upload the above uber JAR to the machine that runs Hive. Finally, add the path of the uber JAR to Hive’s environment variable, `HIVE_AUX_JARS_PATH`. You can find this environment variable in the `hive-env.sh` file, whose location is `/etc/hive/conf/hive-env.sh` on an EMR cluster. This setting will tell Hive where to find the connector JAR.
+#### Add Hive uber JAR
+
+The second step is to upload the above uber JAR to the machine that runs Hive. Next, make the JAR accessible to Hive. There are several ways to do this, listed below. To verify that the JAR was properly added, run `LIST JARS;` in the Hive CLI.
+
+- in the Hive CLI, run `ADD JAR <path-to-jar>;`
+- move the uber JAR to the folder already pointed to by the `HIVE_AUX_JARS_PATH` environmental variable. (Its location is `/usr/lib/hive-hcatalog/share/hcatalog` in an EMR cluster)
```
zsxwing, line 106:
```suggestion
- add the uber JAR to a folder already pointed to by the `HIVE_AUX_JARS_PATH` environmental variable.
```

`/usr/lib/hive-hcatalog/share/hcatalog` is not a great place to put this jar. It's just a hack I did. It's better to not document it.

README.md
```diff
@@ -96,7 +98,20 @@ SET hive.input.format=io.delta.hive.HiveInputFormat;
...
+- modify the same `/etc/hive/conf/hive-site.xml` file as above, and add the following. (Note that this has to be done before you start the Hive CLI)
```
zsxwing, line 107:
```suggestion
- modify the same `hive-site.xml` file as above, and add the following. (Note that this has to be done before you start the Hive CLI)
```

README.md
```diff
@@ -145,7 +160,7 @@ Unfortunately, the table schema is a core concept of Hive and Hive needs it befo
 If the schema in the underlying Delta metadata is not consistent with the schema specified by `CREATE TABLE` statement, the connector will report an error when loading the table and ask you to fix the schema. You must drop the table and recreate it using the new schema. Hive 3.x exposes a new API to allow a data source to hook ALTER TABLE. You will be able to use ALTER TABLE to update a table schema when the connector supports Hive 3.x.
 
 #### Hive has three execution engines, MapReduce, Tez and Spark. Which one does this connector support?
-The connector supports MapReduce and Tez. It doesn't support Spark execution engine in Hive.
+The connector supports Tez (default) and MapReduce. It doesn't support Spark execution engine in Hive.
```
zsxwing, line 163: Let's revert chanage. `Tez` is default on EMR because EMR changes it. The real default value is `MapReduce` in official Apache Hive 2.x.

zsxwing (approved):
Thanks. Left some minor comments. Otherwise LGTM.

README.md
scottsand-db: it doesn't work without the "-SNAPSHOT" when I run it locally. Is that how it is supposed to work?

zsxwing: We usually put the last release version here. But because there is no release version available right now, it will fail. Once we have the first release, it should work.

scottsand-db: makes sense, thanks

PR #59, opened 2020-11-17
Add assembly target for the standalone jar

ucabvas:
I'd like to add an assembly target for the standalone jar. Tested locally with Hadoop 3.2.1. I did need to shade jackson to prevent classpath a clash.


zsxwing:
@ucabvas Could you describe your use case? How are you going to use the assembly jar? If users need to build an application, they will use maven or sbt which won't depend on the assembly jar.

zsxwing:
Closing this due to no update.

PR #61, opened 2020-11-18
Hive connector test case; assertion error msg improvement

scottsand-db:


hive/src/test/scala/io/delta/hive/HiveConnectorTest.scala
```diff
@@ -282,6 +282,35 @@ abstract class HiveConnectorTest extends HiveTest with BeforeAndAfterEach {
     }
   }
 
+  test("read a partitioned table with a partition filter") {
```
zsxwing, line 285: `should not touch files not needed when querying a partitioned table` is the same as this one. We don't need to add this duplicate test.

scottsand-db: sounds good.

zsxwing (approved):
LGTM

PR #62, opened 2020-12-01
Make Hive connector cache DeltaLog to speed up warm queries

scottsand-db:


build.sbt
```diff
@@ -132,6 +132,7 @@ lazy val hive = (project in file("hive")) dependsOn(standalone) settings (
   libraryDependencies ++= Seq(
     "org.apache.hadoop" % "hadoop-client" % hadoopVersion % "provided",
     "org.apache.parquet" % "parquet-hadoop" % "1.10.1" % "provided",
+    "com.google.guava" % "guava" % "14.0.1",
```
zsxwing, line 135: We don't need this line. This should already be provided by Hadoop.

zsxwing (commented):
LGTM except one minor comment

zsxwing (approved):

PR #64, opened 2020-12-07
Add DeltaLog.getCommitInfoAt to query the metadata in a commit

scottsand-db:
- add public `getCommitInfoAt` API to DeltaLog.java
- add internal impl
- add test
- add equals, hashcode to public java action files
- just ignore the goldenTable files


standalone/src/main/java/io/delta/standalone/DeltaLog.java
```diff
@@ -47,6 +50,14 @@
      */
     Snapshot getSnapshotForTimestampAsOf(long timestamp);
 
+    /**
+     * Retrieve a CommitInfo at a desired log version.
+     *
+     * @param version  the CommitInfo version to retrieve
```
zsxwing, line 56:
```suggestion
     * @param version the CommitInfo version to retrieve
```

standalone/src/main/scala/io/delta/standalone/internal/util/ConversionUtils.scala
```diff
@@ -16,16 +16,22 @@
...
+import io.delta.standalone.actions.{AddFile => AddFileJ, CommitInfo => CommitInfoJ, Format => FormatJ, JobInfo => JobInfoJ, Metadata => MetadataJ}
+import io.delta.standalone.internal.actions.{AddFile, CommitInfo, Format, JobInfo, Metadata}
 
 /**
  * Provide helper methods to convert from Scala to Java types.
  */
 private[internal] object ConversionUtils {
 
+  private implicit def convertOption[ScalaT, JavaT](opt: Option[ScalaT]): OptionalJ[JavaT] = {
+    OptionalJ.ofNullable(opt.getOrElse(null).asInstanceOf[JavaT])
```
zsxwing, line 32: This won't work when `ScalaT` and `JavaT` are not the same type, e.g.  the `operationMetrics` field.

standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
```diff
@@ -178,8 +234,30 @@ private[internal] case class SingleAction(
...
+        jgen.writeNullField(key)
+      } else {
+        jgen.writeFieldName(key)
+        // Write value as raw data, since it's already JSON text
+        jgen.writeRawValue(value)
+      }
+    }
+    jgen.writeEndObject()
+  }
 }
```
zsxwing, line 263: nit: missing the empty line at the end of the file

standalone/src/main/scala/io/delta/standalone/internal/util/ConversionUtils.scala
```diff
@@ -52,7 +58,7 @@ private[internal] object ConversionUtils {
       internal.schemaString,
       internal.partitionColumns.toList.asJava,
       internal.configuration.asJava,
-      java.util.Optional.ofNullable(internal.createdTime.getOrElse(null).asInstanceOf[Long]),
```
zsxwing, line 55: `null.asInstanceOf[Long]` will return 0. Let's add the following specialized methods and call them.
```
def toJavaLongOptional(opt: Option[Long]): OptionalJ[java.lang.Long] = opt match {
  case Some(v) => OptionalJ.ofNullable(v)
  case None => OptionalJ.empty()
}

def toJavaBooleanOptional(...)
def toJavaMapOptional(...)
```

standalone/src/main/scala/io/delta/standalone/internal/util/ConversionUtils.scala
```diff
@@ -62,4 +68,50 @@ private[internal] object ConversionUtils {
   def convertFormat(internal: Format): FormatJ = {
     new FormatJ(internal.provider, internal.options.asJava)
   }
+
+  /**
+   * Convert a [[CommitInfo]] (Scala) to a [[CommitInfoJ]] (Java)
+   */
+  def convertCommitInfo(internal: CommitInfo): CommitInfoJ = {
+    val notebookIdOpt: OptionalJ[String] = if (internal.notebook.isDefined) {
```
zsxwing, line 76: Let's also create a public `NotebookInfo` class  in case we may add new fields to `NotebookInfo` in Delta Lake core project.

zsxwing (changes requested):
Great work. Left some minor comments.

golden-tables/src/test/scala/io/delta/golden/GoldenTables.scala
```diff
@@ -337,6 +337,32 @@ class GoldenTables extends QueryTest with SharedSparkSession {
       Iterator(Protocol(99), Metadata(), file).map(a => JsonUtils.toJson(a.wrap)))
   }
 
+  /** TEST: DeltaLogSuiote > get commit info */
```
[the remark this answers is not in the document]
scottsand-db: didn't know I could do that! sure!

standalone/src/main/java/io/delta/standalone/DeltaLog.java
scottsand-db: actually, in javadoc, you want 2 spaces after the param name. that's the convention i've been using consistently throughout the javadoc, too.

standalone/src/main/scala/io/delta/standalone/internal/util/ConversionUtils.scala
scottsand-db: my intellij checker says it properly gets converted into an `Optional[java.util.Map]`, interesting

standalone/src/main/scala/io/delta/standalone/internal/util/ConversionUtils.scala
scottsand-db: ok. sounds good.

zsxwing (approved):

PR #65, opened 2020-12-08
Update Spark and Delta version for the golden tables project

scottsand-db:


build.sbt
```diff
@@ -304,8 +304,8 @@ lazy val goldenTables = (project in file("golden-tables")) settings (
   libraryDependencies ++= Seq(
     // Test Dependencies
     "org.scalatest" %% "scalatest" % "3.0.5" % "test",
-    "org.apache.spark" %% "spark-sql" % sparkVersion % "test",
-    "io.delta" %% "delta-core" % deltaVersion % "test",
+    "org.apache.spark" %% "spark-sql" % "3.0.0" % "test",
+    "io.delta" %% "delta-core" % "0.7.0" % "test",
     "commons-io" % "commons-io" % "2.8.0" % "test",
     "org.apache.spark" %% "spark-catalyst" % sparkVersion % "test" classifier "tests",
```
zsxwing, line 310: we also need to update `sparkVersion` here and below.

golden-tables/src/test/scala/io/delta/golden/GoldenTables.scala
```diff
@@ -359,18 +363,11 @@ class GoldenTables extends QueryTest with SharedSparkSession {
...
+      Some("foo")
     )
 
-    /**
-     * NOTE:
-     * Due to versioning conflicts between DeltaOSS and the Delta version used here, this
-     * CommitInfo is missing two fields. If you re-build this table and get test errors, append to
-     * the end of the commitInfo object in file
-     * src/test/resources/golden/deltalog-commit-info/_delta_log/00000000000000000000.json this:
-     * "operationMetrics":{"test":"test"},"userMetadata":"foo"
```
zsxwing, line 371: Should we also update the test to verify `operationMetrics` and `userMetadata`?

zsxwing (changes requested):

golden-tables/src/test/scala/io/delta/golden/GoldenTables.scala
scottsand-db: the test is already updated actually. i hardcoded the JSON values. now we just have the correct code to have the JSON commit info generated correctly. :)

build.sbt
scottsand-db: good catch. thanks.

zsxwing (approved):

PR #68, opened 2020-12-09
Fix NPE caused by asJava for Scala 2.11

zsxwing:


standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
```diff
@@ -50,6 +50,7 @@ class DeltaLogSuite extends FunSuite {
     withLogForGoldenTable("checkpoint") { log =>
       assert(log.snapshot.getVersion == 14)
       assert(log.snapshot.getAllFiles.size == 1)
+      log.snapshot.getAllFiles.hashCode()
```
zsxwing, line 53: This would trigger NPE without the fix

standalone/src/main/scala/io/delta/standalone/internal/util/ConversionUtils.scala
```diff
@@ -62,7 +74,7 @@ private[internal] object ConversionUtils {
       internal.modificationTime,
       internal.dataChange,
       internal.stats,
-      internal.tags.asJava)
+      mapAsJava(internal.tags))
```
zsxwing, line 77: tags is the only field that may be null.

PR #69, opened 2021-01-04
Wrap `lockInterruptibly` inside try-finally

djKooks:
Hello,
Because `lockInterruptibly` can cause `InterruptException`, seems it will be good to wrap inside try-finally block.
Please review 🙏 

Also, some question:
- How can I try in stand-alone mode?
- Seems it will be good to add exception handler. What will be good to put inside `catch`?


zsxwing:
Thanks a lot for your contribution. The change is not needed. `lockInterruptibly` won't hold the lock if it fails.

PR #75, opened 2021-02-11
Added Support for spark3.

roeyshemtov:
Using delta-standalone with spark3 is cannot be done right now.

The problem is spark3 using jackson-module with higher version of 2.6.7.1, where ScalaObjectMapper was moved from experimental package.


zsxwing:
Thanks for the PR. Could you take a look at the hive test failures? We also need to upgrade the Spark version used by the hive connector. Otherwise, it would fail.

zsxwing:
Closing this as a duplicate of #137.

PR #76, opened 2021-02-11
Added support for partition predicate pushdown.

roeyshemtov:


roeyshemtov:
@zsxwing Please review it, very important feature.

standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
```diff
@@ -69,6 +69,18 @@ private[internal] class SnapshotImpl(
       // the time zone ID if it exists, else null
       hadoopConf.get(StandaloneHadoopConf.PARQUET_DATA_TIME_ZONE_ID))
 
+  override def open(nestedPartition: (String, String)*): CloseableIterator[RowParquetRecordJ] = {
+    if (nestedPartition.isEmpty) return open()
+    CloseableParquetDataIterator(
+      allFilesScala
+        .filter(_.partitionValues.toList == nestedPartition.toList)
```
malonsocasas, line 76: Could we be less strict ? Accept partially defined partitions ?

malonsocasas (approved):

standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
noslowerdna: +1, I think we should only first filter partitionValues to contain only the keys supplied in `nestedPartition`, before checking for equality.

standalone/src/main/java/io/delta/standalone/Snapshot.java
```diff
@@ -53,4 +54,25 @@
...
+     * data will contain only the data related to partitionX=1 and partitionY=2
+     *
+     * Each open() will return data of only 1 partition (nested partitions allow),
+     * This method not supporting in reading multiple partitions
+     *
+     * @param nestedPartition key value of partition name and partition value, if specified will only
+     *                   return the data related to the specified partition
+     * @return
+     */
+    CloseableIterator<RowRecord> open(Tuple2<String, String>... nestedPartition);
```
noslowerdna, line 77: Maybe rename `nestedPartition` to something like `partitionFilterValues` - it should be plural and indicate the purpose of filtering.

standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
```diff
@@ -69,6 +69,18 @@ private[internal] class SnapshotImpl(
       // the time zone ID if it exists, else null
       hadoopConf.get(StandaloneHadoopConf.PARQUET_DATA_TIME_ZONE_ID))
 
+  override def open(nestedPartition: (String, String)*): CloseableIterator[RowParquetRecordJ] = {
+    if (nestedPartition.isEmpty) return open()
+    CloseableParquetDataIterator(
+      allFilesScala
```
noslowerdna, line 75: I wonder if it could be pushed down any further, to the transaction log and checkpoint readers? For a large dataset creating this `allFilesScala` collection that contains everything could still be a performance / memory concern.

dennyglee:
Hi @roeyshemtov  - per the above comment, would it be possible to update your PR and re-base it on the DeltaScan in #156 ?  Thanks!

dennyglee:
Closing this issue but please re-open if appropriate.  Thanks!

Issue #28, opened 2020-03-12
Presto doesn't recognise Hive-External table for Delta: com.facebook.presto.spi.PrestoException: InputFormat is not present in StorageFormat


amarsladdha:
Hello,

I am seeing following exception when I used this connector with presto. Can you please help me with this issue.

### A) My server setup:
1. Spark 2.4.4
2. Hadoop 3.2.1
3. Hive 2.3.6
4. Scala 2.11.12
5. Presto 0.228

### B) Build changes:
I had to change the change some build options in order create connector jars for my cluster

```
-val sparkVersion = "2.4.3"
-val hadoopVersion = "2.7.2"
-val hiveVersion = "2.3.3"
+val sparkVersion = "2.4.4"
+val hadoopVersion = "3.2.1"
+val hiveVersion = "2.3.6"
 val deltaVersion = "0.5.0"
 
 lazy val commonSettings = Seq(
@@ -53,8 +53,8 @@ lazy val core = (project in file("core"))
     name := "delta-core-shaded",
     libraryDependencies ++= Seq(
       "io.delta" %% "delta-core" % deltaVersion excludeAll ExclusionRule("org.apache.hadoop"),
-      "org.apache.spark" %% "spark-sql" % sparkVersion excludeAll ExclusionRule("org.apache.hadoop"),
-      "org.apache.hadoop" % "hadoop-client" % hadoopVersion % "provided"
+      "org.apache.spark" %% "spark-sql" % sparkVersion excludeAll ExclusionRule("org.apache.hadoop") exclude("org.glassfish.hk2", "hk2-utils") exclude("org.glassfish.hk2", "hk2-locator") exclude("javax.validation", "validation-api"),
+      "org.apache.hadoop" % "hadoop-client" % hadoopVersion % "provided" excludeAll ExclusionRule("org.slf4j", "slf4j-log4j12")
     ),

```

### C) Exception:

```
com.facebook.presto.spi.PrestoException: InputFormat is not present in StorageFormat
	at com.facebook.presto.hive.metastore.StorageFormat.getInputFormat(StorageFormat.java:57)
	at com.facebook.presto.hive.BackgroundHiveSplitLoader.loadPartition(BackgroundHiveSplitLoader.java:291)
	at com.facebook.presto.hive.BackgroundHiveSplitLoader.loadSplits(BackgroundHiveSplitLoader.java:264)
[...]

PR #82, opened 2021-03-30
Add Apache PrestoDB support

rumeshkrish:
## Build the uber JAR

- To compile the project, run `build/sbt presto/compile`
- To generate the uber JAR that contains all libraries needed for PrestoDB, run `build/sbt presto/assembly`

The above commands will generate the following JAR:

``` 
hive/target/scala-2.12/delta-presto-assembly_2.12-0.2.0.jar
```
This uber JAR includes the all its dependencies required for Apache PrestoDB Hive connector.

## Background 

- Deltalake PrestoDB connector needs `columns` and `columns.types` for schema validation. It is done [here](https://github.com/prestodb/presto/pull/15852).

## TODO

- PrestoDB missing `DeltaLakeSplitConvertor` for converting delta hive split into presto custom split with partition column information in `customSplitInfo`.  Will be added once `delta-presto` artifact released. 
Note:- For more details about [CustomSplitConverter](https://github.com/prestodb/presto/blob/master/presto-hive/src/main/java/com/facebook/presto/hive/util/CustomSplitConverter.java) and [CustomSplitConversionUtils](https://github.com/prestodb/presto/blob/master/presto-hive/src/main/java/com/facebook/presto/hive/util/CustomSplitConversionUtils.java).

- PrestoDB missing partition pruning information. Presto engine need to pass the push down predicated into `hive.io.filter.expr.serialized` jobConf.

## Setting up PrestoDB 

**Note:-** `TODO` list must be completed before testing this `jar` in [PrestoDB](https://prestodb.io/docs/current/).

Add the `delta-presto-assembly_2.12-0.2.0.jar` into `<presto_install>/plugin/hive-hadoop2/` directory.

## Create a table

After finishing setup, you should be able to create a Delta table in Metastore.

Right now the connector supports only EXTERNAL Hive tables. The Delta table must be created using Spark before an external Hive table can reference it.

Here is an example of a CREATE TABLE command that defines an external Hive table pointing to a Delta table on s3://foo-bucket/bar-dir.
```
CREATE EXTERNAL TABLE deltaTable(col1 INT, col2 STRING)
ROW FORMAT SERDE
  'org.apache.hadoop.hive.ql.io.parquet.serde.ParquetHiveSerDe'
STORED AS INPUTFORMAT
  'io.delta.hive.DeltaInputFormat'
OUTPUTFORMAT
  'io.delta.hive.DeltaOutputFormat'
LOCATION '/delta/table/path'
```

## FAQ

**Do I need to specify the partition columns when creating a Delta table?**
No. The partition columns are read from the underlying Delta metadata. The connector will know the partition columns and use this information to do the partition pruning automatically.


rumeshkrish:
@zsxwing  Can you please review this PR ?

rumeshkrish:
Link https://github.com/delta-io/connectors/issues/28

zsxwing:
Thanks for the contribution. I verified it worked on EMR. Regarding the changes, since the differences in `presto.DeltaInputFormat` are pretty tiny, it would be great if we can just try to publish an assembly jar working for both Hive and Presto.

For the issue that Presto's Hive fork doesn't have the `getUseVectorizedInputFileFormat` method. We can write the following method to use reflection to find the proper method to call:

```
  private def getUseVectorizedInputFileFormat(conf: Configuration): Boolean = {
    val getIsVectorized =
      try {
        classOf[Utilities].getDeclaredMethod(
          "getUseVectorizedInputFileFormat", classOf[Configuration])
      } catch {
        case _: NoSuchMethodException =>
          classOf[Utilities].getDeclaredMethod("getIsVectorized", classOf[Configuration])
      }
    getIsVectorized.invoke(null, conf).asInstanceOf[java.lang.Boolean]
  }

```

Regarding the `checkHiveConf` issue, we can move it to `DeltaStorageHandler` and call it in `DeltaStorageHandler.configureInputJobProperties`.

In addition, I hit the following error when testing this:
```
FAILED: SemanticException [Error 10055]: Output Format must implement HiveOutputFormat, otherwise it should be either IgnoreKeyTextOutputFormat or SequenceFileOutputFormat
```
Could you make `hive.DeltaOutputFormat` extend `HiveOutputFormat` to fix it?

rumeshkrish:
Thanks @zsxwing  for your feedback and review. I will make those changes.

zsxwing:
> Regarding the `checkHiveConf` issue, we can move it to `DeltaStorageHandler` and call it in `DeltaStorageHandler.configureInputJobProperties`.

I just realized that `configureInputJobProperties` cannot access `JobConf`. Do you know how to detect Presto environment to skip the check? Maybe call `Class.forName("com.facebook.presto.hive.HiveUtil")` and if it fails, it should not be in Presto?

YannByron:
If want query the same delta table by both presto and hive,  should I define two external Hive tables which has the same location but different input/output formats in this solution ?
If yes, is there a better way that use unified way to support the two ?

zsxwing:
Closing this. We have decided to build a native Presto connector rather than using the hive connector so that we can provide better support in Presto.

rohanpednekar:
Just to update, Delta lake connector for PrestoDB PR prestodb/presto#16843 is merged to the Presto master. 
Thanks, @vkorukanti for your contribution.

PR #83, opened 2021-04-02
Create AUTHORS.md

dennyglee:
Create authors file.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -1,6 +1,4 @@
 # This is the official list of the Delta Lake Project Authors for copyright purposes.
-# This file is distinct from the CONTRIBUTORS files.
-# See the latter for an explanation.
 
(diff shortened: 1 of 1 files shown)


AUTHORS.md
```diff
@@ -0,0 +1,9 @@
+# This is the official list of the Delta Lake Project Authors for copyright purposes.
+# This file is distinct from the CONTRIBUTORS files.
```
zsxwing, line 2: Can we remove this? There is no explanation in CONTRIBUTORS. I don't see the CONTRIBUTORS file in tensorflow either...

dennyglee: Sorry about that, i was thinking we should update the CONTRIBUTORS as well, but one step at a time on this.  Removed for now.

zsxwing (approved):
LGTM

PR #93, opened 2021-06-14
Pass down hadoopConf to ParquetReader

FX196:
Credit to @zsxwing

### DIFF
--- a/standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
@@ -95,7 +95,7 @@ private[internal] class SnapshotImpl(
           JsonUtils.mapper.readValue[SingleAction](line)
         }
       } else if (path.endsWith("parquet")) {
-        ParquetReader.read[SingleAction](path, ParquetReader.Options(hadoopConf = hadoopConf)).toSeq
+        ParquetReader.read[SingleAction](path, ParquetReader.Options(timeZone = readTimeZone, hadoopConf = hadoopConf)).toSeq
       } else Seq.empty[SingleAction]
     }.toList
   }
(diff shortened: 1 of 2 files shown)


zsxwing:
- Could you also pass hadoopConf down to https://github.com/delta-io/connectors/blob/1912ac5b6d4f3742d7d0d21b52e03846dd5e8683/standalone/src/main/scala/io/delta/standalone/internal/data/CloseableParquetDataIterator.scala#L119 ?
- We can also move `readTimeZone` from https://github.com/delta-io/connectors/blob/1912ac5b6d4f3742d7d0d21b52e03846dd5e8683/standalone/src/main/scala/io/delta/standalone/internal/data/CloseableParquetDataIterator.scala#L42 into SnapshotImpl and pass it to the place you are modifying.

standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
```diff
@@ -84,7 +95,7 @@ private[internal] class SnapshotImpl(
           JsonUtils.mapper.readValue[SingleAction](line)
         }
       } else if (path.endsWith("parquet")) {
-        ParquetReader.read[SingleAction](path).toSeq
+        ParquetReader.read[SingleAction](path, ParquetReader.Options(hadoopConf = hadoopConf)).toSeq
```
zsxwing, line 98: we should also pass the timezone to `ParquetReader.Options`.

standalone/src/main/scala/io/delta/standalone/internal/data/CloseableParquetDataIterator.scala
```diff
@@ -116,6 +114,7 @@ private[internal] case class CloseableParquetDataIterator(
    * @return the iterable for the next data file in `dataFilePathsIter`, not null
    */
   private def readNextFile: ParquetIterable[RowParquetRecord] = {
-    ParquetReader.read[RowParquetRecord](dataFilePathsIter.next(), Options(readTimeZone))
+    ParquetReader.read[RowParquetRecord](
+      dataFilePathsIter.next(), Options(readTimeZone, hadoopConf = hadoopConf))
```
zsxwing, line 118:
```suggestion
      dataFilePathsIter.next(), Options(timeZone = readTimeZone, hadoopConf = hadoopConf))
```

zsxwing (approved):
LGTM. Thanks!

PR #94, opened 2021-06-15
Fix issue with data sources that do not support streaming of binary files

gbrueckl:
fixes the issues thats described here https://blog.crossjoin.co.uk/2021/03/07/parquet-files-in-power-bi-power-query-and-the-streamed-binary-values-error/ by using an optional parameter


dennyglee (approved):
LGTM!

PR #98, opened 2021-07-07
Fix the sbt launch download url

zsxwing:
The bintray url is not working now. Use the `repo.typesafe.com` link instead.


jaceklaskowski:
I'd vote to remove the file whatsoever. Why is this file required actually?

zsxwing:
See my comment in https://github.com/delta-io/delta/pull/711#issuecomment-876565030

PR #102, opened 2021-07-20
[#101] add incremental changes api

### LINKED ISSUE #101: Add an API to expose incremental metadata changes in a Delta table
We need an API similar to https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/DeltaLog.scala#L222 so that a streaming reader can load changes in a Delta table incrementally.

scottsand-db:
- resolves #101 
- adds `DeltaLog::getChanges` API. Exposes an iterator of `VersionDelta`s, which is a java wrapper class containing the version and actions.
- includes `DeltaLog.java` public interface
- includes `DeltaLogImpl.scala` internal implementation
- adds remaining `Actions` missing from Delta OSS (RemoveFile, AddCDCFIle, Protocol, SetTransaction) as both internal scala representations and public java interfaces
- includes tests in `DeltaLogSuite.scala`
- increases goldenTables project delta oss version to 0.8.0 (to get access to AddCDCFile)
- proper javadoc TODO

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,3 +188,6 @@ There are two mediums of communication within the Delta Lake community.
   - [Login here](https://delta-users.slack.com/)
 
 - Public [Mailing list](https://groups.google.com/forum/#!forum/delta-users)
+
+# Local Development & Testing
+- before local debugging of `standalone` tests in IntelliJ, run all `standalone` tests using SBT. This helps IntelliJ recognize the golden tables as class resources.
--- a/README.md
+++ b/README.md
@@ -190,4 +190,4 @@ There are two mediums of communication within the Delta Lake community.
 - Public [Mailing list](https://groups.google.com/forum/#!forum/delta-users)
 
 # Local Development & Testing
-- before local debugging of `standalone` tests in IntelliJ, run all `standalone` tests using SBT. This helps IntelliJ recognize the golden tables as class resources.
+- Before local debugging of `standalone` tests in IntelliJ, run all `standalone` tests using SBT. This helps IntelliJ recognize the golden tables as class resources.
--- a/standalone/src/main/java/io/delta/standalone/VersionDelta.java
+++ b/standalone/src/main/java/io/delta/standalone/VersionDelta.java
@@ -1,0 +1,23 @@
+package io.delta.standalone;
+
+import io.delta.standalone.actions.Action;
+
+import java.util.List;
+
+public class VersionDelta {
+    private final long version;
+    private final List<Action> actions;
+
+    public VersionDelta(long version, List<Action> actions) {
+        this.version = version;
+        this.actions = actions;
+    }
+
+    public long getVersion() {
+        return version;
+    }
+
+    public List<Action> getActions() {
+        return actions;
+    }
+}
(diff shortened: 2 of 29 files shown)


README.md
```diff
@@ -188,3 +188,6 @@ There are two mediums of communication within the Delta Lake community.
   - [Login here](https://delta-users.slack.com/)
 
 - Public [Mailing list](https://groups.google.com/forum/#!forum/delta-users)
+
+# Local Development & Testing
+- before local debugging of `standalone` tests in IntelliJ, run all `standalone` tests using SBT. This helps IntelliJ recognize the golden tables as class resources.
```
tdas, line 193: before -> Before

standalone/src/main/java/io/delta/standalone/actions/RemoveFile.java
```diff
@@ -0,0 +1,29 @@
+package io.delta.standalone.actions;
+
+import java.util.Optional;
+
+public class RemoveFile implements FileAction {
```
zsxwing, line 5: Could you also update `RemoveFile` to the latest version of https://github.com/delta-io/delta/blob/86bbe9905ebf8c57f5bf5078a5d18ad2b5081abe/core/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L308 ?

standalone/src/main/scala/io/delta/standalone/internal/exception/DeltaErrors.scala
```diff
@@ -99,4 +99,19 @@ private[internal] object DeltaErrors {
     new NullPointerException(s"Read a null value for field $fieldName, yet schema indicates " +
       s"that this field can't be null. Schema: ${schema.getTreeString}")
   }
+
+  def failOnDataLossException(expectedVersion: Long, seenVersion: Long): Throwable = {
+    new IllegalStateException(
+      s"""The stream from your Delta table was expecting process data from version $expectedVersion,
+         |but the earliest available version in the _delta_log directory is $seenVersion. The files
+         |in the transaction log may have been deleted due to log cleanup. In order to avoid losing
```
zsxwing, line 107: nit: Remove `In order to avoid ...`. That's for Structured Streaming.

standalone/src/main/java/io/delta/standalone/VersionDelta.java
```diff
@@ -0,0 +1,34 @@
...
+import io.delta.standalone.actions.Action;
+
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * {@link VersionDelta} is the representation of all actions (changes) to the Delta Table
+ * at a specific table version.
+ */
+public class VersionDelta {
```
zsxwing, line 12: Maybe `VersionLog`? @tdas do you have a better name for this? This is the version number and the collection of actions in this version.

zsxwing (commented):
Thanks! Left some nits.

standalone/src/main/java/io/delta/standalone/VersionDelta.java
scottsand-db: `VersionLog` sounds good to me

zsxwing (approved):
LGTM

PR #105, opened 2021-07-28
[SC-80783] Change OSS Connectors repo to use Sonatype for the release

scottsand-db:
Update build.sbt to use sonatype and not bintray, as bintray is sunset.

Made changes similar to those made to delta-io/delta when the same thing was done (https://github.com/delta-io/delta/commit/926e30da6b6e86448cdb4739a082e0b566eec129)

Tested by staging a sonatype release. Changed the `version.sbt` to be `0.2.1-SNAPSHOT-sonatype-test` and was able to see that new package version in the sonatype staging repository. The pgp keys were present, too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 *.swp
 *~
 .DS_Store
+.bsp/
 .cache
 .classpath
 .ensime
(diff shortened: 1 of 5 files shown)


zsxwing (approved):
Cool! LGTM.

PR #107, opened 2021-08-02
Support ADLS Gen1 and all simple data types

gbrueckl:
- added supprt for ADLS Gen1 and recursive load of folder content in general (option IterateFolderContent)
- better support for data types - now all simple data types are supported and casted correctly


dennyglee (approved):
LGTM!

PR #123, opened 2021-08-30
Builder classes for action classes

allisonport-db:
Many of the Java Action classes have large constructors, and Java doesn't allow default values. This pr adds builder classes for `AddFile`, `RemoveFile`, `CommitInfo`, `Metadata`, `JobInfo`.

### DIFF
--- a/hive/src/main/scala/io/delta/hive/DeltaHelper.scala
+++ b/hive/src/main/scala/io/delta/hive/DeltaHelper.scala
@@ -80,8 +80,12 @@ object DeltaHelper {
         snapshotToUse.getAllFiles.asScala
       ).map { addF =>
         // Drop unused potential huge fields
-        val f = new AddFile(addF.getPath, addF.getPartitionValues, addF.getSize,
-          addF.getModificationTime, addF.isDataChange, null, null)
+        val f = new AddFile.AddFileBuilder(
+          addF.getPath,
+          addF.getPartitionValues,
+          addF.getSize,
+          addF.getModificationTime,
+          addF.isDataChange).build()
 
         val status = toFileStatus(fs, rootPath, f, blockSize)
         localFileToPartition +=
--- a/hive/src/main/scala/io/delta/hive/DeltaHelper.scala
+++ b/hive/src/main/scala/io/delta/hive/DeltaHelper.scala
@@ -80,7 +80,7 @@ object DeltaHelper {
         snapshotToUse.getAllFiles.asScala
       ).map { addF =>
         // Drop unused potential huge fields
-        val f = new AddFile.AddFileBuilder(
+        val f = AddFile.builder(
           addF.getPath,
           addF.getPartitionValues,
           addF.getSize,
--- a/standalone/src/main/java/io/delta/standalone/actions/AddFile.java
+++ b/standalone/src/main/java/io/delta/standalone/actions/AddFile.java
@@ -170,6 +170,9 @@ public final class AddFile implements FileAction {
             return this;
         }
 
+        /**
+         * @return a new {@code AddFile} with the same properties as {@code this}
+         */
         public AddFile build() {
             AddFile addFile = new AddFile(this);
             return addFile;
--- a/standalone/src/main/java/io/delta/standalone/actions/AddFile.java
+++ b/standalone/src/main/java/io/delta/standalone/actions/AddFile.java
@@ -49,6 +49,16 @@ public final class AddFile implements FileAction {
         this.tags = tags;
     }
 
+    private AddFile(AddFileBuilder builder) {
+        this.path = builder.path;
+        this.partitionValues = builder.partitionValues;
+        this.size = builder.size;
+        this.modificationTime = builder.modificationTime;
+        this.dataChange = builder.dataChange;
+        this.stats = builder.stats;
+        this.tags = builder.tags;
+    }
+
     /**
      * @return the relative path or the absolute path that should be added to the table. If it's a
      *         relative path, it's relative to the root of the table. Note: the path is encoded and
@@ -126,5 +136,40 @@ public final class AddFile implements FileAction {
     @Override
     public int hashCode() {
         return Objects.hash(path, partitionValues, size, modificationTime, dataChange, stats, tags);
+    }
+
+    public static class AddFileBuilder {
+        // add default values here, or in the constructor (values besides null)
+        private final String path;
+        private final Map<String, String> partitionValues;
+        private final long size;
+        private final long modificationTime;
+        private final boolean dataChange;
+        private String stats;
+        private Map<String, String> tags;
+
+        public AddFileBuilder(String path, Map<String, String> partitionValues, long size,
+                              long modificationTime, boolean dataChange) {
+            this.path = path;
+            this.partitionValues = partitionValues;
+            this.size = size;
+            this.modificationTime = modificationTime;
+            this.dataChange = dataChange;
+        }
+
+        public AddFileBuilder stats(String stats) {
+            this.stats = stats;
+            return this;
+        }
+
+        public AddFileBuilder tags(Map<String, String> tags) {
+            this.tags = tags;
+            return this;
+        }
+
+        public AddFile build() {
+            AddFile addFile = new AddFile(this);
+            return addFile;
+        }
     }
 }
--- a/standalone/src/main/java/io/delta/standalone/actions/AddFile.java
+++ b/standalone/src/main/java/io/delta/standalone/actions/AddFile.java
@@ -49,16 +49,6 @@ public final class AddFile implements FileAction {
         this.tags = tags;
     }
 
-    private AddFile(AddFileBuilder builder) {
-        this.path = builder.path;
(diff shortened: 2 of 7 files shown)


hive/src/main/scala/io/delta/hive/DeltaHelper.scala
```diff
@@ -80,8 +80,12 @@ object DeltaHelper {
         snapshotToUse.getAllFiles.asScala
       ).map { addF =>
         // Drop unused potential huge fields
-        val f = new AddFile(addF.getPath, addF.getPartitionValues, addF.getSize,
-          addF.getModificationTime, addF.isDataChange, null, null)
+        val f = new AddFile.AddFileBuilder(
```
zsxwing, line 83: For the builder pattern, we usually add a static builder method and just call the class `Builder`. E.g,

```
AddFile.builder(...).build()
```

We can use Guava as an example: 
- https://github.com/google/guava/blob/v30.1.1/guava/src/com/google/common/collect/ImmutableMap.java#L179
- https://github.com/google/guava/blob/v30.1.1/guava/src/com/google/common/collect/ImmutableMap.java#L367

standalone/src/main/java/io/delta/standalone/actions/AddFile.java
```diff
@@ -127,4 +137,45 @@ public boolean equals(Object o) {
...
+        public AddFileBuilder tags(Map<String, String> tags) {
+            this.tags = tags;
+            return this;
+        }
+
+        /**
+         * @return a new {@code AddFile} with the same properties as {@code this}
+         */
+        public AddFile build() {
+            AddFile addFile = new AddFile(this);
```
zsxwing, line 177: It's better to call the existing constructor directly rather than adding a new private AddFile constructor.

hive/src/main/scala/io/delta/hive/DeltaHelper.scala
scottsand-db: This is my bad. In the Jira description, I referenced a java builder example that used this static class.
Luckily, this is easy to refactor!

standalone/src/main/java/io/delta/standalone/actions/CommitInfo.java
```diff
@@ -196,4 +213,102 @@ public int hashCode() {
                 jobInfo, notebookInfo, clusterId, readVersion, isolationLevel, isBlindAppend,
                 operationMetrics, userMetadata);
     }
+
+    /**
+     * Builder class for CommitInfo. Enables construction of RemoveFile object with default values.
```
scottsand-db, line 218: nit: CommitInfo instead of RemoveFile

standalone/src/main/java/io/delta/standalone/actions/AddFile.java
```diff
@@ -127,4 +137,45 @@ public boolean equals(Object o) {
     public int hashCode() {
         return Objects.hash(path, partitionValues, size, modificationTime, dataChange, stats, tags);
     }
+
+    /**
+     * Builder class for AddFile. Enables construction of AddFile object with default values.
+     */
+    public static class AddFileBuilder {
+        // add default values here, or in the constructor (values besides null)
```
scottsand-db, line 145: nit: remove comment

scottsand-db: it would also be nice to comment
```
// required AddFile params
private final String path;
...

// optional AddFile params
private String stats;
```

standalone/src/main/java/io/delta/standalone/actions/AddFile.java
scottsand-db: +1. Again, my fault for referencing a java buidler pattern example in the Jira ticket that did it this way. Again, seems easy to refactor :)

standalone/src/main/java/io/delta/standalone/actions/CommitInfo.java
```diff
@@ -196,4 +213,102 @@ public int hashCode() {
...
+        private Map<String, String> operationParameters;
+        private Optional<JobInfo> jobInfo = Optional.empty();
+        private Optional<NotebookInfo> notebookInfo = Optional.empty();
+        private Optional<String> clusterId = Optional.empty();
+        private Optional<Long> readVersion = Optional.empty();
+        private Optional<String> isolationLevel = Optional.empty();
+        private Optional<Boolean> isBlindAppend = Optional.empty();
+        private Optional<Map<String, String>> operationMetrics = Optional.empty();
+        private Optional<String> userMetadata = Optional.empty();
+
```
scottsand-db, line 235: since we are switching to `CommitInfo.build().version(..).timestamp(..)` this will need a default, empty constructor

scottsand-db: this is only because all the member fields here are optional

standalone/src/main/java/io/delta/standalone/actions/JobInfo.java
```diff
@@ -69,4 +77,47 @@ public boolean equals(Object o) {
     public int hashCode() {
         return Objects.hash(jobId, jobName, runId, jobOwnerId, triggerType);
     }
+
+    /**
+     * Builder class for JobInfo. Enables construction of RemoveFile object with default values.
```
scottsand-db, line 82: nit: JobInfo instead of RemoveFile

scottsand-db (changes requested):
Great PR and thanks for doing this! Let a few minor comments.

Also, this may be overkill, but we should also include a simple test suite to, at the very least, exemplify how we expect users to use these builders.

@zsxwing what do you think?

standalone/src/main/java/io/delta/standalone/actions/CommitInfo.java
```diff
@@ -196,4 +196,113 @@ public int hashCode() {
...
                 operationMetrics, userMetadata);
     }
+
+    /**
+     * @return a new {@code CommitInfo.Builder}
+     */
+    public static Builder builder() {
+        return new Builder();
+    }
+
```
scottsand-db, line 206: nit: remove line

scottsand-db (approved):
LGTM

zsxwing (commented):
The code changes look good to me. Could you add tests for these builder classes? Just one test for each type, such as

```
addFileFromBuilder = AddFile.builder().....build()
addFileFromConstructor = new AddFile(...)
assert(addFileFromBuilder == addFileFromConstructor)
```

This will ensure we pass the parameters correctly in builder classes.

standalone/src/main/java/io/delta/standalone/actions/RemoveFile.java
```diff
@@ -54,4 +55,95 @@ public long getSize() {
...
+                Objects.equals(partitionValues, removeFile.partitionValues) &&
+                Objects.equals(size, removeFile.size) &&
+                Objects.equals(tags, removeFile.tags);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(path, deletionTimestamp, dataChange, extendedFileMetadata,
+                partitionValues, size, tags);
+    }
```
allisonport-db, line 77: Is there a reason `RemoveFile` didn't override `equals()` and `hashCode()` before? Needed to override `equals()` for testing purposes.

scottsand-db: it should have been there, so thanks for adding it!

standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
```diff
@@ -347,4 +352,158 @@ class DeltaLogSuite extends FunSuite {
...
+      Collections.emptyMap(),
+      metadataFromBuilderDefaults.getCreatedTime(),
+      null);
+    assert(metadataFromBuilderDefaults == metadataFromConstructorDefaults)
+
+    val metadataFromBuilder = MetadataJ.builder().id("test_id").name("test_name")
+      .description("test_description").format(new FormatJ("csv", Collections.emptyMap()))
+      .partitionColumns(List("id", "name").asJava).configuration(Map("test"->"foo").asJava)
+      .createdTime(0L)
+      .schema(new StructTypeJ(Array(new StructFieldJ("test_field", new IntegerType())))).build()
```
allisonport-db, line 372: Is there any style rule for this sort of chained method invocation?

scottsand-db: I don't see it specifically mentioned on the DB style guide. https://github.com/databricks/scala-style-guide#chaining.

However, I think one on each line looks most readable here?
```
MetadataJ.builder()
  .id("test_id")
  .name("test_name")
  .description("test_description")
  .format(new FormatJ("csv", Collections.emptyMap()))
  ...
```

zsxwing: +1 for one on each line

zsxwing (approved):
Thanks! LGTM.

PR #130, opened 2021-09-07
Add FieldMetadata to StructField

allisonport-db:
Currently the DSR ignores column metadata when reading the schema. This PR introduces a new type `FieldMetadata` which stores the metadata for a given field within the schema.
Changes:
• Add metadata attribute to `StructField`
• Adds a `FieldMetadata` class
• Includes column metadata in `DataTypeParser` `toJson` and `fromJson`
• New tests for schemas with column metadata

### DIFF
--- a/standalone/src/main/java/io/delta/standalone/types/FieldMetadata.java
+++ b/standalone/src/main/java/io/delta/standalone/types/FieldMetadata.java
@@ -106,13 +106,13 @@ public final class FieldMetadata {
 
     @Override
     public int hashCode() {
-            return metadata.entrySet()
-                    .stream()
-                    .mapToInt( entry -> (entry.getValue().getClass().isArray() ?
-                            (entry.getKey() == null ? 0 : entry.getKey().hashCode())^
-                            Arrays.hashCode((Object[]) entry.getValue()) :
-                            entry.hashCode()
-                    )).sum();
+        return metadata.entrySet()
+                .stream()
+                .mapToInt( entry -> (entry.getValue().getClass().isArray() ?
+                        (entry.getKey() == null ? 0 : entry.getKey().hashCode())^
+                                Arrays.hashCode((Object[]) entry.getValue()) :
+                        entry.hashCode())
+                ).sum();
         }
 
     /**
--- a/standalone/src/main/java/io/delta/standalone/types/FieldMetadata.java
+++ b/standalone/src/main/java/io/delta/standalone/types/FieldMetadata.java
@@ -108,11 +108,11 @@ public final class FieldMetadata {
     public int hashCode() {
             return metadata.entrySet()
                     .stream()
-                    .map( entry -> (entry.getValue().getClass().isArray() ?
+                    .mapToInt( entry -> (entry.getValue().getClass().isArray() ?
                             (entry.getKey() == null ? 0 : entry.getKey().hashCode())^
                             Arrays.hashCode((Object[]) entry.getValue()) :
                             entry.hashCode()
-                    )).mapToInt(i -> i.intValue()).sum();
+                    )).sum();
         }
 
     /**
--- a/standalone/src/main/java/io/delta/standalone/types/FieldMetadata.java
+++ b/standalone/src/main/java/io/delta/standalone/types/FieldMetadata.java
@@ -65,7 +65,7 @@ public final class FieldMetadata {
      * @param key  the key to check for
      * @return True if {@code this} contains a mapping for the given key, False otherwise
      */
-    public Boolean contains(String key) {
+    public boolean contains(String key) {
         return metadata.containsKey(key);
     }
 
@@ -80,7 +80,9 @@ public final class FieldMetadata {
 
     @Override
     public String toString() {
-        return metadata.entrySet().stream().map(entry -> entry.getKey() + "=" + entry.toString())
+        return metadata.entrySet()
+                .stream()
+                .map(entry -> entry.getKey() + "=" + entry.toString())
                 .collect(Collectors.joining(", ", "{", "}"));
     }
 
@@ -113,24 +115,24 @@ public final class FieldMetadata {
      * Builder class for FieldMetadata.
      */
     public static class Builder {
-        private Map<String, Object> metadata =new HashMap<String, Object>();
+        private Map<String, Object> metadata = new HashMap<String, Object>();
 
         public Builder putNull(String key) {
             metadata.put(key, null);
             return this;
         }
 
-        public Builder putLong(String key, Long value) {
+        public Builder putLong(String key, long value) {
             metadata.put(key, value);
             return this;
         }
 
-        public Builder putDouble(String key, Double value) {
+        public Builder putDouble(String key, double value) {
             metadata.put(key, value);
             return this;
         }
 
-        public Builder putBoolean(String key, Boolean value) {
+        public Builder putBoolean(String key, boolean value) {
             metadata.put(key, value);
             return this;
         }
--- a/standalone/src/main/java/io/delta/standalone/types/FieldMetadata.java
+++ b/standalone/src/main/java/io/delta/standalone/types/FieldMetadata.java
@@ -79,7 +79,7 @@ public final class FieldMetadata {
     }
 
     @Override
-    public String toString(){
+    public String toString() {
         return metadata.entrySet().stream().map(entry -> entry.getKey() + "=" + entry.toString())
                 .collect(Collectors.joining(", ", "{", "}"));
     }
--- a/standalone/src/main/java/io/delta/standalone/types/FieldMetadata.java
+++ b/standalone/src/main/java/io/delta/standalone/types/FieldMetadata.java
@@ -44,12 +44,10 @@ import java.util.Map;
 import java.util.HashMap;
 import java.util.stream.Collectors;
 
-
 /**
  * The metadata for a given {@link StructField}.
  */
 public final class FieldMetadata {
-    // TODO: should this be an immutable map created via the builder?
     private final Map<String, Object> metadata;
 
     private FieldMetadata(Map<String, Object> metadata) {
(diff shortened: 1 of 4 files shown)


standalone/src/main/scala/io/delta/standalone/internal/util/DataTypeParser.scala
```diff
@@ -131,11 +136,12 @@ private[standalone] object DataTypeParser {
...
   private def parseStructField(json: JValue): StructField = json match {
     case JSortedObject(
-    ("metadata", _: JObject),
+    ("metadata", metadata: JObject),
     ("name", JString(name)),
     ("nullable", JBool(nullable)),
     ("type", dataType: JValue)) =>
-      new StructField(name, parseDataType(dataType), nullable)
+      new StructField(name, parseDataType(dataType), nullable,
+         metadata.obj.toMap[String, AnyRef].asJava)
```
allisonport-db, line 144: This stores the metadata values inn StructField as a JValue, is that what we want?

standalone/src/main/java/io/delta/standalone/types/StructField.java
```diff
@@ -57,6 +61,20 @@ public StructField(String name, DataType dataType, boolean nullable) {
...
         this.dataType = dataType;
         this.nullable = nullable;
+        this.metadata = Collections.emptyMap();
+    }
+
+    /**
+     * @param name  the name of this field
+     * @param dataType  the data type of this field
+     * @param nullable  indicates if values of this field can be {@code null} values
+     * @param metadata metadata for this field
```
scottsand-db, line 71: nit: add another space

standalone/src/test/scala/io/delta/standalone/internal/DeltaDataReaderSuite.scala
```diff
@@ -336,4 +336,15 @@ class DeltaDataReaderSuite extends FunSuite {
       new IntegerType,
       new ArrayType(new DoubleType, true),
       false))
+
+  test("toJson fromJson with metadata") {
+    val schema_str =
+      """{"type":"struct",
+        |"fields":[
+        |{"name":"a","type":"integer","nullable":true,"metadata":{"test":"foo"}},
```
scottsand-db, line 344: you should also test an empty metadata here

scottsand-db (changes requested):
By biggest question, and this will help answer your question of "should we store the metadata map values in StructField as a JValue", is why aren't we including a `FieldMetadataBuilder.java` and `FieldMetadata.java` like [spark/.../Metadata](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/types/Metadata.scala) does?(renamed to put the `Field` keyword in there to avoid confusion with `actions/Metadata.java`)

Since we are exposing a `Map<String, Object>`, users can put an entry `("foo", new Foo())`. This would break things during serialization, and we want to prevent that. Limiting the column metadata value types would help prevent this.

I think this should look more like [spark/.../Metadata](https://github.com/apache/spark/blob/master/sql/catalyst/src/main/scala/org/apache/spark/sql/types/Metadata.scala).

@zsxwing what do you think?

standalone/src/test/scala/io/delta/standalone/internal/DeltaDataReaderSuite.scala
```diff
@@ -336,4 +336,15 @@ class DeltaDataReaderSuite extends FunSuite {
       new IntegerType,
       new ArrayType(new DoubleType, true),
       false))
+
+  test("toJson fromJson with metadata") {
+    val schema_str =
```
scottsand-db, line 341: super small nit: schemaStr instead

standalone/src/main/scala/io/delta/standalone/internal/util/DataTypeParser.scala
zsxwing: We should not expose JValue in public APIs. Could you convert them to java objects instead?

zsxwing:
> Since we are exposing a `Map<String, Object>`, users can put an entry `("foo", new Foo())`. This would break things during serialization, and we want to prevent that. Limiting the column metadata value types would help prevent this.

Yep. This is a good idea.

standalone/src/main/java/io/delta/standalone/types/FieldMetadata.java
```diff
@@ -0,0 +1,180 @@
...
+import java.util.Map;
+import java.util.HashMap;
+import java.util.stream.Collectors;
+
+
+/**
+ * The metadata for a given {@link StructField}.
+ */
+public final class FieldMetadata {
+    // TODO: should this be an immutable map created via the builder?
```
allisonport-db, line 52: should this be an immutable map?

standalone/src/main/scala/io/delta/standalone/internal/util/DataTypeParser.scala
```diff
@@ -111,10 +111,35 @@ private[standalone] object DataTypeParser {
...
+    val metadata = field.getMetadata()
 
     ("name" -> name) ~
       ("type" -> dataTypeToJValue(dataType)) ~
-      ("nullable" -> nullable)
+      ("nullable" -> nullable) ~
+      ("metadata" -> metadataValueToJValue(metadata))
+  }
+
+  def metadataValueToJValue(value: Any): JValue = {
```
allisonport-db, line 122: Why are some functions in DataTypeParser private and others aren't? Should this one be private?

allisonport-db:
@zsxwing ready for another look

standalone/src/main/scala/io/delta/standalone/internal/util/DataTypeParser.scala
scottsand-db: Good catch. The only public ones should be `toJson` and `fromJson`; everything else can be private.

standalone/src/main/java/io/delta/standalone/types/FieldMetadata.java
```diff
@@ -0,0 +1,180 @@
...
+     */
+    public static class Builder {
+        private Map<String, Object> metadata =new HashMap<String, Object>();
+
+        public Builder putNull(String key) {
+            metadata.put(key, null);
+            return this;
+        }
+
+        public Builder putLong(String key, Long value) {
```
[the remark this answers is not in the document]
scottsand-db: @zsxwing what do you think?

scottsand-db (approved):
LGTM with a few minor comments

standalone/src/main/java/io/delta/standalone/types/FieldMetadata.java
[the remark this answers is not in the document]
allisonport-db: Makes sense since there's also already `putNull()`. 

@zsxwing @scottsand-db 
What about for the arrays? Is there a situation where an array metadata value might have null as an element?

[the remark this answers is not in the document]
scottsand-db: I would go with whatever Ryan says here. For me, using primitives seems fine.

standalone/src/main/java/io/delta/standalone/types/FieldMetadata.java
```diff
@@ -0,0 +1,182 @@
...
+     * @return the value to which the specified key is mapped, or null if there is no mapping for
+     * the given key
+     */
+    public Object get(String key) {
+        return metadata.get(key);
+    }
+
+    @Override
+    public String toString() {
+        return metadata.entrySet()
```
zsxwing, line 83: It's better to add a toJson method and make `toString` return it. Otherwise, this `toString` implementation doesn't work for Array type.

standalone/src/main/java/io/delta/standalone/types/FieldMetadata.java
```diff
@@ -0,0 +1,182 @@
...
+                e.getValue().equals(that.metadata.get(e.getKey())) ||
+                        (e.getValue().getClass().isArray() &&
+                                that.metadata.get(e.getKey()).getClass().isArray() &&
+                                Arrays.equals(
+                                        (Object[]) e.getValue(),
+                                        (Object[]) that.metadata.get(e.getKey()))));
+    }
+
+    @Override
+    public int hashCode() { return metadata.hashCode(); }
```
zsxwing, line 105: Similar to `equals`, we need to handle Array type here as well. Otherwise, it's possible that equals return true but the two objects have different hash codes.

zsxwing (commented):
Left some minor comments. Great work!

standalone/src/main/java/io/delta/standalone/types/FieldMetadata.java
allisonport-db: I updated this so it should work with Array type. The `toJson` method is in `DataTypeParser` is it preferable to reference that from here instead? This function is for `StructField.buildFormattedString`

standalone/src/main/java/io/delta/standalone/types/FieldMetadata.java
```diff
@@ -0,0 +1,193 @@
...
+                        (e.getValue().getClass().isArray() &&
+                                that.metadata.get(e.getKey()).getClass().isArray() &&
+                                Arrays.equals(
+                                        (Object[]) e.getValue(),
+                                        (Object[]) that.metadata.get(e.getKey()))));
+    }
+
+    @Override
+    public int hashCode() {
+            return metadata.entrySet()
```
zsxwing, line 109: nit: indent

zsxwing (approved):
LGTM

standalone/src/main/java/io/delta/standalone/types/FieldMetadata.java
scottsand-db: I'll add investigating Java formatting tests as a ticket, should help both of us with our java code.

Issue #128, opened 2021-09-07
Support to define partition columns to create external table


YannByron:
For now, for a partitioned delta table, we can use sql as followed to create external table.

`CREATE EXTERNAL TABLE  tableName (col1 string, pt string)
ROW FORMAT SERDE  'org.apache.hadoop.hive.ql.io.parquet.serde.ParquetHiveSerDe'
STORED BY 'io.delta.hive.DeltaStorageHandler
LOCATION  '/path/to/delta/table'`

But, I wish to use sql as followed to do this, obviously define the partition columns. I think this way will be more suitable to hive-style ：

`CREATE EXTERNAL TABLE  tableName (col1 string) partitioned by (pt string)
ROW FORMAT SERDE  'org.apache.hadoop.hive.ql.io.parquet.serde.ParquetHiveSerDe'
STORED BY 'io.delta.hive.DeltaStorageHandler
LOCATION  '/path/to/delta/table'`

In the current delta-connectors implementation, the exception will be raised:
`Found partition columns (pt)  in table db.tableName. The partition columns in a Delta table will be read from its own metadata and should not be set manually`

If okey, I wanna work for this. Plz think about it, and i wait for your reply.

@zhu @scottsand-db

zsxwing:
Closing this as per the discussion in #132.

PR #132, opened 2021-09-08
support to specify partition columns to create external table (#128)

YannByron:


scottsand-db:
@YannByron is this ready for review? if so, can you please request a review from myself @scottsand-db and @zsxwing

YannByron:
@scottsand-db @zsxwing yes, it's ready. please review it.

zsxwing:
@YannByron can we make it a syntax sugar instead? For example,

```
create external table foo(c1 int)
partitioned by (c2 string)
stored by 'io.delta.hive.DeltaStorageHandler' location '...'
```
should be the same as

```
create external table foo(c1 int, c2 string)
stored by 'io.delta.hive.DeltaStorageHandler' location '...'
```

We don't want to store partition columns into the hive metastore.

YannByron:
@zsxwing 
Storing the partition columns into hive metastore, i think:
1) keep consistency with hive-style table. Users can know whether a delta table is partitioned or not and other related infomations just only via hive-cli;
2) it's useful for some ways accelerating query by skipping replaying delta log, like manifest;
3) backward compatibility. Users can choose one of ddl to create external table.

zsxwing:
There are two reasons we decide to hide the partition information from Hive:
- Hive always puts the partition columns to the end of the table schema. Delta allows to put partition columns in any location of the table schema.
- Delta's partition values need to be read from the transaction logs. The data file path cannot be used to calculate the partition value because Delta allows to use arbitrary file path as long as the partition value in AddFile is correct.

> * keep consistency with hive-style table. Users can know whether a delta table is partitioned or not and other related infomations just only via hive-cli;

This is a great benefit but due to the above reasons we cannot store the partition information in Hive Metastore.

YannByron:
@zsxwing Ok, got it.
If so, the pull request can be closed, and i'll use this as a internal feature.

zsxwing:
Closing the PR. Thanks for the contribution!

PR #136, opened 2021-09-16
Snapshot getAllFilesIter dummy API

scottsand-db:
- implements dummy API to return all AddFiles in a Snapshot as a `CloseableIterator<AddFile>`
- we want it to be a `CloseableIterator<AddFile>`, instead of an iterator, since in the future it will internally use `logStore. readAsIterator` (which is a `CloseableIterator`) and it is possible for the user to stop half-way through iterating over all the files. If they do that, then it is on them to call `close` on the log store iterator, since otherwise we wouldn't be able to detect that they have finished iterating.

### DIFF
--- a/standalone/src/main/java/io/delta/standalone/Snapshot.java
+++ b/standalone/src/main/java/io/delta/standalone/Snapshot.java
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package io.delta.standalone;
 
 import java.util.List;
@@ -35,6 +36,20 @@ public interface Snapshot {
      * @return all of the files present in this snapshot
      */
     List<AddFile> getAllFiles();
+
+    /**
+     * :: DeveloperApi ::
+     *
+     * Dummy API. Will be replaced in a later version with a proper implementation.
+     *
+     * Creates a {@link CloseableIterator} which can iterate over all of the files present in this
+     * snapshot.
+     *
+     * TODO: document ordering (i.e. forwards or reverse)
+     *
+     * @return a {@link CloseableIterator} to iterate over files
+     */
+    CloseableIterator<AddFile> getAllFilesIter();
 
     /**
      * @return the table metadata for this snapshot
--- a/standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedAddFileReplayIterator.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedAddFileReplayIterator.scala
@@ -1,0 +1,33 @@
+/*
+ * Copyright (2021) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.delta.standalone.internal
+
+import io.delta.standalone.actions.{AddFile => AddFileJ}
+import io.delta.standalone.data.CloseableIterator
+
+/**
+ * Dummy class. Will be replaced in a later version with a proper implementation.
+ */
+private[internal] class MemoryOptimizedAddFileReplayIterator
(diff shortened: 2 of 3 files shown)


standalone/src/main/java/io/delta/standalone/Snapshot.java
```diff
@@ -36,6 +37,20 @@
...
+     * Dummy API. Will be replaced in a later version with a proper implementation.
+     *
+     * Creates a {@link CloseableIterator} which can iterate over all of the files present in this
+     * snapshot.
+     *
+     * TODO: document ordering (i.e. forwards or reverse)
+     *
+     * @return a {@link CloseableIterator} to iterate over files
+     */
+    CloseableIterator<AddFile> getAllFilesIter();
```
zsxwing, line 52: It's better to call it `getAllFilesAsIterator` to be consistent with the API in OSS Delta LogStore.

standalone/src/main/java/io/delta/standalone/Snapshot.java
```diff
@@ -36,6 +37,20 @@
      */
     List<AddFile> getAllFiles();
 
+    /**
+     * :: DeveloperApi ::
+     *
+     * Dummy API. Will be replaced in a later version with a proper implementation.
```
zsxwing, line 43:
```
Returns a {@link CloseableIterator} to iterate over all of the files present in this snapshot. The caller should call the close method when the iterator is not used any more to release resources.
```

We should remove `Dummy API` and `TODO`.

standalone/src/main/java/io/delta/standalone/Snapshot.java
```diff
@@ -36,6 +37,20 @@
...
+     *
+     * Creates a {@link CloseableIterator} which can iterate over all of the files present in this
+     * snapshot.
+     *
+     * TODO: document ordering (i.e. forwards or reverse)
```
zsxwing, line 48: Files in a snapshot don't have any ordering guarantee.

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedAddFileReplayIterator.scala
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+
+package io.delta.standalone.internal
+
+import io.delta.standalone.actions.{AddFile => AddFileJ}
+import io.delta.standalone.data.CloseableIterator
+
+/**
+ * Dummy class. Will be replaced in a later version with a proper implementation.
```
zsxwing, line 23: remove this since it's not used right now.

standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
```diff
@@ -65,6 +65,10 @@ private[internal] class SnapshotImpl(
 
   override def getAllFiles: java.util.List[AddFileJ] = activeFiles
 
+  /** Dummy API. Will be replaced in a later version with a proper implementation. */
+  override def getAllFilesIter: CloseableIterator[AddFileJ] =
```
zsxwing, line 69:
```
override def getAllFilesAsIterator: CloseableIterator[AddFileJ] = {
  // TODO Implement an optimized iterator that doesn't load all `AddFile`s into the memory.
  val iter =  activeFiles.iterator()
   new CloseableIterator[AddFileJ] {
      override def hasNext: Boolean = iter.hasNext

      override def next(): AddFileJ = iter.next()

       override def close(): Unit = { }
   }
  iter
}
```

zsxwing (commented):
As per our offline discussion, let's wait for the feedback from people working on Presto connector.

scottsand-db:
TODO MiMa compatibility, too.

From CircleCI:
```
[error]  * abstract method getAllFilesIter()io.delta.standalone.data.CloseableIterator in interface io.delta.standalone.Snapshot is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("io.delta.standalone.Snapshot.getAllFilesIter")
java.lang.RuntimeException: delta-standalone: Binary compatibility check failed!
```

zsxwing:
This mima error is fine to ignore. We don't expect people to implement the Snapshot interface.

scottsand-db:
> This mima error is fine to ignore. We don't expect people to implement the Snapshot interface.

Yup, I'll just be configuring the `StandaloneMimaExcludes` class to ignore it.

standalone/src/main/java/io/delta/standalone/Snapshot.java
scottsand-db: Please discuss here: https://databricks.atlassian.net/browse/SC-85090

scottsand-db:
This PR is old and is replaced by https://github.com/delta-io/connectors/pull/156

PR #142, opened 2021-09-17
[NEW] [2] Checkpoint Prototype

scottsand-db:


standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -39,6 +39,8 @@
     private final Optional<Long> createdTime;
     private final StructType schema;
 
+    // TODO: add back in previous constructor (in DSR)
```
zsxwing, line 42: nit: remove unused comment

zsxwing (approved):
LGTM assuming the TODOs will be addressed in followup PRs.

PR #144, opened 2021-09-17
[DSW] [3] Conflict Checker Prototype

scottsand-db:


zsxwing (approved):
LGTM. Could you move the concurrent exceptions to `io.delta.standalone.exceptions` as public APIs? They are public in OSS Delta. We don't need to hide them.

PR #145, opened 2021-09-17
[DSW] [4] Improved commit API prototype with tests

scottsand-db:


standalone/src/main/java/io/delta/standalone/DeltaLog.java
```diff
@@ -89,6 +89,9 @@
     /** @return the path of the Delta table. */
     Path getPath();
 
+    /** @return the path of the delta log for this Delta table. */
+    Path getLogPath();
```
zsxwing, line 93: Why do we need to expose this API? If it's used by the unit test, we can just twist the test to avoid using the full log path, such as changing `==` to `contains`.

zsxwing (approved):
One minor question about the `logPath` API. Otherwise LGTM

standalone/src/main/java/io/delta/standalone/DeltaLog.java
scottsand-db: We could do that. I also figured that this API could be useful for future connectors, but perhaps that is an incorrect assumption.

scottsand-db: I'll remove.

PR #147, opened 2021-09-17
[DSW] [6] updateMetadata() prototype with tests

scottsand-db:
forgot to delete the old ReadOnlyLogStore files last PR..


standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
```diff
@@ -343,19 +373,11 @@ private[internal] class OptimisticTransactionImpl(
...
-    // TODO: this function is still incomplete
-    val needsProtocolUpdate = Protocol.checkProtocolRequirements(metadata, protocol)
-
-    if (needsProtocolUpdate.isDefined) {
-      newProtocol = needsProtocolUpdate
-    }
+    Protocol.checkMetadataProtocolProperties(metadata, protocol)
   }
 
   private def isCommitLockEnabled: Boolean = {
```
zsxwing, line 379: nit: Could you remove `isCommitLockEnabled` and the related code? In other words, `isCommitLockEnabled` should be `false`. It's unlikely we will hit the concurrent update issue in Delta Standalone Writer.

zsxwing (approved):

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
scottsand-db: but what about a LogStore implementation like this one: https://github.com/delta-io/delta/blob/master/contribs/src/main/scala/io/delta/storage/OracleCloudLogStore.scala

Where `isPartialWriteVisible` == `true`

zsxwing: This doesn't matter. This flag is trying to avoid updating the `_last_checkpoint` file in the same table concurrently (Azure will throw an exception when another writer is writing `_last_checkpoint`). It's unlikely in Delta Standalone Writer.

scottsand-db: SGTM.

PR #149, opened 2021-09-17
[DSW] [9] Production commit API

scottsand-db:


standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
```diff
@@ -130,16 +130,24 @@ private[internal] class OptimisticTransactionImpl(
     new CommitResult(commitVersion)
   }
 
+  /** Returns files matching the given predicates. */
   override def markFilesAsRead(
```
scottsand-db, line 134: in the next PR I will update this to return a DeltaScan and take in a single Expression

zsxwing (approved):

PR #155, opened 2021-09-24
Expression framework skeleton

scottsand-db:


scottsand-db:
I'll fix failing scalstyle Monday

standalone/src/main/java/io/delta/standalone/expressions/Util.java
```diff
@@ -0,0 +1,20 @@
+package io.delta.standalone.expressions;
```
zsxwing, line 1: This should be in private package.

standalone/src/main/java/io/delta/standalone/expressions/CastingComparator.java
```diff
@@ -0,0 +1,15 @@
+package io.delta.standalone.expressions;
+
+import java.util.Comparator;
+
+public class CastingComparator<T extends Comparable<T>> {
```
zsxwing, line 5: Any reason we cannot use `Comparable<T>` directly?

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
```diff
@@ -108,4 +111,24 @@ private[standalone] object DeltaLogImpl {
...
     new DeltaLogImpl(hadoopConf, path, path.getParent)
   }
+
+  /**
+   * Filters the given [[AddFile]]s by the given `partitionFilters`, returning those that match.
+   * @param files The active files in the DeltaLog state, which contains the partition value
+   *              information
+   * @param partitionFilters Filters on the partition columns
+   */
+  def filterFileList(
```
zsxwing, line 121: This is not needed for #136. Could you remove it?

standalone/src/main/scala/io/delta/standalone/internal/data/PartitionRowRecord.scala
```diff
@@ -0,0 +1,93 @@
...
+
+import java.sql.{Date, Timestamp}
+
+import io.delta.standalone.data.{RowRecord => RowRecordJ}
+import io.delta.standalone.types.{BooleanType, IntegerType, StructType}
+
+/**
+ * A RowRecord representing a Delta Lake partition of Map(partitionKey -> partitionValue)
+ */
+private[internal] class PartitionRowRecord(
```
zsxwing, line 27: Remove this since it's not ready. We can use the existing `RowParquetRecordImpl` to write unit tests for expressions.

standalone/src/main/java/io/delta/standalone/expressions/And.java
```diff
@@ -0,0 +1,20 @@
...
+public final class And extends BinaryOperator implements Predicate {
+
+    public And(Expression left, Expression right) {
+        super(left, right, "&&");
+    }
+
+    @Override
+    public Object nullSafeEval(Object leftResult, Object rightResult) {
+        if (!(leftResult instanceof Boolean) || !(rightResult instanceof Boolean)) {
+            throw new RuntimeException("'And' expression children.eval results must be Booleans");
```
zsxwing, line 15: We can use `IllegalArgumentException` for all of checks done in expressions.

standalone/src/main/java/io/delta/standalone/expressions/BinaryExpression.java
```diff
@@ -0,0 +1,38 @@
...
+        Object leftResult = left.eval(record);
+        if (null == leftResult) return null;
+
+        Object rightResult = right.eval(record);
+        if (null == rightResult) return null;
+
+        return nullSafeEval(leftResult, rightResult);
+    }
+
+    protected abstract Object nullSafeEval(Object leftResult, Object rightResult);
```
zsxwing, line 32: nit: missing javadoc

standalone/src/main/java/io/delta/standalone/expressions/Column.java
```diff
@@ -0,0 +1,55 @@
...
+    public Column(String name, DataType dataType) {
+        this.name = name;
+        this.dataType = dataType;
+
+        if (dataType instanceof IntegerType) {
+            evaluator = (record -> record.getInt(name));
+        } else if (dataType instanceof BooleanType) {
+            evaluator = (record -> record.getBoolean(name));
+        } else {
+            throw new RuntimeException("Couldn't find matching rowRecord DataType for column: " + name);
```
zsxwing, line 28: Could you complete the code?

standalone/src/main/java/io/delta/standalone/expressions/BinaryComparison.java
```diff
@@ -0,0 +1,20 @@
...
+
+    public BinaryComparison(Expression left, Expression right, String symbol) {
+        super(left, right, symbol);
+
+        // super asserted that left and right DataTypes were the same
+
+        comparator = Util.createCastingComparator(left.dataType());
+    }
+
+    protected int compare(Object leftResult, Object rightResult) {
```
zsxwing, line 17: How about exposing Comparator to subclasses instead, such as
```
Comparator<Object> getComparator
```
`Comparator` is a pretty common class and subclasses can reuse their existing code/library for comparator.

zsxwing (changes requested):

scottsand-db:
closing since we have opted to instead merge just a bare-bones `Expression.java` interface (with no child implementations) into master. we will merge a more complete expression framework into master when the framework is in better shape and not just prototype.

PR #156, opened 2021-09-27
Add Snapshot.scan API

scottsand-db:
- add `scan()`, `scan(expr`) APIs to Snapshot
- add `Expression` interface with no concrete child implementations (that is coming later)
- add `DeltaLog.java` and `DeltaLogImpl.scala` with Dummy `getFiles` implementation (just returns `activeFilesList.iterator()` instead of properly iterating over the delta log)


standalone/src/main/scala/io/delta/standalone/internal/DeltaScanImpl.scala
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * TODO this is currently a naive implementation, since
+ * a) it takes in the in-memory AddFiles.
+ * b) it uses the metadata.partitionColumns, but the metadata won't be known until the log files
+ *    are scanned
+ */
+final class DeltaScanImpl(
+    files: java.util.List[AddFileJ],
+    expr: Option[Expression] = None,
+    partitionColumns: Seq[String] = Nil) extends DeltaScan {
```
zsxwing, line 37: nit: `partitionColumns: Seq[String] = Nil` is not needed for now.

standalone/src/main/scala/io/delta/standalone/internal/DeltaScanImpl.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  val (metadataConjunction, dataConjunction) = (Option.empty[Expression], Option.empty[Expression])
```
zsxwing, line 39: nit: not needed for now

standalone/src/main/scala/io/delta/standalone/internal/DeltaScanImpl.scala
```diff
@@ -0,0 +1,63 @@
...
+    private val iter = files.iterator
+
+    override def hasNext: Boolean = iter.hasNext
+
+    override def next(): AddFileJ = iter.next()
+
+    override def close(): Unit = { }
+  }
+
+  override def getPushedPredicate: Optional[Expression] = if (metadataConjunction.isDefined) {
```
zsxwing, line 51: nit: we can make this return `Optional.empty()` for now.

standalone/src/main/scala/io/delta/standalone/internal/DeltaScanImpl.scala
```diff
@@ -0,0 +1,63 @@
...
+    Optional.of(metadataConjunction.get)
+  } else {
+    Optional.empty()
+  }
+
+  override def getResidualPredicate: Optional[Expression] = if (dataConjunction.isDefined) {
```
zsxwing, line 57: this should return `Optional.ofNullable(expr.getOrElse(null))`

zsxwing (approved):
LGTM except one minor issue in `InitialSnapshotImpl`

standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
```diff
@@ -218,4 +224,8 @@ private class InitialSnapshotImpl(
       Map.empty[URI, AddFile],
       0L, 0L, 1L, 1L)
   }
+
+  override def scan(): DeltaScan = new DeltaScanImpl(Nil.asJava)
+
+  override def scan(predicate: Expression): DeltaScan = new DeltaScanImpl(Nil.asJava)
```
zsxwing, line 230: `new DeltaScanImpl(Nil.asJava, Some(predicate))`.

zsxwing:
@scottsand-db could you fix the build? https://app.circleci.com/pipelines/github/delta-io/connectors/371/workflows/1ec7921f-920f-43c1-adc1-f16c0896efac/jobs/374 

The PR build didn't fail because it was not merged with the latest master

PR #157, opened 2021-09-28
[DSW] [13] Production Log Store

scottsand-db:


standalone/src/main/java/io/delta/standalone/storage/LogStore.java
```diff
@@ -105,6 +121,31 @@ public abstract void write(
...
+     * it should not generate partial files.
+     *
+     * @since 0.3.0 // TODO: double check this will be the new DSW version
+     *
+     * @param path  the path to write to, as a {@link String}
+     * @param actions  actions to be written
+     * @param overwrite  if true, overwrites the file if it already exists
+     * @param hadoopConf  the latest hadoopConf
+     */
+    public final void write(
```
zsxwing, line 140: ditto

standalone/src/main/java/io/delta/standalone/storage/LogStore.java
```diff
@@ -122,6 +163,26 @@ public abstract void write(
...
+     * (UTF-8 sorting) the given `path`. The result should also be sorted by the file name.
+     *
+     * @since 0.3.0 // TODO: double check this will be the new DSW version
+     *
+     * @param path  the path to load, as a {@link String}
+     * @param hadoopConf  the latest hadoopConf
+     * @return an Iterator of the paths lexicographically greater or equal to (UTF-8 sorting) the
+     *         given `path`
+     */
+    public final Iterator<FileStatus> listFrom(
```
zsxwing, line 179: ditto

standalone/src/main/java/io/delta/standalone/storage/LogStore.java
```diff
@@ -83,6 +83,22 @@ public LogStore(Configuration initHadoopConf) {
...
+     * Load the given file and return an `Iterator` of lines, with line breaks removed from each line.
+     * Callers of this function are responsible to close the iterator if they are done with it.
+     *
+     * @since 0.3.0 // TODO: double check this will be the new DSW version
+     *
+     * @param path  the path to load, as a {@link String}
+     * @param hadoopConf  the latest hadoopConf
+     * @return the CloseableIterator of lines in the given file.
+     */
+    public final CloseableIterator<String> read(String path, Configuration hadoopConf) {
```
zsxwing, line 98: Let's remove these methods and update tests to use the Path version instead. These methods make the interface more complicated but the only benefit is tests can save several chars. This interface should be the same as OSS Delta's LogStore.java.

standalone/src/main/scala/io/delta/standalone/internal/Checkpoints.scala
```diff
@@ -119,17 +121,17 @@ private[internal] trait Checkpoints {
...
+    store.write(LAST_CHECKPOINT, Iterator(json).asJava, true, hadoopConf) // overwrite = true
 
     doLogCleanup()
   }
 
   /** Loads the checkpoint metadata from the _last_checkpoint file. */
   private def loadMetadataFromFile(tries: Int): Option[CheckpointMetaData] = {
     try {
-      val checkpointMetadataJson = store.read(LAST_CHECKPOINT)
+      val checkpointMetadataJson = store.read(LAST_CHECKPOINT, hadoopConf).asScala
```
zsxwing, line 132: `checkpointMetadataJson` needs to be closed.

standalone/src/main/scala/io/delta/standalone/internal/ConflictChecker.scala
```diff
@@ -92,8 +94,12 @@ private[internal] class ConflictChecker(
    */
   private def createWinningCommitSummary(): WinningCommitSummary = {
     val deltaLog = currentTransactionInfo.deltaLog
-    val winningCommitActions = deltaLog.store.read(
-      FileNames.deltaFile(deltaLog.logPath, winningCommitVersion)).map(Action.fromJson)
+    val winningCommitActions = deltaLog.store
+      .read(FileNames.deltaFile(deltaLog.logPath, winningCommitVersion), deltaLog.hadoopConf)
```
zsxwing, line 98: we need to close the iterator returned by `read`.

standalone/src/main/scala/io/delta/standalone/internal/DeltaHistoryManager.scala
```diff
@@ -34,8 +37,9 @@ private[internal] case class DeltaHistoryManager(deltaLog: DeltaLogImpl) {
 
   /** Get the persisted commit info for the given delta file. */
   def getCommitInfo(version: Long): CommitInfo = {
-    val info = deltaLog.store.read(FileNames.deltaFile(deltaLog.logPath, version))
-      .iterator
+    val info = deltaLog.store
+      .read(FileNames.deltaFile(deltaLog.logPath, version), deltaLog.hadoopConf)
```
zsxwing, line 41: ditto: close

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
```diff
@@ -116,8 +117,13 @@ private[internal] class DeltaLogImpl private(
       }
       lastSeenVersion = version
 
-      new VersionLog(version,
-        store.read(p).map(x => ConversionUtils.convertAction(Action.fromJson(x))).toList.asJava)
+      new VersionLog(
+        version,
+        store.read(p, hadoopConf)
```
zsxwing, line 122: ditto: close

standalone/src/main/scala/io/delta/standalone/internal/util/Implicits.scala
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * Convert the [[CloseableIterator]] (Java) to an in-memory [[List]] (Scala).
+     *
+     * [[List]] is used over [[Seq]] so that the full list is actually generated instead of loaded
+     * lazily, in which case `iter.close()` would be called before the Seq was actually generated.
+     */
+    def toAutoClosedList: List[T] = {
+      try {
+        iter.asScala.toList
```
zsxwing, line 34: Can we change this to `toArray`?  Array is preferred for fixed length sequence.

standalone/src/main/scala/io/delta/standalone/internal/util/Implicits.scala
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+package io.delta.standalone.internal.util
+
+import io.delta.standalone.data.CloseableIterator
+
+object Implicits {
```
zsxwing, line 22: nit: `private[internal]`

standalone/src/main/scala/io/delta/standalone/internal/util/Implicits.scala
```diff
@@ -0,0 +1,40 @@
...
+  implicit class CloseableIteratorOps[T](private val iter: CloseableIterator[T]) {
+    import scala.collection.JavaConverters._
+
+    /**
+     * Convert the [[CloseableIterator]] (Java) to an in-memory [[List]] (Scala).
+     *
+     * [[List]] is used over [[Seq]] so that the full list is actually generated instead of loaded
```
zsxwing, line 29: The reason is `Iterator.toSeq` is lazy not `Seq` is lazy. `Seq` doesn't say anything about whether it's lazy or not :)

standalone/src/main/scala/io/delta/standalone/internal/util/Implicits.scala
```diff
@@ -0,0 +1,40 @@
...
+package io.delta.standalone.internal.util
+
+import io.delta.standalone.data.CloseableIterator
+
+object Implicits {
+  implicit class CloseableIteratorOps[T](private val iter: CloseableIterator[T]) {
+    import scala.collection.JavaConverters._
+
+    /**
+     * Convert the [[CloseableIterator]] (Java) to an in-memory [[List]] (Scala).
```
zsxwing, line 27: Could you also update the doc accordingly?

standalone/src/main/scala/io/delta/standalone/internal/util/Implicits.scala
```diff
@@ -0,0 +1,40 @@
...
+     *
+     * [[List]] is used over [[Seq]] so that the full list is actually generated instead of loaded
+     * lazily, in which case `iter.close()` would be called before the Seq was actually generated.
+     */
+    def toAutoClosedList: List[T] = {
```
zsxwing, line 32: nit: better to just call `toSeq`. We are not returning `AutoClosedList`. The accurate method name is `toSeqAndCloseIterator` but that's too long...

standalone/src/main/scala/io/delta/standalone/internal/util/Implicits.scala
scottsand-db: No we can't. If we use `iter.asScala.toArray` ...

1. With a return type of `Seq[T]`, we get:
```
error: polymorphic expression cannot be instantiated to expected type;
```

2. With a return type of `Array[T]`, we get
```
error: No ClassTag available for T
```

standalone/src/main/scala/io/delta/standalone/internal/util/Implicits.scala
scottsand-db: would `toArray` be better?

PR #160, opened 2021-09-29
[DSW] [14] OptimisticTransaction.markFilesAsRead returns a DeltaScan

scottsand-db:
- `OptimisticTransaction.markFilesAsRead` now returns a `DeltaScan`
- We now have two different DeltaScan implementations. It was a bad code smell having so many optional arguments in the original `DeltaScanImpl` constructor.
- `BaseDeltaScanImpl` is the basic version. it defines an `accept` method, iterates over files and returns only those that are `accept`ed, and only has 1 argument in its constructor
- `FilteredDeltaScanImpl` has 4 mandatory arguments in its constructor, overrides the `accept` method, and properly returns the pushed / residual predicates


standalone/src/main/scala/io/delta/standalone/internal/util/PredicateUtils.scala
```diff
@@ -0,0 +1,67 @@
...
+   * Check if condition can be evaluated using only metadata (i.e. partition columns)
+   */
+  def isPredicateMetadataOnly(condition: Expression, partitionColumns: Seq[String]): Boolean = {
+    // TODO: name equality resolver ?
+    condition.references().asScala.forall(partitionColumns.contains(_))
+  }
+
+  private def splitConjunctivePredicates(condition: Expression): Seq[Expression] = {
+    condition match {
+      case a: And => splitConjunctivePredicates(a.left) ++ splitConjunctivePredicates(a.right)
```
zsxwing, line 62: Can we use `children` so that we don't need to make `left` and `right` public?

zsxwing (commented):
Left one comment. Otherwise LGTM

standalone/src/main/scala/io/delta/standalone/internal/util/PredicateUtils.scala
scottsand-db: Hm. Would getters `getLeft`, `getRight` be better? If we expect users to convert our Expressions to their own ExternalExpressions, being able to easily extract the left, right from any BinaryExpression would be useful to convert to their own ExternalBinaryExpression. What do you think?

zsxwing: This is also fine.

standalone/src/main/java/io/delta/standalone/expressions/BinaryExpression.java
```diff
@@ -35,4 +44,18 @@ public final Object eval(RowRecord record) {
     public List<Expression> children() {
         return Arrays.asList(left, right);
     }
+
+    @Override
+    public boolean equals(Object o) {
```
zsxwing, line 49: `BinaryExpression` doesn't know how to implement this without knowing the details of subclasses. This should be done by the subclasses instead.

standalone/src/main/java/io/delta/standalone/expressions/BinaryExpression.java
```diff
@@ -35,4 +44,18 @@ public final Object eval(RowRecord record) {
...
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        BinaryExpression that = (BinaryExpression) o;
+        return Objects.equals(left, that.left) &&
+            Objects.equals(right, that.right);
+    }
+
+    @Override
+    public int hashCode() {
```
zsxwing, line 58: ditto

standalone/src/main/java/io/delta/standalone/expressions/UnaryExpression.java
```diff
@@ -10,7 +10,7 @@
  * if the input is evaluated to null.
  */
 public abstract class UnaryExpression implements Expression {
-    protected final Expression child;
+    public final Expression child;
```
zsxwing, line 13: nit: add `getChild` method instead.

zsxwing (changes requested):

standalone/src/main/java/io/delta/standalone/expressions/UnaryExpression.java
scottsand-db: forgot this, thanks

standalone/src/main/java/io/delta/standalone/expressions/BinaryExpression.java
scottsand-db: Posting for visibility. I responded in Slack:
<<<
BinaryExpression having equals and hashcode seems fine to me.
it will recursively call left.equals(that.left) all the way down to the leaf expressions.
I've then added to LeafExpression.java:
```
public abstract boolean equals(Object o);
public abstract int hashCode();
```
this means all leaves will override those methods

for example, And, Or, LessThan etc (all BinaryExpressions) would all have the exact same equals and hashcode methods

zsxwing: My concern is if someone extends `BinaryExpression` but forgets to override the `equals` and `hashCode` methods for their new field, it would be hard to debug.

Of course, It's painful in Java as it doesn't have Scala case class 😞

scottsand-db: but isn't that precisely why having the equals, hashCode methods in BinaryExpression makes sense?
And, Or, LessThan, GreaterThan etc etc do **not** need to implement those methods, since BinaryExpression does.
This means that if someone extends it, they also do not need to implement those methods. So there is nothing to forget.

Only LeafExpression will define those methods as abstract, so only leaf children need to define them (i.e. Column, Literal).

zsxwing: > Only LeafExpression will define those methods as abstract, so only leaf children need to define them (i.e. Column, Literal).

E.g., the user can define their own `BinaryExpression` such as `Comparator(left, right, type) // type can be '<', '>'...)`. Another example, is `CAST` which have extra fields to compare.

By the way, these two methods cannot be defined as abstract as they have been defined in `Object`.

scottsand-db: sgtm

scottsand-db: btw extending from an abstract super class for the `equals`, `hashcode` methods is fine by the compiler.

zsxwing (approved):
Discussed offline. LGTM

standalone/src/main/java/io/delta/standalone/expressions/BinaryExpression.java
scottsand-db: as discussed offline, we will go ahead with keeping the equals, hashcode in the BinaryExpression.

PR #165, opened 2021-10-06
Fix jackson incompatibility issue for Delta Hive 3 connector

zsxwing:
Found some jackson incompatibility issue when testing on an Amazon EMR 6.4.0. This PR fixes it by 

- Upgrade jackson-module-scala version
- Upgrade json4s version
- Shade jackson-module-scala to fix the incompatibility issue


zsxwing:
@YannByron could you take a look?

PR #175, opened 2021-10-13
[WIP] Flink Delta Sink

pkubit-g:


flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSink.java
```diff
@@ -0,0 +1,400 @@
...
+     * added to the committables information and incremented on every precommit action (after generating the committables).
+     *
+     * @param appId unique identifier for the current application
+     * @return last committed version for the provided appId
+     */
+    private long restoreOrGetNextCheckpointId(String appId) {
+        DeltaLog deltaLog = DeltaLog.forTable(this.bucketsBuilder.configuration.conf(), this.bucketsBuilder.basePath.getPath());
+        long lastCommittedCheckpointId = deltaLog.startTransaction().txnVersion(appId);
+        if (lastCommittedCheckpointId < 0) {
+            return 1;
```
scottsand-db, line 139: we want to "start" at 1, right? what if `lastCommittedCheckpointId <= 0`

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaCommitter.java
```diff
@@ -0,0 +1,47 @@
...
+    public DeltaCommitter(BucketWriter<?, ?> bucketWriter) {
+        this.bucketWriter = checkNotNull(bucketWriter);
+    }
+
+    @Override
+    public List<DeltaCommittable> commit(List<DeltaCommittable> committables) throws IOException {
+        for (DeltaCommittable committable : committables) {
+            if (committable.hasDeltaPendingFile()) {
+                // We should always use commitAfterRecovery which contains additional checks.
+                assert committable.getDeltaPendingFile() != null;
```
scottsand-db, line 28: do we need this check? `committable.hasDeltaPendingFile()` returns true if and only if `getDeltaPendingFile != null`

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -0,0 +1,188 @@
...
+        return new DeltaGlobalCommittable(committables);
+    }
+
+    private String resolveAppId(List<DeltaGlobalCommittable> globalCommittables) {
+        for (DeltaGlobalCommittable globalCommittable : globalCommittables) {
+            for (DeltaCommittable deltaCommittable : globalCommittable.getDeltaCommittables()) {
+                return deltaCommittable.getAppId();
+            }
+        }
+        return null;
```
scottsand-db, line 77: should this ever be the case? shouldn't we fail if this happens?

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -0,0 +1,188 @@
...
+
+    @Override
+    public List<DeltaGlobalCommittable> commit(List<DeltaGlobalCommittable> globalCommittables) {
+        if (!globalCommittables.isEmpty()) {
+            String appId = resolveAppId(globalCommittables);
+
+            Map<Long, List<Action>> addFileActionsPerCheckpoint = prepareAddFileActionsPerCheckpoint(globalCommittables);
+
+            for (long checkpointId : addFileActionsPerCheckpoint.keySet().stream().sorted().collect(Collectors.toList())) {
+                DeltaLog deltaLog = DeltaLog.forTable(conf, basePath.getPath());
```
scottsand-db, line 88: nit: don't need to create a new DeltaLog each time. this can be outside the for loop. `deltaLog.startTransaction()` will update to the latest state of the log.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -0,0 +1,188 @@
...
+                    List<Action> actions = new ArrayList<>();
+
+                    SetTransaction setTransaction = new SetTransaction(appId, checkpointId, Optional.of(System.currentTimeMillis()));
+                    actions.add(setTransaction);
+                    List<Action> addFileActions = addFileActionsPerCheckpoint.get(checkpointId);
+                    actions.addAll(addFileActions);
+
+                    Operation operation = prepareDeltaLogOperation(globalCommittables, addFileActions.size());
+                    StructType structType = deltaLog.snapshot().getMetadata().getSchema();
+                    // TODO add schema validation / comparison using org.apache.flink.connector.delta.sink.SchemaConverter
```
scottsand-db, line 102: I see. so you will compare the current table schema against the incoming data schema.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -0,0 +1,188 @@
...
+                if (deltaCommittable.hasDeltaPendingFile()) {
+                    DeltaPendingFile deltaPendingFile = deltaCommittable.getDeltaPendingFile();
+                    cumulatedRecordCount += deltaPendingFile.getRecordCount();
+                    cumulatedSize += deltaPendingFile.getFileSize();
+                }
+            }
+        }
+
+        Map<String, String> operationMetrics = new HashMap<>();
+        operationMetrics.put("numRemovedFiles", "0"); // TODO
```
scottsand-db, line 161: are there any DSW APIs we could add that would make this easier for you? e.g. what if we exposed some enum like
`Operations.Metrics.STREAMING_UPDATE.numAddedFiles`
`Operations.Metrics.STREAMING_UPDATE.numRemovedFiles`

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/writer/DeltaWriterBucket.java
```diff
@@ -0,0 +1,375 @@
...
+        long inProgressFileCreationTime = Long.MAX_VALUE;
+        String inProgressPartFileName = null;
+        long recordCount = 0;
+        long inProgressPartFileSize = -1L;
+        long lastUpdateTime = -1L;
+
+        if (deltaInProgressPart != null) {
+            InProgressFileWriter<IN, String> inProgressPart = deltaInProgressPart.getInProgressPart();
+            inProgressFileRecoverable = inProgressPart.persist();
+            inProgressFileToCleanup = inProgressFileRecoverable;
```
scottsand-db, line 228: nit: can you move this line `inProgressFileToCleanup = inProgressFileRecoverable;` to below `lastUpdateTime = ` with a blank line in between.

`inProgressFileToCleanup = inProgressFileRecoverable;` is a side effect of this function. but all of the other lines in this if block are local-variable assignment. it would be nice if the side effect of this function was a bit more visible / separated below.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/writer/DeltaWriterBucket.java
```diff
@@ -0,0 +1,375 @@
...
+                inProgressFileRecoverable,
+                inProgressPartFileName,
+                recordCount,
+                inProgressPartFileSize,
+                lastUpdateTime,
+                appId
+        );
+    }
+
+    void onProcessingTime(long timestamp) throws IOException {
```
scottsand-db, line 249: what is this used for?

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/DeltaSinkITBatch.java
```diff
@@ -0,0 +1,205 @@
...
+            new RowType.RowField("surname", new VarCharType(VarCharType.MAX_LENGTH)),
+            new RowType.RowField("age", new IntType())
+    ));
+
+    String DELTA_TABLE_PATH;
+
+    @Before
+    public void setup() {
+        try {
+            DELTA_TABLE_PATH = TEMPORARY_FOLDER.newFolder().getAbsolutePath();
```
scottsand-db, line 84: is this auto-deleted / cleaned up after the test finishes?

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/DeltaSinkITStreaming.java
```diff
@@ -0,0 +1,346 @@
...
+import java.util.Objects;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.CountDownLatch;
+import java.util.stream.LongStream;
+
+
+@RunWith(Parameterized.class)
+public class DeltaSinkITStreaming extends StreamingExecutionFileSinkITCase {
```
scottsand-db, line 78: nit: in the future, I wonder if `DeltaSinkITStreaming` and `DeltaSinkITBatch` can extend some `DeltaSinkITBase` test suite that defines common variables and helper methods.

scottsand-db (commented):
Great work! Asked a few questions / left a few minor comments.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/writer/DeltaWriterBucket.java
pkubit-g: sure, makes sense !

PR #194, opened 2021-11-01
[DSW] Memory optimized iterator

scottsand-db:
This PR adds a proper memory-optimized, lazy iterator implementation for `DeltaScan.getFiles`.

### SnapshotImpl
SnapshotImpl is changed to load the Protocol and Metadata actions efficiently, too (instead of replaying the log and keeping all actions in memory).

We never used `numOfMetadata` or `numOfProtocol`, so we can remove `protocol`, `metadata`, `numOfMetadata`, `numOfProtocol` from Snapshot.State

### MemoryOptimizedLogReplay
Provides a `CloseableIterator[(Action, isFromCheckpoint)]` API, using one iterator for files and another iterator for action-reading inside the current file.

We use this reverse, closeable iterator in two places, and it is important that we close it in both.
1. in `loadTableProtocolAndMetadata`, we use a try finally to close it
2. in `DeltaScanImpl` we expose both `getFilesScala`, which also uses try finally to close it, and `getFiles` which exposes a wrapper over this closeable iterator, and it is up to the user to close it.

### DeltaScanImpl
Now uses the MemoryOptimizedLogReplay and keeps track of seenAddFiles and tombstones to return valid AddFiles

### DeltaScanSuite
Adds tests for FilteredDeltaScanImpl

### DeltaLogSuite
We have now renamed `DeltaLogSuite` to `DeltaLogSuiteBase`, and basically provided an abstract "snapshot get all files" API that children test suites have to override.

So, we have `StandardDeltaLogSuite` which uses `snapshot.getAllFiles` (in memory).
And we have `MemoryOptimizedDeltaLogSuite` which uses `snapshot.scan.getFiles` (memory optimized)

### DIFF
--- a/standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
@@ -295,7 +295,7 @@ private class InitialSnapshotImpl(
 
   override lazy val metadataScala: Metadata = Metadata()
 
-  override def scan(): DeltaScan = new DeltaScanImpl(null)
+  override def scan(): DeltaScan = new DeltaScanImpl(memoryOptimizedLogReplay)
 
   override def scan(predicate: Expression): DeltaScan =
     new FilteredDeltaScanImpl(memoryOptimizedLogReplay, predicate, metadataScala.partitionSchema)
--- a/standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
@@ -29,12 +29,12 @@ import io.delta.standalone.actions.{AddFile => AddFileJ, Metadata => MetadataJ, 
 import io.delta.standalone.data.{CloseableIterator, RowRecord => RowParquetRecordJ}
 import io.delta.standalone.expressions.Expression
 
-import io.delta.standalone.internal.actions.{AddFile, InMemoryLogReplay, Metadata, Parquet4sSingleActionWrapper, Protocol, RemoveFile, SetTransaction, SingleAction}
+import io.delta.standalone.internal.actions.{AddFile, InMemoryLogReplay, MemoryOptimizedLogReplay, Metadata, Parquet4sSingleActionWrapper, Protocol, RemoveFile, SetTransaction, SingleAction}
 import io.delta.standalone.internal.data.CloseableParquetDataIterator
 import io.delta.standalone.internal.exception.DeltaErrors
 import io.delta.standalone.internal.logging.Logging
 import io.delta.standalone.internal.scan.{DeltaScanImpl, FilteredDeltaScanImpl}
-import io.delta.standalone.internal.util.{ConversionUtils, FileNames, JsonUtils, MemoryOptimizedLogReplayUtil}
+import io.delta.standalone.internal.util.{ConversionUtils, FileNames, JsonUtils}
 
 /**
  * Scala implementation of Java interface [[Snapshot]].
@@ -86,6 +86,9 @@ private[internal] class SnapshotImpl(
   // Internal-Only Methods
   ///////////////////////////////////////////////////////////////////////////
 
+  val memoryOptimizedLogReplay =
+    new MemoryOptimizedLogReplay(files, deltaLog.store, hadoopConf, deltaLog.timezone)
+
   /**
    * Returns an implementation that provides an accessor to the files as internal Scala
    * [[AddFile]]s. This prevents us from having to replay the log internally, generate Scala
@@ -121,9 +124,7 @@ private[internal] class SnapshotImpl(
     var metadata: Metadata = null
     // We replay logs from newest to oldest and will stop when we find the latest Protocol and
     // metadata.
-    MemoryOptimizedLogReplayUtil.replayActionsReversely(
-      files, deltaLog.store, hadoopConf, deltaLog.timezone) { (action, _) =>
-
+    memoryOptimizedLogReplay.replayActionsReversely { (action, _) =>
       action match {
         case p: Protocol =>
           if (null == protocol) {
--- a/standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
@@ -122,12 +122,11 @@ private[internal] class SnapshotImpl(
   private def loadTableProtocolAndMetadata(): (Protocol, Metadata) = {
     var protocol: Protocol = null
     var metadata: Metadata = null
-
     val iter = memoryOptimizedLogReplay.getReverseIterator
 
     try {
       // We replay logs from newest to oldest and will stop when we find the latest Protocol and
-      // metadata.
+      // Metadata.
       iter.asScala.foreach { case (action, _) =>
         action match {
           case p: Protocol =>
@@ -155,7 +154,6 @@ private[internal] class SnapshotImpl(
     } finally {
       iter.close()
     }
-
 
     // Sanity check. Should not happen in any valid Delta logs.
     if (protocol == null) {
--- a/standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
@@ -34,7 +34,7 @@ import io.delta.standalone.internal.data.CloseableParquetDataIterator
 import io.delta.standalone.internal.exception.DeltaErrors
 import io.delta.standalone.internal.logging.Logging
 import io.delta.standalone.internal.scan.{DeltaScanImpl, FilteredDeltaScanImpl}
-import io.delta.standalone.internal.util.{ConversionUtils, FileNames, JsonUtils}
+import io.delta.standalone.internal.util.{ConversionUtils, FileNames, JsonUtils, MemoryOptimizedLogReplayUtil}
 
 /**
  * Scala implementation of Java interface [[Snapshot]].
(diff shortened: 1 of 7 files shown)


standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
```diff
@@ -92,27 +95,79 @@ private[internal] class SnapshotImpl(
...
+              }
+            }
+          case _ => // do nothing
+        }
+      }
+    } finally {
+      iter.close()
+    }
+
+
```
scottsand-db, line 159: delete empty line

standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
```diff
@@ -419,3 +430,31 @@ class DeltaLogSuite extends FunSuite {
...
+    new StandardSnapshot(snapshot)
+  }
+}
+
+class MemoryOptimizedDeltaLogSuite extends DeltaLogSuiteBase {
+  class MemoryOptimizedSnapshot(snapshot: Snapshot) extends CustomAddFilesAccessor(snapshot) {
+    override def _getFiles(): java.util.List[AddFileJ] = {
+      import io.delta.standalone.internal.util.Implicits._
+
+      snapshot.scan().getFiles.toArray.toList.asJava
```
allisonport-db, line 453: is the `toArray` necessary? (also close the iterator?)

standalone/src/test/scala/io/delta/standalone/internal/DeltaScanSuite.scala
```diff
@@ -41,43 +46,135 @@ class DeltaScanSuite extends FunSuite {
...
   private val dataConjunct = new EqualTo(schema.column("col3"), Literal.of(5))
 
-  test("properly splits metadata (pushed) and data (residual) predicates") {
-    val mixedConjunct = new LessThan(schema.column("col2"), schema.column("col4"))
-    val filter = new And(new And(metadataConjunct, dataConjunct), mixedConjunct)
-    val scan = new FilteredDeltaScanImpl(files, filter, partitionSchema)
+  def withLog(actions: Seq[Action])(test: DeltaLog => Unit): Unit = {
+    withTempDir { dir =>
+      val log = DeltaLog.forTable(new Configuration(), dir.getCanonicalPath)
+      log.startTransaction().commit(metadata :: Nil, op, "engineInfo")
```
allisonport-db, line 65: nit: `Operation(Operation.Name.CREATE_TABLE)`?

standalone/src/main/scala/io/delta/standalone/internal/scan/DeltaScanImpl.scala
```diff
@@ -16,77 +16,146 @@
...
+          case add: AddFile =>
+            val canonicalizeAdd = add.copy(
+              dataChange = false,
+              path = canonicalizePath(add.path, replay.hadoopConf))
+
+            val alreadyDeleted = tombstones.contains(canonicalizeAdd.pathAsUri)
+            val alreadyReturned = addFiles.contains(canonicalizeAdd.pathAsUri)
+
+            if (!alreadyReturned) {
+              addFiles += canonicalizeAdd.pathAsUri
```
zsxwing, line 97:
```
if (!isCheckpoint) {
  addFiles += canonicalizeAdd.pathAsUri
}
```
Checkpoint will do the AddFile dedup. So no files will appear twice in a checkpoint. In addition, we will not scan any json files before checkpoint. So it's safe to not store AddFiles from a checkpoint.

standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
```diff
@@ -92,27 +95,77 @@ private[internal] class SnapshotImpl(
...
-  def protocolScala: Protocol = state.protocol
-  def metadataScala: Metadata = state.metadata
   def numOfFiles: Long = state.numOfFiles
 
-  private def load(paths: Seq[Path]): Seq[SingleAction] = {
+  /** A map to look up transaction version by appId. */
+  lazy val transactions: Map[String, Long] =
+    setTransactionsScala.map(t => t.appId -> t.version).toMap
+
+  lazy val (protocolScala, metadataScala) = loadTableProtocolAndMetadata()
```
zsxwing, line 120: nit: `lazy` is redundant since we call `deltaLog.assertProtocolRead(protocolScala)` in the constructor.

standalone/src/test/scala/io/delta/standalone/internal/DeltaScanSuite.scala
scottsand-db: technically yes, but none of these tests actually check or use the operation after it is written :p

standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
scottsand-db: great question. `toArray` is coming from `io.delta.standalone.internal.util.Implicits.CloseableIteratorOps.toArray`, which closes the iter for us

standalone/src/main/scala/io/delta/standalone/internal/scan/DeltaScanImpl.scala
```diff
@@ -16,77 +16,146 @@
...
    * This is a utility method for internal use cases where we need the filtered files
    * as their Scala instances, instead of Java.
-   *
-   * Since this is for internal use, we can keep this as a [[Seq]].
    */
-  def getFilesScala: Seq[AddFile] = files.filter(accept)
+  def getFilesScala: Array[AddFile] = {
+    import io.delta.standalone.internal.util.Implicits._
+
+    getIterScala.toArray
```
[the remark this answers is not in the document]
scottsand-db: io.delta.standalone.internal.util.Implicits.CloseableIteratorOps.toArray :)

standalone/src/main/scala/io/delta/standalone/internal/scan/DeltaScanImpl.scala
```diff
@@ -16,77 +16,146 @@
...
+      }
+
+      // No next matching found
+      nextMatching = None
     }
 
     override def hasNext: Boolean = {
-      nextValid.isDefined
+      // nextMatching will be empty if
+      // a) this is the first time hasNext has been called
```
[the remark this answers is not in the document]
scottsand-db: great point!

standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
scottsand-db: this actually does need to be lazy (when it is not, 70 tests fail).
I added this comment to explain:
```
// These values need to be declared lazy. In Scala, strict values (i.e. non-lazy) in superclasses
  // (e.g. SnapshotImpl) are fully initialized before subclasses (e.g. InitialSnapshotImpl).
  // If these were 'strict', or 'eager', vals, then `loadTableProtocolAndMetadata` would be called
  // for all new InitialSnapshotImpl instances, causing an exception.
  lazy val (protocolScala, metadataScala) = loadTableProtocolAndMetadata()
```

standalone/src/main/scala/io/delta/standalone/internal/scan/DeltaScanImpl.scala
```diff
@@ -16,77 +16,146 @@
...
+        if (nextValid.isEmpty) {
+          nextMatching = None
+          return
+        } else if (accept(nextValid.get)) {
+          nextMatching = nextValid
+          return
+        }
+      }
+
+      // No next matching found
```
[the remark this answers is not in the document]
scottsand-db: Great catch, yup it's never reached. I'll use this.

standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
```diff
@@ -92,27 +95,77 @@ private[internal] class SnapshotImpl(
...
+    var protocol: Protocol = null
+    var metadata: Metadata = null
+    val iter = memoryOptimizedLogReplay.getReverseIterator
+
+    try {
+      // We replay logs from newest to oldest and will stop when we find the latest Protocol and
+      // Metadata.
+      iter.asScala.foreach { case (action, _) =>
+        action match {
+          case p: Protocol =>
```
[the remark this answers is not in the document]
scottsand-db: nice

standalone/src/main/scala/io/delta/standalone/internal/actions/MemoryOptimizedLogReplay.scala
```diff
@@ -0,0 +1,134 @@
...
+// Helper Classes
+///////////////////////////////////////////////////////////////////////////
+
+private class CustomJsonIterator(iter: CloseableIterator[String])
+  extends CloseableIterator[(Action, Boolean)] {
+
+  override def hasNext: Boolean = iter.hasNext
+
+  override def next(): (Action, Boolean) = {
+    (JsonUtils.mapper.readValue[SingleAction](iter.next).unwrap, false)
```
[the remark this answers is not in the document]
scottsand-db: nice

standalone/src/main/scala/io/delta/standalone/internal/actions/MemoryOptimizedLogReplay.scala
```diff
@@ -0,0 +1,134 @@
...
+   */
+  def getReverseIterator: CloseableIterator[(Action, Boolean)] =
+    new CloseableIterator[(Action, Boolean)] {
+      private val reverseFilesIter: Iterator[Path] = files.sortWith(_.getName > _.getName).iterator
+      private var actionIter: Option[CloseableIterator[(Action, Boolean)]] = None
+
+      /**
+       * If the current `actionIter` has no more elements, this function repeatedly reads the next
+       * file, if it exists, and creates the next `actionIter` until we find a non-empty file.
+       */
```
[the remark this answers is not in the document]
scottsand-db: sgtm. does scala prefer tail recursion over iteration as a paradigm? i'd prefer a while loop here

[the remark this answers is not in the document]
allisonport-db: I'm not sure. Just was stating that since it'd be tail recursive it's not immediately worse than using a loop. A while loop might be more readable?

zsxwing (approved):
LGTM

PR #197, opened 2021-11-04
Fix RowRecord to allow read partition values

scottsand-db:
This PR fixes a lacking feature in the RowRecord implementation where partition (metadata) values were not included and able to be read, only data values were.


standalone/src/main/scala/io/delta/standalone/internal/data/CloseableParquetDataIterator.scala
```diff
@@ -116,12 +126,53 @@ private[internal] case class CloseableParquetDataIterator(
...
+    if (null != nextPartitionVals) {
+      nextPartitionVals.foreach { case (fieldName, value) =>
+        if (value == null || "null" == value) {
+          partitionValues += (fieldName -> null)
+        } else {
+          val schemaField = schema.get(fieldName)
+          if (schemaField != null) {
+            val decodedFieldValue = decodePartition(schemaField.getDataType, value)
+            partitionValues += (fieldName -> decodedFieldValue)
+          }
```
zsxwing, line 147: it's better to throw an exception for the invalid state (`schemaField==null`).

standalone/src/main/scala/io/delta/standalone/internal/data/CloseableParquetDataIterator.scala
```diff
@@ -116,12 +126,53 @@ private[internal] case class CloseableParquetDataIterator(
...
+   * @return the iterable for the next data file in `dataFilePathsAndPartitionsIter`, not null
    */
   private def readNextFile: ParquetIterable[RowParquetRecord] = {
+    val (nextDataFilePath, nextPartitionVals) = dataFilePathsAndPartitionsIter.next()
+
+    partitionValues = Map()
+
+    if (null != nextPartitionVals) {
+      nextPartitionVals.foreach { case (fieldName, value) =>
+        if (value == null || "null" == value) {
```
zsxwing, line 140: I missed this in my previous review. `"null"` should be a string `"null"`.
```
if (value == null) {
  partitionValues += (fieldName -> null)
}
```
Could you update the test to test the string `"null"` case?

standalone/src/main/scala/io/delta/standalone/internal/data/RowParquetRecordImpl.scala
```diff
@@ -82,9 +84,18 @@ private[internal] case class RowParquetRecordImpl(
 
   override def getSchema: StructType = schema
 
-  override def getLength: Int = record.length
+  override def getLength: Int = record.length + partitionValues.size
 
-  override def isNullAt(fieldName: String): Boolean = record.get(fieldName) == NullValue
+  override def isNullAt(fieldName: String): Boolean = {
+    if (partitionValues.contains(fieldName)) { // is partition field
+        if (partitionValues(fieldName) == null) {
```
zsxwing, line 91: nit: `return partitionValues(fieldName) == null`

standalone/src/main/scala/io/delta/standalone/internal/data/RowParquetRecordImpl.scala
```diff
@@ -133,6 +144,11 @@ private[internal] case class RowParquetRecordImpl(
    */
   private def getAs[T](fieldName: String): T = {
     val schemaField = schema.get(fieldName)
+
+    if (partitionValues.contains(fieldName)) { // partition field
+      if (partitionValues(fieldName) == null) return null.asInstanceOf[T]
```
zsxwing, line 149: do we really need this line?

zsxwing (commented):
Left some minor comments

standalone/src/main/scala/io/delta/standalone/internal/data/RowParquetRecordImpl.scala
scottsand-db: good catch sorry

zsxwing (commented):
LGTM. Manually merged: https://github.com/delta-io/connectors/commit/117e7820467f2c5b801373974e10e3e6fb77eb5c

Issue #96, opened 2021-06-17
Illegal Access Error while querying  the created hive table through deltalake connector


bkumaran08:
Hive version :  2.6
delta lake connector version : delta-hive-assembly_2.11-0.2.0.jar
Spark version : 3.1.1
Scala version :2.12
delta-core : 1.0.0

I have created a deltalake table using spark, and then I am able to create a hive table out of it using delta lake connector.

When we tried to access the created table in hive, it is throwing below error.

Delta lake table is working fine when I query it through spark.

**Error Trace** 

Exception in thread "main" java.lang.IllegalAccessError: tried to access method org.apache.hadoop.hive.ql.io.parquet.read.DataWritableReadSupport.getColumnNames(Ljava/lang/String;)Ljava/util/List; from class io.delta.hive.DeltaStorageHandler
        at io.delta.hive.DeltaStorageHandler.configureInputJobProperties(DeltaStorageHandler.scala:71)
        at org.apache.hadoop.hive.ql.plan.PlanUtils.configureJobPropertiesForStorageHandler(PlanUtils.java:820)
        at org.apache.hadoop.hive.ql.plan.PlanUtils.configureInputJobPropertiesForStorageHandler(PlanUtils.java:790)
        at org.apache.hadoop.hive.ql.optimizer.SimpleFetchOptimizer$FetchData.convertToWork(SimpleFetchOptimizer.java:385)
        at org.apache.hadoop.hive.ql.optimizer.SimpleFetchOptimizer$FetchData.access$000(SimpleFetchOptimizer.java:323)
        at org.apache.hadoop.hive.ql.optimizer.SimpleFetchOptimizer.optimize(SimpleFetchOptimizer.java:134)
        at org.apache.hadoop.hive.ql.optimizer.SimpleFetchOptimizer.transform(SimpleFetchOptimizer.java:105)
        at org.apache.hadoop.hive.ql.optimizer.Optimizer.optimize(Optimizer.java:215)
        at org.apache.hadoop.hive.ql.parse.SemanticAnalyzer.analyzeInternal(SemanticAnalyzer.java:10572)
        at org.apache.hadoop.hive.ql.parse.CalcitePlanner.analyzeInternal(CalcitePlanner.java:219)
        at org.apache.hadoop.hive.ql.parse.BaseSemanticAnalyzer.analyze(BaseSemanticAnalyzer.java:238)
        at org.apache.hadoop.hive.ql.Driver.compile(Driver.java:474)
[...]

PR #198, opened 2021-11-05
Make hive connector support both hive 2 and 3

zsxwing:
Since we share the jackson libraries in standalone jar, it's easy to make hive connector support both Hive 2 and 3. Here are the changes in the PR:

- Make hive connector depend on the standalone jar that has the shaded jackson libraries.
- Create a new hiveAssembly project to shade the hive connector.
- Create a new hiveTest project and move existing code in hive-test/src/test to this project.
- hiveAssembly and hiveTest will be built using hive 3.

However, to make sure the assembly hive connector work with hive 2. We also create two new projects: hive2MR and hive2Tez. They will just depend on the jars created by `hiveAssembly` and `hiveTest` (We need the common test codes).

After this change, hive connector tests will also help us test the production standalone jar we will release.


.github/workflows/test.yaml
```diff
@@ -18,18 +18,20 @@ jobs:
             ~/.sbt
             ~/.ivy2
             ~/.cache/coursier
-          key: delta-sbt-cache
+          key: delta-sbt-cache-hive23
```
zsxwing, line 21: change the key so that GitHub Actions will cache the new dependencies. It will cache the above directories only if the cache key doesn't exist.

.github/workflows/test.yaml
```diff
@@ -18,18 +18,20 @@ jobs:
...
       - name: Run Scala/Java and Python tests
         shell: bash -l {0}
         run: |
           build/sbt "++ 2.12.8 standalone/test"
-          build/sbt "++ 2.12.8 hive/test"
```
zsxwing, line 26: hive/test has no tests now

build.sbt
```diff
@@ -289,57 +372,16 @@ lazy val standaloneCosmetic = project
...
-            </dependencies>
-          case _ => n
-        }
-      }
-      new RuleTransformer(hardcodeDeps).transform(node).head
-    },
     exportJars := true,
-    packageBin in Compile := (assembly in standalone).value
+    packageBin in Compile := (assembly in standalone).value,
+    libraryDependencies ++= Seq(
```
zsxwing, line 377: Found adding `libraryDependencies` can avoid the hard code `dependencies` xml.

hive/src/main/scala/io/delta/hive/DeltaStorageHandler.scala
```diff
@@ -62,13 +61,23 @@ class DeltaStorageHandler extends DefaultStorageHandler with HiveMetaHook
 
   override def getSerDeClass(): Class[_ <: AbstractSerDe] = classOf[ParquetHiveSerDe]
 
+  /**
+   * DataWritableReadSupport.getColumnNames is private before Hive 2.3. Using reflection to make it
+   * compatible with versions before Hive 2.3.
+   */
+  private def getColumnNames(columns: String): java.util.List[String] = {
```
zsxwing, line 68: This fixes #96

hive2-mr/src/test/scala/io/delta/hive/HiveMRSuite.scala
```diff
@@ -0,0 +1,50 @@
...
+import java.io.{Closeable, File}
+
+import scala.collection.JavaConverters._
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.mapred.{JobConf, MiniMRCluster}
+import org.apache.hadoop.mapreduce.MRJobConfig
+import org.apache.hadoop.yarn.conf.YarnConfiguration
+
+class HiveMRSuite extends HiveConnectorTest {
```
zsxwing, line 28: Duplicate the file in hive-mr and hive2-mr. This is fine since we usually don't need to change it. Alternatively, we can create a new sbt project to try sharing the code. But that's too heavy.

hive2-tez/src/test/scala/io/delta/hive/HiveTezSuite.scala
```diff
@@ -0,0 +1,112 @@
...
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.hive.conf.HiveConf
+import org.apache.hadoop.mapred.JobConf
+import org.apache.hadoop.mapreduce.MRJobConfig
+import org.apache.hadoop.yarn.conf.YarnConfiguration
+import org.apache.tez.dag.api.TezConfiguration
+import org.apache.tez.runtime.library.api.TezRuntimeConfiguration
+import org.apache.tez.test.MiniTezCluster
+
+class HiveTezSuite extends HiveConnectorTest {
```
zsxwing, line 32: ditto

build.sbt
scottsand-db: nice!

build.sbt
```diff
@@ -513,7 +555,9 @@ lazy val sqlDeltaImport = (project in file("sql-delta-import"))
     publishArtifact := scalaBinaryVersion.value == "2.12",
     publishArtifact in Test := false,
     libraryDependencies ++= Seq(
-      "org.apache.spark" %% "spark-sql" % sparkVersion % "provided",
+      // We don't need to publish this project in Scala 2.11. Define Spark and Delta versions with
```
scottsand-db, line 558: but we aren't 'publishing' it in Scala 2.12 either? we are just using this for tests? can you clarify to me / explain

build.sbt
```diff
@@ -513,7 +555,9 @@ lazy val sqlDeltaImport = (project in file("sql-delta-import"))
...
+      // 2.12 so that running `build/sbt "++ 2.11.12 publishLocal"` doesn't fail.
```
scottsand-db, line 559: Do you mean`build/sbt "++ 2.11.12 standaloneCosmetic/publishLocal"` or "++ 2.11.12 testStandaloneCosmetic/publishLocal"` (I doubt the latter)? Can you specify

hive2-mr/src/test/scala/io/delta/hive/HiveMRSuite.scala
scottsand-db: Should we comment in the code this? That way we know if someone ever changes one file that they should change the other?

scottsand-db (approved):
LGTM. Added a couple minor comments / questions.

zsxwing:
> Do you mean`build/sbt "++ 2.11.12 standaloneCosmetic/publishLocal"` or "++ 2.11.12 testStandaloneCosmetic/publishLocal"` (I doubt the latter)? Can you specify

I meant `build/sbt "++ 2.11.12 publishLocal" which will publish all projects. We will use this to publish our projects. I updated the comment to clarify more.

YannByron:
LGTM.

PR #200, opened 2021-11-05
EngineInfo: replace spaces with '-' for engineVersion and deltaVersion

allisonport-db:
Should we also ask the user to format this way? Or is it okay to just replace spaces with dashes before saving?


zsxwing:
We can suggest the format in the doc, and also replace spaces with dashes.

zsxwing:
The change looks good to me. Could you write a quick test?

scottsand-db:
+1 to Ryan's comments above.

> We can suggest the format in the doc

This would be in OptimisticTransaction.java::commit btw

scottsand-db (approved):

zsxwing (approved):

PR #201, opened 2021-11-08
Add `getRemoveFile` API to AddFile

scottsand-db:
- add various `AddFile::remove` APIs and use those instead of `RemoveFile` constructor calls in tests
- we cannot make the `RemoveFile` constructor private or package-private, so I document that users should use of one the `AddFile::remove` APIs
- added missing @Nullable and @Nonnull annotations to classes along the way

### DIFF
--- a/standalone/src/main/java/io/delta/standalone/actions/AddFile.java
+++ b/standalone/src/main/java/io/delta/standalone/actions/AddFile.java
@@ -68,8 +68,7 @@ public final class AddFile implements FileAction {
      */
     @Nonnull
     public RemoveFile getRemoveFile() {
-        return new RemoveFile(path, Optional.of(System.currentTimeMillis()), dataChange, true,
-            partitionValues, size, tags);
+        return getRemoveFile(System.currentTimeMillis());
     }
 
     /**
--- a/standalone/src/main/java/io/delta/standalone/actions/AddFile.java
+++ b/standalone/src/main/java/io/delta/standalone/actions/AddFile.java
@@ -90,8 +90,8 @@ public final class AddFile implements FileAction {
      *         {@code dataChange} flag
      */
     @Nonnull
-    public RemoveFile remove(boolean _dataChange) {
-        return remove(System.currentTimeMillis(), _dataChange);
+    public RemoveFile remove(boolean dataChange) {
+        return remove(System.currentTimeMillis(), dataChange);
     }
 
     /**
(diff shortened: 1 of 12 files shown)


allisonport-db (approved):
LGTM. Is this an API we want to expose?

scottsand-db:
> LGTM. Is this an API we want to expose?

It's useful. For example, during an update operation, you want to remove all the files currently in a partition.
```
val currentAdds = txn.markFilesAsRead(someFilter)
val removes = currentAdds.map(_.getRemoveFile)
val newAdds = currentAdds.map(someChange)
txn.commit(removes ++ newAdds, ..., ...)
```

standalone/src/main/java/io/delta/standalone/actions/AddFile.java
```diff
@@ -62,6 +63,24 @@ public AddFile(@Nonnull String path, @Nonnull Map<String, String> partitionValue
         this.tags = tags;
     }
 
+    /**
+     * @return the corresponding {@link RemoveFile} for this file
+     */
+    @Nonnull
+    public RemoveFile getRemoveFile() {
```
zsxwing, line 70: It's better to call `remove`. We also need the `dataChange` parameter. It's needed when doing file compaction.

```
public RemoveFile remove()
public RemoveFile remove(long deletionTimestamp)
public RemoveFile remove(boolean dataChange)
public RemoveFile remove(long deletionTimestamp, boolean dataChange)
```

zsxwing (commented):
Good point. I missed this method when reviewing RemoveFile.

Using AddFile to create RemoveFile is a recommended way to create RemoveFile. In OSS Delta, we had some correctness bugs caused by not creating RemoveFile From AddFile (e.g, manually create RemoveFile but its path is not exactly the same as AddFile.path, so the file is not removed).

Can we try to make the constructor of `RemoveFile` not public, and always create `RemoveFile` from `AddFile` so that we can ensure `RemoveFile.path` always matches some `AddFile.path`? You may need to change a lot of tests, but I think it's worth since this would avoid a class of correctness issues.

standalone/src/main/java/io/delta/standalone/actions/AddFile.java
```diff
@@ -62,6 +68,42 @@ public AddFile(@Nonnull String path, @Nonnull Map<String, String> partitionValue
...
+    public RemoveFile remove(long deletionTimestamp) {
+        return remove(deletionTimestamp, dataChange);
+    }
+
+    /**
+     * @return the corresponding {@link RemoveFile} for this file, instantiated with the given
+     *         {@code dataChange} flag
+     */
+    @Nonnull
+    public RemoveFile remove(boolean _dataChange) {
```
zsxwing, line 93: nit: it's better to use `dataChange`. It will looks weird if people see `_dataChange` in the java doc.

zsxwing (approved):
LGTM except one nit of the `_dataChange` parameter name.

PR #204, opened 2021-11-10
Update package object path

allisonport-db:
Updated to follow https://docs.scala-lang.org/tour/package-objects.html


zsxwing (approved):
LGTM

scottsand-db:
@allisonport-db @zsxwing why was this merged into delta_standalone_writer_feature instead of master? are we still using that branch?

My understanding was that once we merged it into master, all future PRs will just go right into master.

zsxwing:
Good catch. I didn't notice this was merged into `delta_standalone_writer_feature`. @allisonport-db could you open a new PR against master?

PR #208, opened 2021-11-12
[#124] Fix for getBigDecimal decoding bug with parquet4s LongValue

### LINKED ISSUE #124: call getBigDecimal(fieldname) when decimal(10,0) column with a value 1000000
an exception:
```
java.lang.RuntimeException: Unknown decimal decode type LongValue(1000000)
```
the fix advice:
```
diff --git a/standalone/src/main/scala/io/delta/standalone/internal/data/RowParquetRecordImpl.scala b/standalone/src/main/scala/io/delta/standalone/internal/data/RowParquetRecordImpl.scala
index 1847a76..d1fb062 100644
--- a/standalone/src/main/scala/io/delta/standalone/internal/data/RowParquetRecordImpl.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/data/RowParquetRecordImpl.scala
@@ -232,6 +232,7 @@ private[internal] case class RowParquetRecordImpl(
           case DoubleValue(double) => BigDecimal.decimal(double).bigDecimal
           case FloatValue(float) => BigDecimal.decimal(float).bigDecimal
           case BinaryValue(binary) => Decimals.decimalFromBinary(binary).bigDecimal
+          case LongValue(long) => new java.math.BigDecimal(long)
           case _ => throw new RuntimeException(s"Unknown decimal decode type $value")
         }
       }
```

allisonport-db:
Resolves #124 
• Added `Parquet4s.LongValue` to `customDecimalCodec`
• Added test case to `DeltaDataReaderSuite`
• Created table for test in `GoldenTables.scala`


zsxwing:
Could you add the code to generate parquet files into GoldenTables similar to https://github.com/delta-io/connectors/blob/master/golden-tables/src/test/scala/io/delta/golden/GoldenTables.scala#L728

standalone/src/test/scala/io/delta/standalone/internal/DeltaDataReaderSuite.scala
```diff
@@ -413,6 +413,15 @@ class DeltaDataReaderSuite extends FunSuite {
     assert(struct == DataTypeParser.fromJson(struct.toJson()))
   }
 
+  test("#124: getBigDecimal decode correctly for LongValue") {
```
scottsand-db, line 416: does this test fail without `case LongValue(long) => new java.math.BigDecimal(long)` above?

allisonport-db: yup, I checked

allisonport-db: It seems Decimal(1, 0) is decoded as IntValue and Decimal(10, 0) is decoded as LongValue. Didn't look much more specifically than that.

scottsand-db: Yup that makes sense. Thanks!

scottsand-db (approved):
LGTM

zsxwing (approved):
LGTM

PR #211, opened 2021-11-15
Time travel bug fix

scottsand-db:
This PR adds a basic fix for the case when the delta log consists of
```
N.checkpoint.parquet // no N.json
N+1.json
...
```
and users try to time travel to version N. Previously, this would trigger `deltas.last` in `SnapshotManagement:: getLogSegmentForVersion` to trigger a `no next element exception` error. Now, this is fixed.

Future work would be to check the similar case except when doing time travel using timestamp instead of version.


standalone/src/main/scala/io/delta/standalone/internal/DeltaHistoryManager.scala
```diff
@@ -128,7 +128,7 @@ private[internal] case class DeltaHistoryManager(deltaLog: DeltaLogImpl) extends
 
         // Note that we also check this condition at the end of the function - we check it
         // here too to to try and avoid more file listing when it's unnecessary.
-        if (lastCompleteCheckpoint.exists(_ >= smallestDeltaVersion)) {
+        if (lastCompleteCheckpoint.exists(_ >= smallestDeltaVersion - 1)) {
```
scottsand-db, line 131: this - 1 should be here. we forgot to copy it over from Delta OSS.

standalone/src/main/scala/io/delta/standalone/internal/Checkpoints.scala
```diff
@@ -53,7 +53,8 @@ private[internal] case class CheckpointMetaData(
  */
 private[internal] case class CheckpointInstance(
     version: Long,
-    numParts: Option[Int]) extends Ordered[CheckpointInstance] {
+    numParts: Option[Int],
+    modificationTime: Option[Long] = None) extends Ordered[CheckpointInstance] {
```
zsxwing, line 57: What do we do with a checkpoint modification time? A snapshot at version X should always use the X.json modification time. Otherwise, Two snapshots at version X may have two different timestamps. For example, if the snapshot X loads from checkpoint X, it will use checkpoint X's modification time; if it loads from checkpoint A, json A + 1, ... json X, it will use  json X's modification time.

scottsand-db: For a given checkpoint `N.checkpoint.parquet`, isn't it written just after `N.json` is written?
We only use this checkpoint modification time for the rare, rare edge case of when
- `N.checkpoint.parquet` exists
- `N.json` doesn't exist
- N is the earliest possible table version to load

Previously, in this case, we were getting a `no next element exception` on the line `deltas.last.getModificationTime`

Now it has been changed to:

```
val lastCommitTimestamp = if (deltas.nonEmpty) {
        deltas.last.getModificationTime
      } else {
        newCheckpoint.get.modificationTime.getOrElse(
          throw new IllegalStateException("CheckpointInstance modification time was not recorded")
        )
      }
```

If using the checkpoint modification time isn't valid, that's okay, and we can scrap this PR.

scottsand-db: Btw,
> A snapshot at version X should always use the X.json modification time

seems to imply that we can never delete N.json if N.checkpoint is the earliest version. Doesn't this violate the delta protocol?

zsxwing: > Previously, in this case, we were getting a `no next element exception` on the line `deltas.last.getModificationTime`

I understand the issue. The fix we should do is to make sure the log clean up should always keep N.json when N.checkpoint.parquet exists.

> seems to imply that we can never delete N.json if N.checkpoint is the earliest version. Doesn't this violate the delta protocol?

I think it's not documented today. Maybe I missed something today.

> For a given checkpoint `N.checkpoint.parquet`, isn't it written just after `N.json` is written?

No. A checkpoint can also be generated offline manually. We don't say anything about when a checkpoint should be generated in the protocol. Even it it's written just after `N.json`, the timestamp differences can be minutes or even hours depending on how fast the checkpoint files can be written.

zsxwing: > It's tempting then just to store the N.json timestamp into the N.parquet, perhaps in the metadata

That's a good idea. However, this is hard to be backward compatible.

scottsand-db:
Closing since this isn't the right way to fix this issue.

PR #196, opened 2021-11-03
Flink Delta Sink - PR 1

pkubit-g:
Add base classes and interfaces for Flink Delta Sink

PR #212, opened 2021-11-17
Fix Github actions to fail when an sbt test fails

allisonport-db:
A few issues fixed:
- Github actions wasn't reporting any failed tests unless the last test on the list failed, since it restarted `build/sbt` each time, and Github actions only considers the final exit code.
- Moved `compatibility/test` to only run on Scala 2.12 since it uses Delta 1.0.0
-  Fixed a few JavaDoc compilation errors that were merged in #196.
   - [QUESTION] Documentation in #196 uses `@implNote` Javadoc tag. I added it to the Javadoc compilation, but should we just change that doc instead?
   - [NOTE] Seems related to https://github.com/sbt/sbt/issues/875, but when there is a compilation error, all the other warnings are reported as errors. For example in the previous commit, it showed 18 `[error]` messages but also: 
```
[info] 2 errors
[info] 18 warnings
[error] (standalone/javaunidoc:doc) javadoc returned nonzero exit code
```
- Fixed the other tests that were failing

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -24,11 +24,21 @@ jobs:
           key: delta-sbt-cache-hive23-scala${{ matrix.scala }}
       - name: Run sqlDeltaImport tests
         shell: bash
-        run: build/sbt "++ ${{ matrix.scala }}" sqlDeltaImport/test compatibility/test
+        run: build/sbt "++ ${{ matrix.scala }}" sqlDeltaImport/test
         if: matrix.scala != '2.11.12'
-      - name: Run SBT tests
+      - name: Run testScalastyle
         shell: bash
-        run: >
-          build/sbt "++ ${{ matrix.scala }}" testScalastyle standalone/test
-          testStandaloneCosmetic/test hiveMR/test hiveTez/test hive2MR/test
-          hive2Tez/test flinkConnector/test
+        run: build/sbt "++ ${{ matrix.scala }}" testScalaStyle
+      - name: Run standalone and testStandaloneCosmetic tests
+        shell: bash
+        run: build/sbt "++ ${{ matrix.scala }}" standalone/test testStandaloneCosmetic/test
+      - name: Run compatibility tests
+        shell: bash
+        run: build/sbt "++ ${{ matrix.scala }}" compatibility/test
+        if: matrix.scala != '2.11.12'
+      - name: Run hive tests
+        shell: bash
+        run: build/sbt "++ ${{ matrix.scala }}" hiveMR/test hiveTez/test hive2MR/test hive2Tez/test
+      - name: Run flink tests
+        shell: bash
+        run: build/sbt "++ ${{ matrix.scala }}" flinkConnector/test
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -3,7 +3,7 @@ on: [push, pull_request]
 jobs:
   build:
     name: "Run tests "
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-18.04
     strategy:
       matrix:
         scala: [2.12.8, 2.11.12]
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -24,13 +24,5 @@ jobs:
       - name: Run compatibility tests
         run: build/sbt "++ ${{ matrix.scala }}" compatibility/test
         if: matrix.scala != '2.11.12'
-      - name: Run hive 3 mr tests
-        run: build/sbt "++ ${{ matrix.scala }}" hiveMR/test
-      - name: Run hive 3 tez tests
-        run: build/sbt "++ ${{ matrix.scala }}" hiveTez/test
-      - name: Run hive 2 mr tests
-        run: build/sbt "++ ${{ matrix.scala }}" hive2MR/test
-      - name: Run hive 2 tez tests
-        run: build/sbt "++ ${{ matrix.scala }}" hive2Tez/test
       - name: Run flink tests
         run: build/sbt "++ ${{ matrix.scala }}" flinkConnector/test
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -28,7 +28,7 @@ jobs:
         if: matrix.scala != '2.11.12'
       - name: Run testScalastyle
         shell: bash
-        run: build/sbt "++ ${{ matrix.scala }}" testScalaStyle
+        run: build/sbt "++ ${{ matrix.scala }}" testScalastyle
       - name: Run standalone and testStandaloneCosmetic tests
         shell: bash
         run: build/sbt "++ ${{ matrix.scala }}" standalone/test testStandaloneCosmetic/test
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -14,14 +14,6 @@ jobs:
         with:
           distribution: 'zulu'
           java-version: '8'
-      - name: Cache Scala, SBT
-        uses: actions/cache@v2
-        with:
-          path: |
-            ~/.sbt
-            ~/.ivy2
-            ~/.cache/coursier
-          key: delta-sbt-cache-hive23-scala${{ matrix.scala }}
       - name: Run sqlDeltaImport tests
         run: build/sbt "++ ${{ matrix.scala }}" sqlDeltaImport/test
         if: matrix.scala != '2.11.12'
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yaml
```diff
@@ -29,12 +29,6 @@ jobs:
...
-          build/sbt "++ ${{ matrix.scala }} testScalastyle"
-          build/sbt "++ ${{ matrix.scala }} standalone/test"
-          build/sbt "++ ${{ matrix.scala }} testStandaloneCosmetic/test"
-          build/sbt "++ ${{ matrix.scala }} compatibility/test"
-          build/sbt "++ ${{ matrix.scala }} hiveMR/test"
-          build/sbt "++ ${{ matrix.scala }} hiveTez/test"
-          build/sbt "++ ${{ matrix.scala }} hive2MR/test"
-          build/sbt "++ ${{ matrix.scala }} hive2Tez/test"
-          build/sbt "++ ${{ matrix.scala }} flinkConnector/test"
+          build/sbt "++ ${{ matrix.scala }}" testScalastyle standalone/test
```
allisonport-db, line 32: Each could be a separate step also

.github/workflows/test.yaml
```diff
@@ -24,17 +24,11 @@ jobs:
           key: delta-sbt-cache-hive23-scala${{ matrix.scala }}
       - name: Run sqlDeltaImport tests
         shell: bash -l {0}
```
zsxwing, line 26: The real problem is this line. It doesn't use `-e` for `bash`. We can simply remove this line as `-e` is [the default behavior](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference):

> Fail-fast behavior using set -eo pipefail: Default for bash and built-in shell. It is also the default when you don't provide an option on non-Windows platforms.

There is another one below.

Let's also keep the rest of your changes as it's simpler.

.github/workflows/test.yaml
```diff
@@ -24,17 +24,11 @@ jobs:
...
-          build/sbt "++ ${{ matrix.scala }} compatibility/test"
-          build/sbt "++ ${{ matrix.scala }} hiveMR/test"
-          build/sbt "++ ${{ matrix.scala }} hiveTez/test"
-          build/sbt "++ ${{ matrix.scala }} hive2MR/test"
-          build/sbt "++ ${{ matrix.scala }} hive2Tez/test"
-          build/sbt "++ ${{ matrix.scala }} flinkConnector/test"
+        run: >
+          build/sbt "++ ${{ matrix.scala }}" testScalastyle standalone/test
+          testStandaloneCosmetic/test hiveMR/test hiveTez/test hive2MR/test
+          hive2Tez/test flinkConnector/test
```
zsxwing, line 34: nit: missing a new line

build.sbt
```diff
@@ -500,6 +500,7 @@ lazy val standalone = (project in file("standalone"))
       "-windowtitle", "Delta Standalone Reader " + version.value.replaceAll("-SNAPSHOT", "") + " JavaDoc",
       "-noqualifier", "java.lang",
       // `doclint` is disabled on Circle CI. Need to enable it manually to test our javadoc.
```
zsxwing, line 502: nit: remove this line since we don't use Circle CI any more.

zsxwing (approved):
Thanks for fixing this. Left some minor style comments

scottsand-db:
Ryan's comments SGTM. Thanks for investigating + fixing this!

.github/workflows/test.yaml
zsxwing: By the way, could you also submit a PR in delta-io/delta to remove this as well?

.github/workflows/test.yaml
zsxwing: Looks like we still need to split it into multiple steps otherwise we would exceed the memory limit: https://github.com/delta-io/connectors/runs/4243063137?check_suite_focus=true

.github/workflows/test.yaml
```diff
@@ -23,18 +23,22 @@ jobs:
             ~/.cache/coursier
           key: delta-sbt-cache-hive23-scala${{ matrix.scala }}
       - name: Run sqlDeltaImport tests
-        shell: bash -l {0}
-        run: build/sbt "++ ${{ matrix.scala }} sqlDeltaImport/test"
+        shell: bash
```
zsxwing, line 26: could you remove `shell: bash`?

.github/workflows/test.yaml
```diff
@@ -23,18 +23,16 @@ jobs:
...
-          build/sbt "++ ${{ matrix.scala }} testScalastyle"
-          build/sbt "++ ${{ matrix.scala }} standalone/test"
-          build/sbt "++ ${{ matrix.scala }} testStandaloneCosmetic/test"
-          build/sbt "++ ${{ matrix.scala }} compatibility/test"
-          build/sbt "++ ${{ matrix.scala }} hiveMR/test"
-          build/sbt "++ ${{ matrix.scala }} hiveTez/test"
-          build/sbt "++ ${{ matrix.scala }} hive2MR/test"
-          build/sbt "++ ${{ matrix.scala }} hive2Tez/test"
-          build/sbt "++ ${{ matrix.scala }} flinkConnector/test"
+      - name: Run testScalastyle
```
zsxwing, line 28: Looks like we need to go back the original one `run` approach. Different runs are running in parallel and we hit the memory limit of GitHub Actions...

allisonport-db: From my understanding, only the jobs are running in parallel, and not each step (those are run sequentially.) So the change we made shouldn't have affected that.

allisonport-db: We could use a macOS vm which has [twice as much ram ](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources)

scottsand-db: I'm a bit out of the loop. Were we running into RAM issues before? Why is this a problem now?

.github/workflows/test.yaml
zsxwing: @scottsand-db the change in delta-io/delta is not a release blocker. We can make a release without it.

.github/workflows/test.yaml
zsxwing: right now we put multiple SBT commands in one `build/sbt` command. SBT may keep some memory when a SBT command is finished. This is why we had one SBT command in one line.

allisonport-db: So if that is the problem, it should be fine broken down into steps. Do you think the hive tests in particular need to be separated into each their own `build/sbt` command? i.e.
```
run: |
    build/sbt hiveMR/test
    ....
    build/sbt hive2Tez/test
```

allisonport-db: I will try that^

allisonport-db: Wait, but it failed before even finishing `hiveMR/test`, not sure why separating them would fix this

allisonport-db: Looking back through merged PR's we had this OOM issue the whole time, we just weren't being notified because of how it was set up ([one example](https://github.com/delta-io/connectors/runs/4168496586?check_suite_focus=true)), so this seems like a larger issue. Last commit broke up the hive tests and they're failing similarly.

zsxwing: Yep. You are right. It never passes. Looks like breaking up them like this fixes the issue: https://github.com/zsxwing/connectors/commit/0e0de9cf0a1c5cd26b16f5ce0fea5307cb3822b2

allisonport-db: Sounds good. Only thing, is the failure has been inconsistent (in past PR's and even within the commits on this PR) so not sure how we can confirm that fixes it?

allisonport-db: Same issue.

zsxwing: Hm that's weird. It passed in my repo: https://github.com/zsxwing/connectors/runs/4257627550?check_suite_focus=true

zsxwing: Okay. I retriggered the above job and it failed... Let me try to increase the jvm heap size. We should have enough memory according to https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources 

```
2-core CPU
7 GB of RAM memory
14 GB of SSD disk space
```

zsxwing: Tried multiple times. Looks like downgrading ubuntu fixes it https://github.com/zsxwing/connectors/commit/3b5a3683b03222a63219384cdee2f54e409ce9f9

allisonport-db: Same error

allisonport-db:
@zsxwing Removed the hive tests. Should I add back a new cache?

zsxwing:
> Should I add back a new cache?

No. I can take care of it when I'm adding the hive tests back.

zsxwing (approved):
LGTM

PR #214, opened 2021-11-17
Time travel bug fix

scottsand-db:
This PR tests the following scenario:
```
N.checkpoint.parquet // N is earliest version of the table, N.json does not exist
N+1.json
N+2.json
...
```

and users time travel to version > N.

This PR does not fix the edge case of when users time travel to version N. That was *incorrectly* solved in PR https://github.com/delta-io/connectors/pull/211, which is closed.


standalone/src/main/scala/io/delta/standalone/internal/DeltaHistoryManager.scala
```diff
@@ -128,7 +128,7 @@ private[internal] case class DeltaHistoryManager(deltaLog: DeltaLogImpl) extends
 
         // Note that we also check this condition at the end of the function - we check it
         // here too to to try and avoid more file listing when it's unnecessary.
-        if (lastCompleteCheckpoint.exists(_ >= smallestDeltaVersion)) {
+        if (lastCompleteCheckpoint.exists(_ >= smallestDeltaVersion - 1)) {
```
scottsand-db, line 131: this is the fix

zsxwing: I don't get this fix. If `lastCompleteCheckpoint` is 20 and `smallestDeltaVersion` is 21, we should continue to move to the next checkpoint, since we cannot construct Snapshot 20.

standalone/src/test/scala/io/delta/standalone/internal/DeltaTimeTravelSuite.scala
```diff
@@ -225,4 +225,33 @@ class DeltaTimeTravelSuite extends FunSuite {
...
+        log.getSnapshotForTimestampAsOf(19)
+      }
+
+      // this tests an edge case where we travel to version N where N.parquet exists, N.json does
+      // not exist, and N is the earliest delta version in the table
+      //
+      // in the future, log cleanup logic should be changed so that we always keep N.json during
+      // this edge case
+      intercept[NoSuchElementException] {
+        log.getSnapshotForVersionAsOf(20)
```
zsxwing, line 249: IIUC, your change will make `getEarliestReproducibleCommitVersion` return 20, and the failure will happen when constructing snapshot 20. Why do we need to change this?

scottsand-db: Without my change, time traveling to version 20 will fail **and** time traveling to version 23 will fail.

With my change, time traveling to version 20 still fails **but** time traveling to version 23 succeeds.

So my change improves functionality

standalone/src/main/scala/io/delta/standalone/internal/DeltaHistoryManager.scala
scottsand-db: this is the code that is in DBR and OSS (including the `- 1`) so I think having it is correct.

if lastCompleteCheckpoint is 20, we should be able to construct snapshot 20, right?

the reason we can't is due to that `deltas.last` NoElementException error thrown in SnapshotManagement, which we've discussed in Slack.

zsxwing: I see. It's better to get this method more robust instead. It should not return 20 when 20.checkpoint exists but 20.json doesn't. Otherwise, people will still query version 20 and hit `NoElementException`.

Even if we fix the log clean up code, people may still use old versions to write to a Delta table. It's better to just fix this special case in this method.

scottsand-db: sgtm

scottsand-db:
Closing this PR. Can revisit later.

PR #216, opened 2021-11-23
DeltaLog.tableExists API

scottsand-db:
@scottsand-db Changed implementation to match Delta and check for snapshot version. If we just checked for `_delta_log` existence, (I assume) we would encounter the same error later on for an empty directory.

[Question] Are these tests sufficient or should it be tested for `/dir_a/` or `/dir_a/dir_b/_delta_log/dir_c` or `/dir_a/dir_b/_delta_log/` or something along those lines.
```
├── dir_a
│   └── dir_b
│         └── _delta_log
│                 └── dir_c
```

### DIFF
--- a/standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
@@ -139,7 +139,7 @@ private[internal] class DeltaLogImpl private(
   }
 
   /** Whether a Delta table exists at this directory. */
-  def tableExists: Boolean = snapshot.version >= 0
+  override def tableExists: Boolean = snapshot.version >= 0
 
   ///////////////////////////////////////////////////////////////////////////
   // Internal Methods
--- a/standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
@@ -138,6 +138,9 @@ private[internal] class DeltaLogImpl private(
     new OptimisticTransactionImpl(this, snapshot)
   }
 
+  /** Whether a Delta table exists at this directory. */
+  def tableExists: Boolean = snapshot.version >= 0
+
   ///////////////////////////////////////////////////////////////////////////
   // Internal Methods
   ///////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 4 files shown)


scottsand-db:
> [Question] Are these tests sufficient or should it be tested for /dir_a/ or /dir_a/dir_b/_delta_log/dir_c or /dir_a/dir_b/_delta_log/ or something along those lines.

Does Delta OSS have any tests for similar tableExists APIs?

allisonport-db:
> > [Question] Are these tests sufficient or should it be tested for /dir_a/ or /dir_a/dir_b/_delta_log/dir_c or /dir_a/dir_b/_delta_log/ or something along those lines.
> 
> Does Delta OSS have any tests for similar tableExists APIs?

No tests.

scottsand-db:
I guess my question is what do you expect the answers to be for those cases you provided?
```
/dir_a/_delta_log --> true
/dir_a/dir_b/_delta_log/dir_c --> ? 
/dir_a/dir_b/_delta_log/  --> im assuming false? since dir_a isn't a delta table, dir_b is
```

If you want to manually compare against similar APIs in Delta OSS, that makes sense to me. Nonetheless having tests for the above examples ^ seems like a good idea to me.

allisonport-db:
I mean I expect them all to be false. In `forTable` we do `apply(hadoopConf, new Path(dataPath, "_delta_log"))` only `/dir_a/dir_b/` should be true. I guess it might be worth just adding the tests.

zsxwing:
An instance method is probably better. With the current API, the code would be

```
if (!DeltaLog.tableExists(hadoopConf, path)) {
  throw exception
} 
val deltaLog = DeltaLog.forPath(hadoopConf, path)
```

With an instance method, it can be
```
val deltaLog = DeltaLog.forPath(hadoopConf, path)
if (!deltaLog.tableExists()) {
  throw exception
}
```
The latter is cleaner as you don't need to pass `(hadoopConf, path)` twice. It also avoids creating DeltaLogImpl twice. Creating DeltaLogImpl is a heavy operation for big tables.

Delta has a similar API: https://github.com/delta-io/delta/blob/master/core/src/main/scala/io/delta/tables/DeltaTable.scala#L628-L647

But I think the current unit test in the PR is enough.

scottsand-db:
You can add this under `// Public API changes in 0.2.0 -> 0.3.0` in StandaloneMimaExcludes

`ProblemFilters.exclude[ReversedMissingMethodProblem]("io.delta.standalone.DeltaLog.tableExists")`

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
```diff
@@ -138,6 +138,9 @@ private[internal] class DeltaLogImpl private(
     new OptimisticTransactionImpl(this, snapshot)
   }
 
+  /** Whether a Delta table exists at this directory. */
+  def tableExists: Boolean = snapshot.version >= 0
```
zsxwing, line 142: nit: override

zsxwing (approved):

PR #217, opened 2021-11-23
Fix flaky build and add Hive tests back

zsxwing:
I noticed the following output in each failed build.

```
Files to process must be specified, found 0.
Files to process must be specified, found 0.
Error: Process completed with exit code 255.
```

After digging more, I found the root cause of our flaky build is the same as https://github.com/etsy/sbt-checkstyle-plugin/issues/32

sbt-checkstyle calls Java checkstyle library which will [call System.exit when the style check fails](https://github.com/etsy/sbt-checkstyle-plugin/blob/58b5c5fc4fc2d45b56ea07a9ad858089e34853e7/src/main/scala/com/etsy/sbt/checkstyle/Checkstyle.scala#L47-L52). In order to avoid exiting SBT JVM process, sbt-checkstyle [uses its own NoExitSecurityManager to prevent exits from the JVM](https://github.com/etsy/sbt-checkstyle-plugin/blob/58b5c5fc4fc2d45b56ea07a9ad858089e34853e7/src/main/scala/com/etsy/sbt/checkstyle/Checkstyle.scala#L122-L141). However, the implementation is [not thread thread](https://github.com/etsy/sbt-checkstyle-plugin/issues/32#issuecomment-976206148), and cannot prevent from exiting the JVM when two concurrent threads are running the [noExit](https://github.com/etsy/sbt-checkstyle-plugin/blob/58b5c5fc4fc2d45b56ea07a9ad858089e34853e7/src/main/scala/com/etsy/sbt/checkstyle/Checkstyle.scala#L122-L141) method.

As most of our projects have zero Java files, Java checkstyle library will fail and call System.exit for most of our projects. Due to the above race condition, the SBT JVM may be killed when sbt-checkstyle's `noExit` method cannot prevent from exiting the JVM. `exit code 255` in our output matches the following code in Checkstyle:
- https://github.com/checkstyle/checkstyle/blob/a42a3e1733f3c7583177a3c96de65e6a2fd7671f/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L87
- https://github.com/checkstyle/checkstyle/blob/a42a3e1733f3c7583177a3c96de65e6a2fd7671f/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L184-L189
- https://github.com/checkstyle/checkstyle/blob/a42a3e1733f3c7583177a3c96de65e6a2fd7671f/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L810-L812

The fix is just disabling parallel execution. While it makes the build slower, the speed is still acceptable since most of time is spent on tests which cannot run in parallel anyway.

This PR also adds build cache, hive tests back and upgrade Ubuntu back to 20.04. The SBT heap size is increased as well as I hit SBT OOM occasionally during my investigation.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -23,7 +23,7 @@ jobs:
             ~/.cache/coursier
             ~/.m2
           key: build-cache-with-scala_${{ matrix.scala }}
-      - name: Run Scala Style tests
+      - name: Run Scala Style tests on test sources
         run: build/sbt "++ ${{ matrix.scala }}" testScalastyle
       - name: Run sqlDeltaImport tests (Scala 2.12 only)
         run: build/sbt "++ ${{ matrix.scala }}" sqlDeltaImport/test
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -3,7 +3,7 @@ on: [push, pull_request]
 jobs:
   build:
     name: "Run tests "
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
         scala: [2.12.8, 2.11.12]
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yaml
```diff
@@ -14,15 +14,28 @@ jobs:
...
+      - name: Cache Scala, SBT
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.sbt
+            ~/.ivy2
+            ~/.cache/coursier
+            ~/.m2
+          key: build-cache-with-scala_${{ matrix.scala }}
+      - name: Run Scala Style tests
```
scottsand-db, line 26: `testScalastyle` only runs scalastyle on test sources.
So perhaps rename this to "Run Scala Style tests on test sources"

scottsand-db (approved):
LGTM with 1 minor comment

allisonport-db (approved):
LGTM

PR #219, opened 2021-11-24
Add null-check for data reader partition field

scottsand-db:
Our `RowRecord` APIs say that we will throw a `NullPointerException if field is not nullable and {@code null} data value read`. We are doing this for data fields, but we forgot to do this for partition fields.

We can't test this as we can't insert a row with a null value for a a non-nullable collumn.


zsxwing (approved):
LGTM

PR #220, opened 2021-11-30
Flink Delta Sink - PR 3

pkubit-g:
PR adding transactional ids (appId + checkpointId) to the Flink DeltaSink.

### DIFF
--- a/flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSink.java
+++ b/flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSink.java
@@ -36,6 +36,8 @@ import org.apache.flink.core.fs.Path;
 import org.apache.flink.core.io.SimpleVersionedSerializer;
 import org.apache.flink.formats.parquet.ParquetWriterFactory;
 import org.apache.flink.formats.parquet.row.ParquetRowDataBuilder;
+import org.apache.flink.streaming.api.functions.sink.filesystem.bucketassigners.BasePathBucketAssigner;
+import org.apache.flink.streaming.api.functions.sink.filesystem.rollingpolicies.OnCheckpointRollingPolicy;
 import org.apache.flink.table.data.RowData;
 import org.apache.flink.table.types.logical.RowType;
 import org.apache.flink.util.FlinkRuntimeException;
@@ -94,6 +96,14 @@ public class DeltaSink<IN>
     /**
      * This method creates the {@link SinkWriter} instance that will be responsible for passing
      * incoming stream events to the correct bucket writer and then flushed to the underlying files.
+     * <p>
+     * The logic for resolving constructor params differ depending on whether any previous writer's
+     * states were provided.
+     * If there are no previous states then we assume that this is a fresh start of the app and set
+     * next checkpoint id in {@link DeltaWriter} to 1 and app id is taken from the
+     * {@link DeltaSinkBuilder#getAppId} what guarantees us that each writer will get the same
+     * value. In other case, if we are provided by the Flink framework with some previous writers'
+     * states then we use those to restore values of appId and nextCheckpointId.
      *
      * @param context {@link SinkWriter} init context object
      * @param states  restored states of the writers. Will be empty collection for fresh start.
@@ -105,9 +115,52 @@ public class DeltaSink<IN>
         InitContext context,
         List<DeltaWriterBucketState> states
     ) throws IOException {
-        DeltaWriter<IN> writer = sinkBuilder.createWriter(context);
+        String appId = restoreOrCreateAppId(states);
+        long nextCheckpointId = restoreOrGetNextCheckpointId(states);
+        DeltaWriter<IN> writer = sinkBuilder.createWriter(context, appId, nextCheckpointId);
         writer.initializeState(states);
         return writer;
+    }
+
+    /**
+     * In order to gurantee the idempotency of the GlobalCommitter we need unique identifier of the
+     * app. We obtain it with simple logic: if it's the first run of the application (so no restart
+     * from snapshot or failure recovery happened and the writer's state is empty) then assign appId
+     * to a newly generated UUID that will be further stored in the state of each writer.
+     * Alternatively if the writer's states are not empty then we resolve appId from one of the
+     * restored states.
+     *
+     * @param states restored list of writer's buckets states that include previously generated
+     *               appId
+     * @return newly created or resolved from restored writer's states unique identifier of the app.
+     */
+    private String restoreOrCreateAppId(List<DeltaWriterBucketState> states) {
+        if (states.isEmpty()) {
+            return sinkBuilder.getAppId();
+        }
+        return states.get(0).getAppId();
+    }
+
+    /**
+     * In order to gurantee the idempotency of the GlobalCommitter we need to version consecutive
+     * commits with consecutive identifiers. For this purpose we are using checkpointId that is
+     * being "manually" managed in writer's internal logic, added to the committables information
+     * and incremented on every precommit action (after generating the committables).
+     *
+     * @param states restored list of writer's buckets states that include previously generated
+     *               appId
+     * @return value of the nextCheckpointId to be passed to the new writer's instance
+     */
+    private long restoreOrGetNextCheckpointId(List<DeltaWriterBucketState> states) {
+        if (states.isEmpty()) {
+            return 1;
+        }
+        return states
+            .stream()
+            .map(DeltaWriterBucketState::getCheckpointId)
+            .mapToLong(v -> v)
+            .max()
+            .getAsLong();
     }
 
     @Override
@@ -150,10 +203,7 @@ public class DeltaSink<IN>
     }
 
     /**
-     * Convenience method for creating {@link DeltaSink}
-     * <p>
-     * For configuring additional options (e.g. bucket assigners in case of partitioning tables)
-     * see {@link DeltaSinkBuilder}.
+     * Convenience method for creating {@link DeltaSink} to a non-partitioned DeltaLake's table.
      *
      * @param basePath root path of the DeltaLake's table
      * @param conf     Hadoop's conf object that will be used for creating instances of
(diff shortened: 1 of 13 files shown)


flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSink.java
```diff
@@ -105,11 +115,54 @@
...
+     */
+    private long restoreOrGetNextCheckpointId(List<DeltaWriterBucketState> states) {
+        if (states.isEmpty()) {
+            return 1;
+        }
+        return states
+            .stream()
+            .map(DeltaWriterBucketState::getCheckpointId)
+            .mapToLong(v -> v)
+            .max()
```
tdas, line 162: why are we taking max? shouldnt we be taking min? if different buckets are in different states such that its spread between multiple checkpoint Ids.... if we are recovering ... flink would restart the earliest incomplete checkpoint interval .... isnt it?

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSinkBuilder.java
```diff
@@ -80,7 +90,7 @@
     // FileSink-specific fields
     ///////////////////////////////////////////////////////////////////////////
 
-    private long bucketCheckInterval;
+    private final long bucketCheckInterval;
```
tdas, line 93: i have already forgotten what this is... can you add docs for this?

tdas: also why final?

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSinkBuilder.java
```diff
@@ -230,8 +236,10 @@ SerializableConfiguration getSerializableConfiguration() {
         DefaultDeltaFormatBuilder(
```
tdas, line 236: why is the name DefaultDeltaFormatBuilder? is there going to be a non-default one?

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSink.java
```diff
@@ -105,11 +115,54 @@
...
+     */
+    private String restoreOrCreateAppId(List<DeltaWriterBucketState> states) {
+        if (states.isEmpty()) {
+            return sinkBuilder.getAppId();
+        }
+        return states.get(0).getAppId();
+    }
+
+    /**
+     * In order to gurantee the idempotency of the GlobalCommitter we need to version consecutive
```
tdas, line 145: this is the reason for the existence of checkpoint id, but does not explain what this function does.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSink.java
```diff
@@ -162,7 +212,7 @@
      * @param rowType  Flink's logical type to indicate the structure of the events in the stream
      * @return builder for the DeltaSink
      */
-    public static DeltaSinkBuilder<RowData> forDeltaFormat(
+    public static DeltaSink<RowData> forDeltaFormat(
```
tdas, line 215: I am not sure I want to change this. I actually like the fact that we are exposing the builder class... which allows us to extend it in the future to allow specification of new info through new builder methods. 

but I also agree that there should be only one static method to create the builder .. and then all info are specified via builder methods. parttitioned and unpartitioned table should go through the same static method and builder.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSinkBuilder.java
```diff
@@ -164,35 +177,28 @@ SerializableConfiguration getSerializableConfiguration() {
...
             rollingPolicy,
             outputFileConfig,
             context.getProcessingTimeService(),
-            bucketCheckInterval);
+            bucketCheckInterval,
+            appId,
+            nextCheckpointId);
     }
 
     SimpleVersionedSerializer<DeltaWriterBucketState> getWriterStateSerializer()
```
tdas, line 204: why is the sinkbuilder (a class solely for user APIs) creating these objects? Why make DeltaSink delegate to these methods... instead of creating the objects directly.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/writer/DeltaWriter.java
```diff
@@ -75,6 +76,34 @@
...
+     */
+    private final String appId;
+
+    /**
+     * Unique identifier of a checkpoint interval. It's necessary to maintain and increment this
+     * value inside writer as this needs to be passed as a part of committables' information during
+     * {@link DeltaWriter#prepareCommit} method. Its value is always incremented by one after
+     * generating set of committables for given checkpoint interval. It's also snapshotted as a part
+     * of the writer's state in order to support failure recovery and provide exactly-once delivery
+     * guarantee.
```
tdas, line 103: does this start with 1 or 0?

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/writer/DeltaWriter.java
```diff
@@ -75,6 +76,34 @@
...
+     * the writer's states with no active buckets.
+     */
+    public static final String NOOP_WRITER_STATE = "<noop-writer-state>";
+
+    ///////////////////////////////////////////////////////////////////////////
+    // DeltaSink-specific fields
+    ///////////////////////////////////////////////////////////////////////////
+
+    /**
+     * Unique identifier of the application that will be passed as part of committables' information
```
tdas, line 90: what is the lifecycle of this appId? that is, mention that this appid will be unique to a streaming job as long as it is being restarted using checkpoint information

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSink.java
pkubit-g: max is correct.
Please note that we are generating committables in DeltaWriter#prepareCommit method and its before emitting the `snapshotPreBarrier`. So once all of the writers will emit their committables for given checkpoint interval and reach the barrier only then framework will proceed with the snapshot and the commit.
And the states that we are handling here are a result of DeltaWriter#snapshotState that is always called after DeltaWriter#prepareCommit  and after emitting the pre-snapshot barrier so checkpoint id will be always aligned among the writers for given checkpoint interval. 
And if we are given states from different checkpoint intervals (with different checkpoint ids) then we should not care for the older ones as it means that Flink has already snapshotted their committables and will try to replay them during the next commit phase - that's why we should be using max. We want to resolve here the nextCheckpointId value that will be passed to the newly generated set of committables.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSinkBuilder.java
pkubit-g: added, will commit in a moment

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSinkBuilder.java
pkubit-g: I've just followed the semantics from Flink's FileSink that have similar construct. 
I'm guessing that the reason for this is to expose all the most important properties for the Sink which will make constructing it more concise. The non-default one would be simply calling all the `with<property>` methods.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSink.java
pkubit-g: added:
```
     * Restores the last checkpoint id snapshotted in one of the most recent {@link DeltaWriter}s'
     * states or sets it to "1" in case there is no previous states.
```
i've also added similar sentence

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSink.java
pkubit-g: ok, so i must've understood you incorrectly during the status when i got the impression that you don't want to expose the builder at all.
I'm changing it back here and let's align again during a status on this user exposed static methods and builder.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSinkBuilder.java
pkubit-g: It was the same in Flink's FileSink and i've decided to stick to that keeping in mind that while it may not possible currently to directly reuse all of its classes, but if it provides few minor changes in future (in terms of generics and access modifiers) then it will be possible to drop some code and obsolete from `DeltaSink`. In order to keep such a refactor to a minimum i thought it will be a good idea to stick closesly

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/writer/DeltaWriter.java
pkubit-g: With 1. same as for Flink's checkpoint id. I've added this info to the description.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/writer/DeltaWriter.java
pkubit-g: added

pkubit-g:
@tdas FYI i've just committed corrected code review comments

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSink.java
tdas: thanks!

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSink.java
tdas: this is a very key explanation here. please add this as inline comments. without this, its impossible to grasp why `max` is correct.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSink.java
tdas: yes. and i think its my fault. I didnt grasp that the builder is the public API. my apologies for causing this flip flopping. yes lets discuss offline what is the right builder API

tdas (approved):

PR #224, opened 2021-12-02
Flink Delta Sink - PR 4

pkubit-g:


build.sbt
```diff
@@ -641,7 +641,18 @@ lazy val flinkConnector = (project in file("flink-connector"))
       "org.apache.flink" %% "flink-streaming-java" % flinkVersion % "test",
       "org.apache.flink" % "flink-connector-test-utils" % flinkVersion % "test",
       "com.github.sbt" % "junit-interface" % "0.12" % Test
-    )
+    ),
+    sourceGenerators in Compile += Def.task {
```
tdas, line 645: add comments on why this is needed

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committables/DeltaGlobalCommittableSerializer.java
```diff
@@ -68,18 +77,30 @@ public DeltaGlobalCommittable deserialize(int version, byte[] serialized) throws
     }
 
     private void serializeV1(DeltaGlobalCommittable committable, DataOutputView dataOutputView)
-            throws IOException {
+        throws IOException {
+        dataOutputView.writeInt(committable.getDeltaCommittables().size());
+        for (DeltaCommittable deltaCommittable : committable.getDeltaCommittables()) {
+            new DeltaCommittableSerializer(pendingFileSerializer)
```
tdas, line 83: nit: do you need to create the serializer once for every committable? cant the serializer instance be reused?

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -49,27 +68,301 @@
...
+     * During this process we not only map the single committables to {@link AddFile} actions and
+     * group them by checkpointId but also flatten collection of {@link DeltaGlobalCommittable}
+     * objects (each containing its own collection of {@link DeltaCommittable}). Additionally,
+     * during the iteration process we also validate whether the committables for the same
+     * checkpoint interval have the same set of partition columns and throw
+     * a {@link RuntimeException} when this condition is not met.
+     *
+     * @param globalCommittables list of combined @link DeltaGlobalCommittable} objects
+     * @return {@link AddFile} actions grouped by the checkpoint interval in the form of
+     * {@link java.util.Map}. It is guaranteed that all actions per have the same partition columns
```
tdas, line 278: actions per checkpoint

tdas: wait... this line does not make sense. all actions in this entire pipeline should have the same set of partition columns.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committables/DeltaGlobalCommittableSerializer.java
```diff
@@ -68,18 +77,30 @@ public DeltaGlobalCommittable deserialize(int version, byte[] serialized) throws
...
+        }
     }
 
     private DeltaGlobalCommittable deserializeV1(DataInputView dataInputView) throws IOException {
-        return new DeltaGlobalCommittable(Collections.emptyList());
+        int deltaCommittablesSize = dataInputView.readInt();
+        List<DeltaCommittable> deltaCommittables = new ArrayList<>(deltaCommittablesSize);
+        for (int i = 0; i < deltaCommittablesSize; i++) {
+            DeltaCommittable deserializedCommittable =
+                new DeltaCommittableSerializer(pendingFileSerializer).deserializeV1(dataInputView);
```
tdas, line 93: same qs as above

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -49,27 +68,301 @@
...
+        this.conf = conf;
+        this.basePath = basePath;
     }
 
+    /**
+     * Filters committables that will be provided to {@link GlobalCommitter#commit} method.
+     * <p>
+     * We are always returning all the committables as we do not implement any retry behaviour
+     * in {@link GlobalCommitter#commit} method and always want to try to commit all the received
+     * committables.
```
tdas, line 103: can you also make a note that if there are committables accros multiple checkpoint intervals, we will handle them correctly when trying to commit... not by filtering them here

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -49,27 +68,301 @@
...
+        for (DeltaGlobalCommittable globalCommittable : globalCommittables) {
+            for (DeltaCommittable deltaCommittable : globalCommittable.getDeltaCommittables()) {
+                return deltaCommittable.getAppId();
+            }
+        }
+        return null;
+    }
+
+    /**
+     * Commits already written files to the DeltaLake table using unique identifier for the given
```
tdas, line 149:
```suggestion
     * Commits already written files to the Delta table using unique identifier for the given
```

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -49,27 +68,301 @@
...
+     * Flink job (appId) and checkpointId delivered with every committable object. Those ids
+     * together construct transactionId that will be used for verification whether given set of
+     * files has already been committed to the Delta table.
+     *
+     * <p>During commit preparation phase:
+     *
+     * <ol>
+     *   <li>first appId is resolved from any of the provided committables.
```
tdas, line 157:
```suggestion
     *   <li>First, appId is resolved from any of the provided committables.
```

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -49,27 +68,301 @@
...
+     *       If no appId is resolved then it means that no committables were provided and no commit
+     *       is performed. Such situations may happen when e.g. there were no stream events received
+     *       within given checkpoint interval,
+     *   <li>If appId is successfully resolved then the provided set of committables needs to be
+     *       flatten (as one {@link DeltaGlobalCommittable} contains a list of
```
tdas, line 162:
```suggestion
     *       flattened (as each {@link DeltaGlobalCommittable} contains a list of
```

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -49,27 +68,301 @@
...
+     *       {@link DeltaCommittable}), mapped to {@link AddFile} objects and then grouped by
+     *       checkpointId. The grouping part is necessary as committer object may receive
+     *       committables from different checkpoint intervals,
+     *   <li>we process each of the resolved checkpointId incrementally,
```
tdas, line 166: incrementally -> in increasing order

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -49,27 +68,301 @@
...
+     *   <li>during processing each of the checkpointId and their committables we first query
```
tdas, line 167:
```suggestion
     *   <li>during processing each of the checkpointId and their committables, we first query
```

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -49,27 +68,301 @@
...
+     *       the DeltaLog for last checkpointed version for given appId. Here checkpointed version
```
tdas, line 168:
```suggestion
     *       the DeltaLog for last committed transaction version for given appId. Here transaction version
```

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -49,27 +68,301 @@
...
+     *       equals checkpointId. We proceed with the transaction only if current checkpointId is
+     *       greater than last checkpointed version.
```
tdas, line 170:
```suggestion
     *       greater than last committed transaction version.
```

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -49,27 +68,301 @@
...
+     *   <li>if above condition is met the we handle the metadata for data in given stream by
```
tdas, line 171:
```suggestion
     *   <li>if above condition is met, then we handle the metadata for data in given stream by
```

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -49,27 +68,301 @@
...
+     *       comparing the stream's schema with current table snapshot's schema. We proceed with
+     *       the transaction only when the schemas are matching or when it was explicitly configured
+     *       during creation of the sink that we can try to update the schema.
+     *   <li>if above validation passes then we prepare the final set of {@link Action} objects to
+     *       be committed along with transaction's metadata and mandatory parameters,
+     *   <li>we try to commit the prepared transaction
+     *   <li>if the commit fails then we fail the application as well. If it succeeds then we
```
tdas, line 178: it would be more appropriate to capitalize the first letter in each bullet point since they are complete sentences

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -49,27 +68,301 @@
...
     @Override
     public List<DeltaGlobalCommittable> commit(List<DeltaGlobalCommittable> globalCommittables) {
+        String appId = resolveAppId(globalCommittables);
+        if (appId != null) { // means there are committables to process
+            Map<Long, List<AddFile>> addFileActionsPerCheckpoint =
+                prepareAddFileActionsPerCheckpoint(globalCommittables);
+            DeltaLog deltaLog = DeltaLog.forTable(conf, basePath.getPath());
+
+            List<Long> sortedCheckpointIds =
+                getSortedListOfCheckpointIds(addFileActionsPerCheckpoint);
```
tdas, line 194: cant you squash `prepareAddFileActionsPerCheckpoint` and `getSortedListOfCheckpointIds` together? that is `prepareAddFileActionsPerCheckpoint` returns a sorted sequence of `checkpointId -> actions`?

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -49,27 +68,301 @@
...
+                    );
+
+                    transaction.commit(actions, operation, ENGINE_INFO);
+                }
+            }
+        }
         return Collections.emptyList();
     }
 
+    private List<Long> getSortedListOfCheckpointIds(
```
tdas, line 216: as said above this can be squashed

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -49,27 +68,301 @@
...
+     *
+     * @param globalCommittables list of combined committables objects
+     * @return always empty collection as we do not want any retry behaviour
+     */
     @Override
     public List<DeltaGlobalCommittable> commit(List<DeltaGlobalCommittable> globalCommittables) {
+        String appId = resolveAppId(globalCommittables);
+        if (appId != null) { // means there are committables to process
+            Map<Long, List<AddFile>> addFileActionsPerCheckpoint =
+                prepareAddFileActionsPerCheckpoint(globalCommittables);
```
tdas, line 190: minor optimization. you could delay the conversion to AddFile actions to later, when you have passed all the validations for a checkpoint id. until then you can keep it as Seq of [checkpointId -> list of committables]

then this would help in squashing the addfile generation with `prepareActionsForTransaction`

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -49,27 +68,301 @@
...
+            for (long checkpointId : sortedCheckpointIds) {
+                OptimisticTransaction transaction = deltaLog.startTransaction();
+                long lastCommittedVersion = transaction.txnVersion(appId);
+                if (checkpointId > lastCommittedVersion) {
+                    List<AddFile> addFilesForCurrentCheckpointId =
+                        addFileActionsPerCheckpoint.get(checkpointId);
+                    List<Action> actions = prepareActionsForTransaction(
+                        appId, checkpointId, addFilesForCurrentCheckpointId);
+                    Operation operation = prepareDeltaLogOperation(
+                        addFileActionsPerCheckpoint.get(checkpointId).size(),
```
tdas, line 205:
```suggestion
                        addFilesForCurrentCheckpointId.size(),
```
??

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -49,27 +68,301 @@
...
+        return operationMetrics;
+    }
+
+    /**
+     * Converts {@link DeltaPendingFile} object to a {@link AddFile} object
+     *
+     * @param deltaPendingFile input object to be converted to {@link AddFile} object
+     * @return {@link AddFile} object generated from input
+     */
+    private AddFile convertDeltaPendingFileToAddFileAction(DeltaPendingFile deltaPendingFile) {
```
tdas, line 354: this can be a method in the `DeltaPendingFile` class. less clutter in this class.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -49,27 +68,301 @@
...
+                addFilesPerCheckpoint.get(checkpointId).add(action);
+
+                // prepare operation's metrics
+                if (!operationMetricsPerCheckpoint.containsKey(checkpointId)) {
+                    operationMetricsPerCheckpoint.put(checkpointId, new HashMap<>());
+                }
+                Map<String, Long> currentOperationMetricsPerCheckpoint =
+                    operationMetricsPerCheckpoint.get(checkpointId);
+
+                long currentNumOutputRows = currentOperationMetricsPerCheckpoint
```
tdas, line 304: as i suggested above, if you write the function to convert all committables for one checkpoint id at a time, then this calculation will get much simpler. you do not have to maintain this map and all.

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/DeltaSinkStreamingExecutionITCase.java
```diff
@@ -230,7 +270,9 @@ public void run(SourceContext<RowData> ctx) throws Exception {
                 // run until finished.
                 sendRecordsUntil(numberOfRecords, ctx);
 
-                // Wait the last checkpoint to commit all the pending records.
+                // need to add tiny time buffer as in very rare cases the job is terminated by test
+                // runner before the global commit is finalized
+                Thread.sleep(50);
```
tdas, line 275: this is recipe for test flakiness. rather use a loop to check repeatedly until a large timeout

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/DeltaSinkStreamingExecutionITCase.java
```diff
@@ -112,9 +127,34 @@ public void runDeltaSinkTest() throws Exception {
         }
 
         // THEN
```
tdas, line 129: when will the above code unblock?

tdas: when it unblocks... is there are guarantee that files will already be written in the path as tested in the next line? or can there be a race condition

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/DeltaSinkStreamingExecutionITCase.java
```diff
@@ -112,9 +127,34 @@ public void runDeltaSinkTest() throws Exception {
...
+        DeltaSinkTestUtils.validateIfPathContainsParquetFilesWithData(deltaTablePath);
         int writtenRecordsCount =
             DeltaSinkTestUtils.validateIfPathContainsParquetFilesWithData(deltaTablePath);
```
tdas, line 132: why checking twice??

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/DeltaSinkBatchExecutionITCase.java
```diff
@@ -69,9 +84,31 @@ public void runDeltaSinkTest() throws Exception {
         }
 
         // THEN
+        DeltaSinkTestUtils.validateIfPathContainsParquetFilesWithData(deltaTablePath);
         int writtenRecordsCount =
             DeltaSinkTestUtils.validateIfPathContainsParquetFilesWithData(deltaTablePath);
```
tdas, line 89: same qs as above. why duplicate check

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTest.java
```diff
@@ -0,0 +1,261 @@
...
+    @Before
+    public void setup() throws IOException {
+        tablePath = new Path(TEMPORARY_FOLDER.newFolder().toURI());
+    }
+
+    @Test
+    public void testCommitTwice() throws Exception {
+        //GIVEN
+        int numAddedFiles = 3;
+        DeltaSinkTestUtils.initializeTestStateForNonPartitionedDeltaTable(tablePath.getPath());
```
tdas, line 69: man.. these function names are huuuugee. its a chore to read them. can you refactor them to be shorter and sweeter? like `initTestForNonPartitionedTable`

tdas: if this is being called in every test... why not put it in `setup()`?

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
pkubit-g: Yes they should and that's why my understanding it that during global commit phase we should validate for this as it's technically possible that some records may have been written to paths that do not correspond to any of Delta table's partition.
It's because partition assignment is resolved on an event-level using `org.apache.flink.streaming.api.functions.sink.filesystem.BucketAssigner` and users need to provide DeltaSink with its implementation. If the implementation would be wrong then the files are written to wrong paths and that's why during global commit i'm checking every committable for partitions' consistency.

build.sbt
pkubit-g: done

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committables/DeltaGlobalCommittableSerializer.java
pkubit-g: yeah, you're right, thanks for catching this!

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committables/DeltaGlobalCommittableSerializer.java
pkubit-g: done

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
pkubit-g: yeah... 
good catch!

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/DeltaSinkStreamingExecutionITCase.java
pkubit-g: my bad. I simply got lost in merging & rebasing as in earlier versions i didn't need to have `writtenRecordsCount` value so i was just simply calling validation method. Thank you for catching it, will remove it.

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/DeltaSinkStreamingExecutionITCase.java
pkubit-g: i think your questions relates to [this one](https://github.com/delta-io/connectors/pull/224#discussion_r761698662) so let me squash my answers here.

It unblocks when `DeltaStreamingExecutionTestSource` has sent number of events equal to `FileSinkITBase.NUM_RECORDS` * num_sources and the very next checkpoint after it happens is set to be the last one. There is this whole setup with `DeltaSinkStreamingExecutionITCase.LATCH_MAP` that accounts for this behaviour that i've borrowed from `StreamingExecutionFileSinkITCase.StreamingExecutionTestSource`, where sources synchronize on this latch and release only when all of them have sent their part. And then Flink recognizes that all of the sources for the job have terminated so it terminates the job. 

And in general it works great, except some race condition when the job terminates before the files are committed to the DeltaLog. It was extremely hard to debug and i didn't managed to reproduce as this is happening literally 1/125 cases (every time i run a test loop it fails around 250 run where half of those are failure scenarios).

After adding this `Thread.sleep(50);` i was able to run it 400-450 times (>800 test runs in total, but only half were failure scenarios) without issue. 
Since i cannot think of any condition to put here, i will simply add a workaround for the job to run one additional checkpoint interval after the termination conditions are met - let me know if this sounds good to you. I've already tested this workaround with running it 550 times and all of them passed - will commit it with the rest of the code review corrections.

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/DeltaSinkBatchExecutionITCase.java
pkubit-g: replied above

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
pkubit-g: ok, i have decoupled grouping the committables by checkpoint id from converting them to AddFiles (added methods: `groupCommittablesByCheckpointInterval` and `prepareAddFileActions` and removed `prepareAddFileActionsPerCheckpoint`).

However i haven't squashed generating the add files with preparing all transaction's actions since in the incoming PRs i need to reuse some information about the addFiles. Please look at the reference [implementation](https://github.com/pkubit-g/connectors/blob/flink_sink_feature/flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java#L223) (with schema management and partition support) for details.

I think thath squasing all of those functions would make the code a bit messy.

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/DeltaSinkStreamingExecutionITCase.java
pkubit-g: please look [here](https://github.com/delta-io/connectors/pull/224#discussion_r762642265) for my response

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTest.java
pkubit-g: i've changed the name of the utliity method. 

As for putting it in `setup()` - please mark that we have other PRs ingoing and we will be testing other configuration as well, so not every test will have the same setup.

pkubit-g:
@tdas thank you for your thorough review. I've just pushed corrections for most of the comments and also marked the pretty obvious discussions as resolved. Other i've answered and left open so you can revisit.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
tdas: that is fine then.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
tdas: resolving of partition column can happen directly on the committables... isnt it? so both the partition columns and schema resolution can be done on committables. And only after all validation + update are done, you can prepare the final set of actions by converting committables to AddFiles AND collecting stats to prepare the operation action.... all in one function. without using these hashmap as fields in OptTxn... isnt it?

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/DeltaSinkStreamingExecutionITCase.java
tdas: this still sounds flaky to me. there is some underlying race condition that we are not fixing with some deterministic condition... so the best thing we can do here to avoid ANY FLAKINESS is to iteratively check until time out. in scala... the pattern is very simple using this helper class `Eventually` - http://doc.scalatest.org/1.8/org/scalatest/concurrent/Eventually.html

See if this pattern can be replicated in Java. quick google search led me to this - https://github.com/awaitility/awaitility

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTest.java
tdas: okay. I think it will be good to pass after all the tests are in to dedup code across them.

tdas:
@pkubit-g answered those questions. thank you for your updates.

flink-connector/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/DeltaPendingFile.java
```diff
@@ -94,6 +97,23 @@ public long getLastUpdateTime() {
         return lastUpdateTime;
     }
 
+    /**
+     * Converts {@link DeltaPendingFile} object to a {@link AddFile} object
+     *
+     * @return {@link AddFile} object generated from input
+     */
+    public AddFile convertDeltaPendingFileToAddFileAction() {
```
tdas, line 105: too verbose!?? why say "convertX" when it is already a method of X? 
Just `generateAddFile` or `toAddFile()` is sufficient.

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/DeltaSinkStreamingExecutionITCase.java
pkubit-g: `iteratively check until time out` i agree that this would also help resolving the issue, however i cannot think of any condition that should be checked during those iterations. 
Also, i don't think that the current setup is flaky - we simply run the source function (and so the job) for couple checkpoint intervals, then we hit the condition that all of the input data have been generated and simply let the source/job run for one more checkpoint interval. Cannot see any flakiness inhere and hundreds of test runs that i've made seem to proof this.

I agree though that the test setup is not perfect since the `DeltaStreamingExecutionTestSource` relies on `notifyCheckpointComplete` method which is kind of internal one and is not advised to be used and relied on in production environments. But this is how Flink devs implemented their test cases for FileSink and i don't think i can do any better. It's showing no issues for "local" commits, only in very rare cases causes premature finalization of the test job before the end of the GlobalCommit, but it's only connected to the implementation of the test source, not the DeltaSink itself and simply allowing the job to run for one additional checkpoint interval resolves the issue. 

Screenshot from 2021-12-07 12-45-43

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
pkubit-g: @tdas can you look at [this](https://github.com/pkubit-g/connectors/commit/3d69e3f662a15e8e470600e73865504fcb8eedfc) commit and tell me if this is what you mean ? 
I've squashed those operations into one `commit` method. Please mark that it's for the final version of the global commit, if you confirm then i will migrate this change to the PR4 (and also correct the java docs)

tdas: In that commit... i still dont like [this code range](https://github.com/pkubit-g/connectors/commit/3d69e3f662a15e8e470600e73865504fcb8eedfc#diff-f4312421878fffa98010cec250092b721cdeca23eb57a8d0f3d1f86037aa818bR250-R261) .... why do we need the map at all. since the calculation is all inside this one function... isnt it just 2 counters for numBytes and numOutputRows? wouldnt that make the code much simpler?

tdas: If you are okay with my suggestion to make them counters... then go ahead with all the refactoring in this and subsequent PRs. I am okay with this.

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/DeltaSinkStreamingExecutionITCase.java
tdas: > simply let the source/job run for one more checkpoint interval.
If that is what Thread.sleep(50) is for, then that is where the flakiness could happen. Why are you sleeping for this specific number = 50 ms? why not 1 second or 10 second. What happens if there is a Java GC pause (can happen in overloaded test infra) and instead of 50ms the next checkpoint interval took 1 second. Then this test can fail right? We have seen all kinds of flakiness and its better in the long term to invest in guarding against them rather than fixing them when they start randomly failing builds.

pkubit-g: @tdas ok, we may have gotten slightly lost in the discussion :) [here](https://github.com/delta-io/connectors/pull/224#discussion_r762642265) i was trying to communicate that the commit correcting code review remarks will rework this issue as well and remove `Thread.sleep(50)`.
[This](https://github.com/delta-io/connectors/pull/224/commits/94e83f48a0b425280f501e85bc2bc9851ec101b7#diff-ef63a9cf3a3ec76e4c9275581869e416a57b6a839e3f27206c2566674aaa4008R228) is the change that i'm talking about and it's already part of this PR. Specifically i've added `isLastCheckpointInterval` flag that allows me to run one additional checkpoint interval after all the input data have been generated which seems to eliminate the flakiness (i've run the failover scenario >550 times without any issue).

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
pkubit-g: yes of course. Now after squashing all those functions it totally makes sense to make those as counters and generate the map only afterwards. Sorry for missing that, i've just focuses too much on making sure that i got right the main idea of your proposition. Refactored.
FYI [here](https://github.com/pkubit-g/connectors/blob/flink_sink_feature/flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java)'s how it will look like for the complete implementation and i'm commiting the same for PR4 in a minute.

flink-connector/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/DeltaPendingFile.java
pkubit-g: done

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
tdas: 👍

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/DeltaSinkStreamingExecutionITCase.java
tdas: sounds good for now.

tdas (approved):

PR #227, opened 2021-12-03
Automate adding new PR's and moving updated PR's on project board

allisonport-db:
### TODO
1. Set up credentials (requires admin access)
    - Generate an [access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) with `repo` access
    - [Store the token](https://docs.github.com/en/actions/security-guides/encrypted-secrets#creating-encrypted-secrets-for-an-organization) as a secret at the organization level
    - Match the secret name in the `yaml` files with the stored token
2. Exclude internal github usernames from this workflow (how should I determine this "list"?)
   ```yaml
   if: !contains(["username1", "username2", ..., "usernameN"], github.event.actor.login)
   ```

### DIFF
--- a/.github/workflows/updated_pull_request.yaml
+++ b/.github/workflows/updated_pull_request.yaml
@@ -14,6 +14,6 @@ jobs:
       - uses: alex-page/github-project-automation-plus@2af3cf061aeca8ac6ab40a960eee1968a7f9ce0e
         if: ${{ github.event.pull_request || github.event.issue.pull_request }}
         with:
-          project: oss-eng-delta-ecosystem
+          project: oss-delta-prs
           column: Needs Review
           repo-token: ${{ secrets.PROJECT_BOARD_AUTOMATION_TOKEN }}
--- a/.github/workflows/updated_pull_request.yaml
+++ b/.github/workflows/updated_pull_request.yaml
@@ -8,11 +8,10 @@ on:
 
 jobs:
   automate-updated-pull-requests:
-    if: ${{ !contains(["allisonport-db", "dennyglee", "scottsand-db", "tdas", "zsxwing"], github.event.actor.login) }}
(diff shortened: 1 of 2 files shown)


allisonport-db:
So the `github.event.actor` is whoever triggered the event, so in a new or updated PR that should be the owner of the PR, and in a comment it should be the commenter, but we should double check once it's merged.

allisonport-db:
test comment

.github/workflows/updated_pull_request.yaml
```diff
@@ -0,0 +1,19 @@
...
+    types: [created]
+  pull_request_target:
+    types: [synchronize]
+
+jobs:
+  automate-updated-pull-requests:
+    if: ${{ !contains(["allisonport-db", "dennyglee", "scottsand-db", "tdas", "zsxwing"], github.event.actor.login) }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: alex-page/github-project-automation-plus@2af3cf061aeca8ac6ab40a960eee1968a7f9ce0e
```
zsxwing, line 14: why this is not using a version?

zsxwing (approved):
Since this has to be merged into the master, feel free to merge it. This doesn't affect any code.

.github/workflows/updated_pull_request.yaml
allisonport-db: There was a bug in `github-project-automation-plus` so it wouldn't work for comments on PR's (only on issues) that I fixed in a PR. It hasn't been merged / rereleased yet so this points to the commit.

zsxwing: I see. Let's add a comment and a TODO item to use a released version.

PR #230, opened 2021-12-08
Flink Delta Sink - PR 5

pkubit-g:
This PR adds support for schema validation in flink-connector

### DIFF
--- a/flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSinkBuilder.java
+++ b/flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSinkBuilder.java
@@ -42,6 +42,7 @@ import org.apache.flink.streaming.api.functions.sink.filesystem.BucketWriter;
 import org.apache.flink.streaming.api.functions.sink.filesystem.DeltaBulkBucketWriter;
 import org.apache.flink.streaming.api.functions.sink.filesystem.OutputFileConfig;
 import org.apache.flink.streaming.api.functions.sink.filesystem.rollingpolicies.CheckpointRollingPolicy;
+import org.apache.flink.table.types.logical.RowType;
 import org.apache.hadoop.conf.Configuration;
 import static org.apache.flink.util.Preconditions.checkNotNull;
 
@@ -75,11 +76,23 @@ public class DeltaSinkBuilder<IN> implements Serializable {
     private final Path tableBasePath;
 
     /**
+     * Flink's logical type to indicate the structure of the events in the stream
+     */
+    private final RowType rowType;
+
+    /**
      * Unique identifier of the current Flink's app. Value from this builder will be read
      * only during the fresh start of the application. For restarts or failure recovery
      * it will be resolved from the snapshoted state.
      */
     private final String appId;
+
+    /**
+     * Indicator whether we should try to update table's schema with stream's schema in case
+     * those will not match. The update is not guaranteed as there will be still some checks
+     * performed whether the updates to the schema are compatible.
+     */
+    private boolean canTryUpdateSchema;
 
     /**
      * Serializable wrapper for {@link Configuration} object
@@ -113,7 +126,9 @@ public class DeltaSinkBuilder<IN> implements Serializable {
         Configuration conf,
         ParquetWriterFactory<IN> writerFactory,
         BucketAssigner<IN, String> assigner,
-        CheckpointRollingPolicy<IN, String> policy) {
+        CheckpointRollingPolicy<IN, String> policy,
+        RowType rowType,
+        boolean canTryUpdateSchema) {
         this(
             basePath,
             conf,
@@ -122,7 +137,9 @@ public class DeltaSinkBuilder<IN> implements Serializable {
             assigner,
             policy,
             OutputFileConfig.builder().withPartSuffix(".snappy.parquet").build(),
-            generateNewAppId()
+            generateNewAppId(),
+            rowType,
+            canTryUpdateSchema
         );
     }
 
@@ -134,7 +151,9 @@ public class DeltaSinkBuilder<IN> implements Serializable {
         BucketAssigner<IN, String> assigner,
         CheckpointRollingPolicy<IN, String> policy,
         OutputFileConfig outputFileConfig,
-        String appId) {
+        String appId,
+        RowType rowType,
+        boolean canTryUpdateSchema) {
         this.tableBasePath = checkNotNull(basePath);
         this.serializableConfiguration = new SerializableConfiguration(checkNotNull(conf));
         this.bucketCheckInterval = bucketCheckInterval;
@@ -143,6 +162,38 @@ public class DeltaSinkBuilder<IN> implements Serializable {
         this.rollingPolicy = checkNotNull(policy);
         this.outputFileConfig = checkNotNull(outputFileConfig);
         this.appId = appId;
+        this.rowType = rowType;
+        this.canTryUpdateSchema = canTryUpdateSchema;
+    }
+
+    public DeltaSinkBuilder<IN> withRowType(RowType rowType) {
+        return new DeltaSinkBuilder<>(
+            tableBasePath,
+            serializableConfiguration.conf(),
+            bucketCheckInterval,
+            writerFactory,
+            bucketAssigner,
+            rollingPolicy,
+            outputFileConfig,
+            appId,
+            rowType,
+            canTryUpdateSchema);
+    }
+
+    /**
+     * Sets the sink's option whether in case of any differences between stream's schema and Delta
+     * table's schema we should try to update it during commit to the
+     * {@link io.delta.standalone.DeltaLog}. The update is not guaranteed as there will be some
+     * compatibility checks performed.
+     *
+     * @param canTryUpdateSchema whether we should try to update table's schema with stream's schema
+     *                           in case those will not match. See
+     *                           {@link DeltaSinkBuilder#canTryUpdateSchema} for details.
+     * @return builder for {@link DeltaSink}
+     */
+    public DeltaSinkBuilder<IN> withCanTryUpdateSchema(final boolean canTryUpdateSchema) {
+        this.canTryUpdateSchema = canTryUpdateSchema;
+        return this;
     }
 
     DeltaCommitter createCommitter() throws IOException {
@@ -150,7 +201,8 @@ public class DeltaSinkBuilder<IN> implements Serializable {
     }
 
     DeltaGlobalCommitter createGlobalCommitter() {
-        return new DeltaGlobalCommitter(serializableConfiguration.conf(), tableBasePath);
+        return new DeltaGlobalCommitter(
+            serializableConfiguration.conf(), tableBasePath, rowType, canTryUpdateSchema);
     }
 
     Path getTableBasePath() {
@@ -247,8 +299,10 @@ public class DeltaSinkBuilder<IN> implements Serializable {
             final Configuration conf,
             ParquetWriterFactory<IN> writerFactory,
             BucketAssigner<IN, String> assigner,
-            CheckpointRollingPolicy<IN, String> policy) {
-            super(basePath, conf, writerFactory, assigner, policy);
+            CheckpointRollingPolicy<IN, String> policy,
+            RowType rowType,
+            boolean canTryUpdateSchema) {
+            super(basePath, conf, writerFactory, assigner, policy, rowType, canTryUpdateSchema);
         }
     }
 }
--- a/flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSinkBuilder.java
+++ b/flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSinkBuilder.java
@@ -92,7 +92,7 @@ public class DeltaSinkBuilder<IN> implements Serializable {
      * those will not match. The update is not guaranteed as there will be still some checks
      * performed whether the updates to the schema are compatible.
      */
-    private boolean canTryUpdateSchema;
+    private boolean shouldTryUpdateSchema;
 
     /**
      * Serializable wrapper for {@link Configuration} object
@@ -128,7 +128,7 @@ public class DeltaSinkBuilder<IN> implements Serializable {
         BucketAssigner<IN, String> assigner,
         CheckpointRollingPolicy<IN, String> policy,
         RowType rowType,
-        boolean canTryUpdateSchema) {
+        boolean shouldTryUpdateSchema) {
         this(
             basePath,
             conf,
@@ -139,7 +139,7 @@ public class DeltaSinkBuilder<IN> implements Serializable {
             OutputFileConfig.builder().withPartSuffix(".snappy.parquet").build(),
             generateNewAppId(),
             rowType,
-            canTryUpdateSchema
+            shouldTryUpdateSchema
         );
     }
 
@@ -153,7 +153,7 @@ public class DeltaSinkBuilder<IN> implements Serializable {
         OutputFileConfig outputFileConfig,
         String appId,
         RowType rowType,
-        boolean canTryUpdateSchema) {
+        boolean shouldTryUpdateSchema) {
         this.tableBasePath = checkNotNull(basePath);
         this.serializableConfiguration = new SerializableConfiguration(checkNotNull(conf));
         this.bucketCheckInterval = bucketCheckInterval;
@@ -163,7 +163,7 @@ public class DeltaSinkBuilder<IN> implements Serializable {
         this.outputFileConfig = checkNotNull(outputFileConfig);
         this.appId = appId;
         this.rowType = rowType;
-        this.canTryUpdateSchema = canTryUpdateSchema;
+        this.shouldTryUpdateSchema = shouldTryUpdateSchema;
     }
 
     public DeltaSinkBuilder<IN> withRowType(RowType rowType) {
@@ -177,7 +177,7 @@ public class DeltaSinkBuilder<IN> implements Serializable {
             outputFileConfig,
             appId,
             rowType,
-            canTryUpdateSchema);
+            shouldTryUpdateSchema);
     }
 
     /**
@@ -186,13 +186,13 @@ public class DeltaSinkBuilder<IN> implements Serializable {
      * {@link io.delta.standalone.DeltaLog}. The update is not guaranteed as there will be some
      * compatibility checks performed.
      *
-     * @param canTryUpdateSchema whether we should try to update table's schema with stream's schema
-     *                           in case those will not match. See
-     *                           {@link DeltaSinkBuilder#canTryUpdateSchema} for details.
+     * @param shouldTryUpdateSchema whether we should try to update table's schema with stream's
+     *                              schema in case those will not match. See
+     *                              {@link DeltaSinkBuilder#shouldTryUpdateSchema} for details.
      * @return builder for {@link DeltaSink}
      */
-    public DeltaSinkBuilder<IN> withCanTryUpdateSchema(final boolean canTryUpdateSchema) {
-        this.canTryUpdateSchema = canTryUpdateSchema;
+    public DeltaSinkBuilder<IN> withShouldTryUpdateSchema(final boolean shouldTryUpdateSchema) {
+        this.shouldTryUpdateSchema = shouldTryUpdateSchema;
         return this;
     }
 
@@ -202,7 +202,7 @@ public class DeltaSinkBuilder<IN> implements Serializable {
 
     DeltaGlobalCommitter createGlobalCommitter() {
         return new DeltaGlobalCommitter(
-            serializableConfiguration.conf(), tableBasePath, rowType, canTryUpdateSchema);
+            serializableConfiguration.conf(), tableBasePath, rowType, shouldTryUpdateSchema);
     }
 
     Path getTableBasePath() {
@@ -301,8 +301,8 @@ public class DeltaSinkBuilder<IN> implements Serializable {
             BucketAssigner<IN, String> assigner,
             CheckpointRollingPolicy<IN, String> policy,
             RowType rowType,
-            boolean canTryUpdateSchema) {
-            super(basePath, conf, writerFactory, assigner, policy, rowType, canTryUpdateSchema);
+            boolean shouldTryUpdateSchema) {
+            super(basePath, conf, writerFactory, assigner, policy, rowType, shouldTryUpdateSchema);
         }
     }
 }
--- a/flink-connector/src/main/java/org/apache/flink/connector/delta/sink/SchemaConverter.java
+++ b/flink-connector/src/main/java/org/apache/flink/connector/delta/sink/SchemaConverter.java
@@ -31,37 +31,16 @@ import io.delta.standalone.types.*;
 public class SchemaConverter {
 
     /**
-     * Main method for converting from {@link RowType} into {@link StructType}
-     *
-     * @param rowType Flink's logical type of stream's events
-     * @return DeltaLake's specific type of stream's events
-     */
-    public StructType toDeltaFormat(RowType rowType) {
-        StructField[] fields = rowType.getFields()
-                .stream()
-                .map(rowField -> {
-                    DataType rowFieldType = toDeltaDataType(rowField.getType());
-                    return new StructField(
-                            rowField.getName(),
-                            rowFieldType,
-                            rowField.getType().isNullable());
-                })
-                .toArray(StructField[]::new);
-
-        return new StructType(fields);
-    }
-
-    /**
      * Method containing the actual mapping between Flink's and DeltaLake's types.
      *
      * @param flinkType Flink's logical type
      * @return DeltaLake's data type
      */
-    public DataType toDeltaDataType(LogicalType flinkType) {
+    public static DataType toDeltaDataType(LogicalType flinkType) {
         switch (flinkType.getTypeRoot()) {
             case ARRAY:
                 org.apache.flink.table.types.logical.ArrayType arrayType =
-                        (org.apache.flink.table.types.logical.ArrayType) flinkType;
+                    (org.apache.flink.table.types.logical.ArrayType) flinkType;
                 LogicalType flinkElementType = arrayType.getElementType();
                 DataType deltaElementType = toDeltaDataType(flinkElementType);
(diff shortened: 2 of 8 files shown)


flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSinkBuilder.java
```diff
@@ -74,13 +75,25 @@ private static String generateNewAppId() {
...
      * it will be resolved from the snapshoted state.
      */
     private final String appId;
 
+    /**
+     * Indicator whether we should try to update table's schema with stream's schema in case
+     * those will not match. The update is not guaranteed as there will be still some checks
+     * performed whether the updates to the schema are compatible.
+     */
+    private boolean canTryUpdateSchema;
```
tdas, line 95: can -> should

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/SchemaConverter.java
```diff
@@ -0,0 +1,114 @@
...
+            case ARRAY:
+                org.apache.flink.table.types.logical.ArrayType arrayType =
+                        (org.apache.flink.table.types.logical.ArrayType) flinkType;
+                LogicalType flinkElementType = arrayType.getElementType();
+                DataType deltaElementType = toDeltaDataType(flinkElementType);
+                return new ArrayType(deltaElementType, flinkElementType.isNullable());
+            case BIGINT:
+                return new LongType();
+            case BINARY:
+            case VARBINARY:
```
tdas, line 71: continuing the conversation here, what is the difference between flink's  BINARY and VARBINARY? why are they being mapped to the same Spark type?

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/SchemaConverter.java
```diff
@@ -0,0 +1,114 @@
...
+                DataType valueType = toDeltaDataType(mapType.getValueType());
+                boolean valueCanContainNull = mapType.getValueType().isNullable();
+                return new MapType(keyType, valueType, valueCanContainNull);
+            case NULL:
+                return new NullType();
+            case SMALLINT:
+                return new ShortType();
+            case TIMESTAMP_WITHOUT_TIME_ZONE:
+            case TIMESTAMP_WITH_TIME_ZONE:
+            case TIMESTAMP_WITH_LOCAL_TIME_ZONE:
```
tdas, line 100: same qs as above

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -245,6 +269,59 @@ private void doCommit(OptimisticTransaction transaction,
...
+     * <p>
+     *
+     * @param currentTableVersion current version of the table's snapshot
+     * @param transaction         DeltaLog's transaction object
+     */
+    private void handleMetadataUpdate(long currentTableVersion,
+                                      OptimisticTransaction transaction) {
+        Metadata transactionMetadata = transaction.metadata();
+
+        Metadata metadataAction = prepareMetadata(transactionMetadata);
```
tdas, line 298: can we call these two metadata objects consistently but clearly differentiating them - currentMetadata and updatedMetadata?

tdas: also.. i dont get this. why are you cloning the metadata object and use that compare the schema. why not directly first compare the schema, and if they are different, then create the metadata object with the new schema? thats a lot more intuitive to read code. because as of now, its not obvious just by reading this that metadataAction has the new schema generated from rowType.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -245,6 +269,59 @@ private void doCommit(OptimisticTransaction transaction,
...
+                : commitSchema.toPrettyJson();
+
+            throw new RuntimeException(
+                "DataStream's schema is different from current table's schema. \n" +
+                    "provided: " + printableCurrentTableSchema + "\n" +
+                    "is different from: " + printableCommitSchema);
+        }
+    }
+
+    private boolean areSchemasEqual(@Nullable StructType first,
```
tdas, line 317: @scottsand-db @zsxwing arent StructType objects directly comparable? shouldnt we override their hashcode and make them directly comparable?

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTestParametrized.java
```diff
@@ -0,0 +1,142 @@
...
+    @ClassRule
+    public static final TemporaryFolder TEMPORARY_FOLDER = new TemporaryFolder();
+
+    @Parameterized.Parameters(
+        name = "canTryUpdateSchema = {0}, " +
+            "initializeTableBeforeCommit = {1}"
+    )
+    public static Collection<Object[]> params() {
+        return Arrays.asList(
+            new Object[]{false, false},
```
tdas, line 62:
```suggestion
            // canTryUpdateSchema, initializeTableBeforeCommit
            new Object[]{false, false},
```

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTestParametrized.java
```diff
@@ -0,0 +1,142 @@
...
+
+    @Before
+    public void setup() throws IOException {
+        tablePath = new Path(TEMPORARY_FOLDER.newFolder().toURI());
+        if (initializeTableBeforeCommit) {
+            DeltaSinkTestUtils.initTestForNonPartitionedTable(
+                tablePath.getPath());
+        }
+        deltaLog = DeltaLog.forTable(DeltaSinkTestUtils.getHadoopConf(), tablePath.getPath());
+        expectedTableVersionAfterCommit = initializeTableBeforeCommit ? 1 : 0;
```
tdas, line 89: why set this here and not just infer it the same way in the validate function? the logic stays closer to where its used.

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTestParametrized.java
```diff
@@ -0,0 +1,142 @@
...
+        List<DeltaGlobalCommittable> globalCommittables =
+            Collections.singletonList(new DeltaGlobalCommittable(deltaCommittables));
+
+        // WHEN
+        globalCommitter.commit(globalCommittables);
+
+        // THEN
+        validateCurrentSnapshotState(
+            deltaLog,
+            expectedTableVersionAfterCommit,
```
tdas, line 113: no need to pass this

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTestParametrized.java
```diff
@@ -0,0 +1,142 @@
...
+            deltaCommittables.size(),
+            initializeTableBeforeCommit
+        );
+        validateCurrentTableFiles(deltaLog.update());
+    }
+
+    private static void validateCurrentSnapshotState(DeltaLog deltaLog,
+                                                     int expectedTableVersionAfterUpdate,
+                                                     int deltaCommittablesSize,
+                                                     boolean initializeTableBeforeCommit) {
```
tdas, line 123: why is this function static? seems like you are passing all of this because its a static function?

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTestParametrized.java
```diff
@@ -0,0 +1,142 @@
...
+            expectedTableVersionAfterCommit,
+            deltaCommittables.size(),
+            initializeTableBeforeCommit
+        );
+        validateCurrentTableFiles(deltaLog.update());
+    }
+
+    private static void validateCurrentSnapshotState(DeltaLog deltaLog,
+                                                     int expectedTableVersionAfterUpdate,
+                                                     int deltaCommittablesSize,
```
tdas, line 122: more intuitive to call this numFilesAdded

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTestParametrized.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    @Parameterized.Parameter(0)
+    public boolean canTryUpdateSchema;
+
+    @Parameterized.Parameter(1)
+    public boolean initializeTableBeforeCommit;
+
+    private int expectedTableVersionAfterCommit;
+    private RowType testRowType;
```
tdas, line 76: rowTypeToCommit

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTestParametrized.java
```diff
@@ -0,0 +1,142 @@
...
+                                                     int expectedTableVersionAfterUpdate,
+                                                     int deltaCommittablesSize,
+                                                     boolean initializeTableBeforeCommit) {
+        int initialTableFilesCount = 0;
+        if (initializeTableBeforeCommit) {
+            initialTableFilesCount = deltaLog.snapshot().getAllFiles().size();
+        }
+        Snapshot snapshot = deltaLog.update();
+        assertEquals(snapshot.getVersion(), expectedTableVersionAfterUpdate);
+        assertEquals(snapshot.getAllFiles().size(), deltaCommittablesSize + initialTableFilesCount);
```
tdas, line 130: you are not testing the schema after committing? when it updated the schema correctly?

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTest.java
```diff
@@ -123,34 +195,31 @@ public void testCommittablesFromDifferentCheckpointInterval() throws IOException
...
         // validation unambiguous
-        DeltaSinkTestUtils.initTestForNonPartitionedTable(tablePath.getPath());
         int numAddedFiles1FirstTrial = 3;
         int numAddedFiles1SecondTrial = 4;
         int numAddedFiles2 = 10;
         DeltaLog deltaLog = DeltaLog.forTable(
             DeltaSinkTestUtils.getHadoopConf(), tablePath.getPath());
-        int initialTableFilesCount = deltaLog.snapshot().getAllFiles().size();
-        assertEquals(0, deltaLog.snapshot().getVersion());
+        assertEquals(-1, deltaLog.snapshot().getVersion());
```
tdas, line 216: why these changes?

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/TestSchemaConverter.java
```diff
@@ -0,0 +1,196 @@
...
+import static org.junit.Assert.assertEquals;
+
+import io.delta.standalone.types.DataType;
+import io.delta.standalone.types.StructField;
+import io.delta.standalone.types.StructType;
+
+public class TestSchemaConverter {
+
+    @Test
+    public void testConvertFlinkSchemaToIcebergSchema() {
```
tdas, line 36: errrr... iceberg?

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/SchemaConverter.java
pkubit-g: I've followed this [this](https://nightlies.apache.org/flink/flink-docs-master/docs/connectors/table/formats/parquet/#data-type-mapping) mapping

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/SchemaConverter.java
pkubit-g: same answer, i 've followed [this](mapping) mapping that states that for Parquet all timestamp types (no matter the precision) will be mapped to int96. So my understanding is that we either map them to `io.delta.standalone.TimestampType` or just support only `TIMESTAMP_WITHOUT_TIME_ZONE` and throw an exception for other.

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/TestSchemaConverter.java
pkubit-g: wow, sorry for that ! i recall checking what's schema utils's interfaces in Iceberg and also how they implemented test cases and must have copied the test name ...

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
pkubit-g: you are totally right. I didn't noticed that after last small refactor this code got a little bit confusing. Corrected

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/SchemaConverter.java
```diff
@@ -0,0 +1,114 @@
...
+ */
+public class SchemaConverter {
+
+    /**
+     * Main method for converting from {@link RowType} into {@link StructType}
+     *
+     * @param rowType Flink's logical type of stream's events
+     * @return DeltaLake's specific type of stream's events
+     */
+    public StructType toDeltaFormat(RowType rowType) {
```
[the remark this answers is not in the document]
pkubit-g: basically the one place i need it as a `StructType` is the one line in `DeltaGlobalCommitter`. So wouldn't it just be cleaner to leave the interface:
`public DataType toDeltaDataType(LogicalType flinkType) {`
and in that one place cast the result ?
`return (StructType) new SchemaConverter().toDeltaDataType(rowType);`

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/SchemaConverter.java
```diff
@@ -0,0 +1,114 @@
...
+        return new StructType(fields);
+    }
+
+    /**
+     * Method containing the actual mapping between Flink's and DeltaLake's types.
+     *
+     * @param flinkType Flink's logical type
+     * @return DeltaLake's data type
+     */
+    public DataType toDeltaDataType(LogicalType flinkType) {
```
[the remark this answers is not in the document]
pkubit-g: changed

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTestParametrized.java
pkubit-g: changed

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTestParametrized.java
pkubit-g: fixed

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTestParametrized.java
pkubit-g: fixed

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTestParametrized.java
pkubit-g: fixed

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTestParametrized.java
pkubit-g: fixed

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTestParametrized.java
pkubit-g: fixed

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTestParametrized.java
```diff
@@ -0,0 +1,142 @@
...
+import io.delta.standalone.DeltaLog;
+import io.delta.standalone.Snapshot;
+import io.delta.standalone.actions.AddFile;
+import io.delta.standalone.data.CloseableIterator;
+
+/**
+ * Tests for {@link DeltaGlobalCommitter}.
+ */
+@RunWith(Parameterized.class)
+public class DeltaGlobalCommitterTestParametrized {
```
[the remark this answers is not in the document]
pkubit-g: in the parametrized version i'm just testing all of the combinations of 
- shouldTryUpdateSchema
- is table initialized
- is table partitioned (next PR)
and whether global committer will correctly behave in each of those combinations.

The tests in non-parametrized are not exhaustive for all of those cases and there i'm focusing on edge and special cases that need to be covered. Would need to add much more repetitive test cases to cover all of those possiblities. Also, the reason is splitted into 2 java classes is that junit 4 does not support test-method level parametrization.

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTest.java
pkubit-g: it's just a matter of jumping between different versions of the sink - in the final one this version of the test does not initialize the table, but i needed to add it specifically for PR4 since it didn't support writing to not existing table.

The test itself is not affected in any way, as it's irrelevant for this scenario whether the table is initialized or not, however it removes redundant code that may hinder understanding what's going on in the test.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/SchemaConverter.java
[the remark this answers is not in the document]
tdas: I am not so sure. What is returned should be the most intuitively typed. if the input param of type RowType then the function can only return a StructType and not some other DataType. If the param is of a more specific type than LogicalType, then why not the return type be specifically typed as well? So i think we should return structtype here. And the name can make it very clear what is being returned .. by naming this method toDeltaStructType.

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTestParametrized.java
```diff
@@ -0,0 +1,137 @@
...
+            SchemaConverter.toDeltaDataType(rowTypeToCommit).toJson());
+    }
+
+    private void validateCurrentTableFiles(
+        Snapshot snapshot) {
+        CloseableIterator<AddFile> filesIterator = snapshot.scan().getFiles();
+        while (filesIterator.hasNext()) {
+            AddFile addFile = filesIterator.next();
+            assertTrue(addFile.getSize() > 0);
+            assertFalse(addFile.getPath().isEmpty());
```
tdas, line 134: why empty? whats the point of testing this?

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/utils/DeltaSinkTestUtils.java
```diff
@@ -91,6 +91,23 @@
         return rows;
     }
 
+    public static RowType addNewColumnToSchema() {
```
tdas, line 94: confusing. method says add new column to some schema... but you are not passing any schema. ultimately you are  just generating another constant schema object... in which case why not just define TEST_ROW_TYPE_EXTRA_COL

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/utils/DeltaSinkTestUtils.java
```diff
@@ -91,6 +91,23 @@
...
+        List<RowType.RowField> fields = new ArrayList<>(
+            DeltaSinkTestUtils.TEST_ROW_TYPE.getFields());
+        fields.add(new RowType.RowField("someNewField", new IntType()));
+        return new RowType(fields);
+    }
+
+    public static RowType dropOneColumnFromSchema() {
```
tdas, line 101: same comment as above.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -264,6 +341,38 @@ private void doCommit(OptimisticTransaction transaction,
...
+            .build();
+    }
+
+    /**
+     * This method transforms {@link RowType} object containing schema for the current stream
+     * to Delta's {@link StructType}.
+     *
+     * @return {@link StructType} object corresponding to the schema of the events in the stream
+     */
+    private StructType getStreamSchema() {
```
tdas, line 372: its really not useful to have one line function like this. just increases code bloat.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
scottsand-db: @tdas we have exposed this: https://github.com/delta-io/connectors/pull/232/files

scottsand-db: We will update to the latter.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/SchemaConverter.java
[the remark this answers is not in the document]
pkubit-g: ok, so for now i've followed your original advice and named both functions `toDeltaDataType`.
Let me know if you actually prefer `toDeltaStructType` or any other solution.
Screenshot from 2021-12-11 15-13-18

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/TestSchemaConverter.java
pkubit-g: fixed

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -245,6 +269,59 @@ private void doCommit(OptimisticTransaction transaction,
...
+     * @param transaction DeltaLog's transaction object
+     */
+    private void handleMetadataUpdate(boolean tableExists,
+                                      OptimisticTransaction transaction) {
+        Metadata currentMetadata = transaction.metadata();
+        StructType currentTableSchema = currentMetadata.getSchema();
+        StructType streamSchema = getStreamSchema();
+        boolean schemasAreMatching = areSchemasEqual(currentTableSchema, streamSchema);
+        if (!tableExists || (!schemasAreMatching && shouldTryUpdateSchema)) {
+            Metadata updatedMetadata = updateMetadata(currentMetadata);
```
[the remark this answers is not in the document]
pkubit-g: fixed

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTestParametrized.java
pkubit-g: i've put it here to make sure that we're not committing `AddFile`s with empty path, as it's allowed to put an empty string to the [AddFile](https://github.com/delta-io/connectors/blob/master/standalone/src/main/java/io/delta/standalone/actions/AddFile.java) constructr/builder. 
But i've just tested and it seems that the Scala API for the delta standalone performs this check [inhere](https://github.com/delta-io/connectors/blob/master/standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala#L119) before the commit, so i'm removing this validation from my tests.  
btw seems like you can consider put this check to the `AddFile` builder for Java API as well.

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/utils/DeltaSinkTestUtils.java
pkubit-g: fixed

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/utils/DeltaSinkTestUtils.java
pkubit-g: fixed

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
pkubit-g: removed

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTestParametrized.java
tdas: this check is in the AddFile class constructor. so it will be invoked even if the java builder is creating hits object.

tdas (approved):

PR #231, opened 2021-12-10
Create a Metadata builder from an existing instance

allisonport-db:
Since actions are immutable, it is cumbersome to update an existing table metadata with one or two value changes. This introduces `Metadata::copyBuilder()` which instantiates a `Metadata.Builder` with properties copied over.
- Added `Metadata::copyBuilder()`
- Added tests
- Updated `OptimisticTransaction::updateMetadata(Metadata metadata)` documentation with example usage

### DIFF
--- a/standalone/src/main/java/io/delta/standalone/OptimisticTransaction.java
+++ b/standalone/src/main/java/io/delta/standalone/OptimisticTransaction.java
@@ -80,6 +80,18 @@ public interface OptimisticTransaction {
      * Records an update to the metadata that should be committed with this transaction.
      *
      * <p>
+     * Use {@link Metadata#copyBuilder()} to build a new {@link Metadata} instance based on the
+     * current table metadata. For example:
+     *
+     * <pre>{@code
+     * Metadata newMetadata = optimisticTransaction.metadata().copyBuilder()
+     *     .schema(newSchema)
+     *     .createdTime(System.currentTimeMillis())
+     *     .build();
+     * optimisticTransaction.updateMetadata(newMetadata);
+     * }</pre>
+     *
+     * <p>
      * IMPORTANT: It is the responsibility of the caller to ensure that files currently
      * present in the table are still valid under the new metadata.
      *
--- a/standalone/src/main/java/io/delta/standalone/OptimisticTransaction.java
+++ b/standalone/src/main/java/io/delta/standalone/OptimisticTransaction.java
@@ -86,7 +86,6 @@ public interface OptimisticTransaction {
      * <pre>{@code
      * Metadata newMetadata = optimisticTransaction.metadata().copyBuilder()
      *     .schema(newSchema)
-     *     .createdTime(System.currentTimeMillis())
      *     .build();
      * optimisticTransaction.updateMetadata(newMetadata);
      * }</pre>
--- a/standalone/src/main/java/io/delta/standalone/actions/Metadata.java
+++ b/standalone/src/main/java/io/delta/standalone/actions/Metadata.java
@@ -155,6 +155,22 @@ public final class Metadata implements Action {
     }
 
     /**
+     * @return a new {@link Metadata.Builder} initialized with the same properties as this
+     *         {@link Metadata} instance
+     */
+    public Builder copyBuilder() {
+        return builder()
+                .id(id)
+                .name(name)
+                .description(description)
+                .format(format)
+                .partitionColumns(partitionColumns)
+                .configuration(configuration)
+                .createdTime(createdTime)
+                .schema(schema);
+    }
+
+    /**
      * @return a new {@link Metadata.Builder}
      */
     public static Builder builder() {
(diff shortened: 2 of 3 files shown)


standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -210,6 +226,11 @@ public Builder createdTime(Long createdTime) {
             return this;
         }
 
+        public Builder createdTime(Optional<Long> createdTime) {
```
scottsand-db, line 229: nit: `@Nonnull`

scottsand-db, line 229: nit: `@Nonnull`

scottsand-db (approved):
left one comment but otherwise LGTM!

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -154,6 +154,22 @@ public int hashCode() {
...
+     */
+    public Builder copyBuilder() {
+        return builder()
+                .id(id)
+                .name(name)
+                .description(description)
+                .format(format)
+                .partitionColumns(partitionColumns)
+                .configuration(configuration)
+                .createdTime(createdTime)
```
scottsand-db, line 169: in response to your PR description, this can be `System.currentTimeMillis()`, and then update the description of this method, too.

standalone/src/main/java/io/delta/standalone/OptimisticTransaction.java
```diff
@@ -80,6 +80,18 @@
      * Records an update to the metadata that should be committed with this transaction.
      *
      * <p>
+     * Use {@link Metadata#copyBuilder()} to build a new {@link Metadata} instance based on the
+     * current table metadata. For example:
+     *
+     * <pre>{@code
+     * Metadata newMetadata = optimisticTransaction.metadata().copyBuilder()
+     *     .schema(newSchema)
+     *     .createdTime(System.currentTimeMillis())
```
zsxwing, line 89: nit: `createdTime` is usually not changed after creating the table. It's better to remove this from javadoc.

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -154,6 +154,22 @@ public int hashCode() {
                 createdTime, schema);
     }
 
+    /**
+     * @return a new {@link Metadata.Builder} initialized with the same properties as this
+     *         {@link Metadata} instance
+     */
+    public Builder copyBuilder() {
+        return builder()
```
zsxwing, line 162: If we add a new field to Metadata in future, it's pretty easy to forget to add it here. Alternatively, we can add a constructor in Metadata.Builder such as 

```
Builder(
            @Nonnull String id,
            @Nullable String name,
            @Nullable String description,
            @Nonnull Format format,
            @Nonnull List<String> partitionColumns,
            @Nonnull Map<String, String> configuration,
            @Nonnull Optional<Long> createdTime,
            @Nullable StructType schema) {
 ...
}
```
then call this constructor here.

After that, we can write a unit test and use Java Reflection APIs to verify the numbers of constructor parameters in Metadata and Metadata.Builder are the same.

zsxwing (commented):
Left a comment to propose a safer approach.

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -175,6 +184,27 @@ public static Builder builder() {
         @Nonnull private Optional<Long> createdTime = Optional.of(System.currentTimeMillis());
         @Nullable private StructType schema;
 
+        public Builder(){};
+
+        public Builder(
```
zsxwing, line 189: We can remove `public` here to not expose this and change `getConstructors` to `getDeclaredConstructors` in the test.

standalone/src/test/scala/io/delta/standalone/internal/ActionBuildersSuite.scala
```diff
@@ -62,6 +62,42 @@ class ActionBuildersSuite extends FunSuite {
     assert(metadataFromBuilder == metadataFromConstructor)
   }
 
+  test("Metadata constructor matches Metadata.Builder constructor") {
+    assert(
+      classOf[MetadataJ].getConstructors.map(_.getParameterCount()).toList.max ==
+      classOf[MetadataJ.Builder].getConstructors.map(_.getParameterCount()).toList.max
```
zsxwing, line 68: nit: could you add a better message, such as
```
Metadata and Metadata.Builder's constructors are not the same. Please update them accordingly if you add a new field to Metadata.
```

zsxwing (commented):
Two minor comments

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
allisonport-db: Do we want the empty constructor to also be private? For the other actions the constructors are public (should I make them private as well?)

zsxwing: That's fine. The empty constructor is easy to maintain the backward compatibility.

zsxwing (approved):
LGTM

PR #232, opened 2021-12-10
Add StructType.isWriteCompatible

allisonport-db:
Add `StructType.isWriteCompatible` to allow users to check whether a new schema can be used in a Delta table safely.


scottsand-db (approved):
Thanks for this! LGTM

zsxwing (commented):
It's better to add this method as an instance method of StructType rather than adding a new utils class.

zsxwing (approved):
The change looks good to me. Could you also update `isWriteCompatible` usages in our code to use this new method rather than `SchemaUtils.isWriteCompatible`? This would make our tests cover this new public API.

zsxwing (approved):

PR #233, opened 2021-12-13
Flink Delta Sink - PR 6

pkubit-g:
PR added support for partitioned tables for flink-connector


flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaTablePartitionAssigner.java
```diff
@@ -0,0 +1,92 @@
...
+
+/**
+ * Custom implementation of {@link BucketAssigner} class required to provide behaviour on how
+ * to map particular events to buckets (aka partitions).
+ * <p>
+ * This implementation can be perceived as a utility class for complying to the DeltaLake's
+ * partitioning style (that follows Apache Hive's partitioning style by providing the partitioning
+ * column's and its values as FS directories paths, e.g. "/some_path/table_1/date=2020-01-01")
+ * It's still possible for users to roll out their own version of {@link BucketAssigner}
+ * and pass it to the {@link DeltaSinkBuilder} during creation of the sink.
```
tdas, line 37: users put their own bucketAssigner .... how will it interact with partition values? 

also why do we need to allow users to specify their own bucket assigners? what kind of bucket assigners do they use?

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -245,26 +248,49 @@ private void doCommit(OptimisticTransaction transaction,
...
+            if (partitionColumnsSet == null) {
+                partitionColumnsSet = currentPartitionCols;
+            }
+            boolean isPartitionColumnsMetadataRetained = compareKeysOfLinkedSets(
+                currentPartitionCols,
+                partitionColumnsSet);
+            if (!isPartitionColumnsMetadataRetained) {
+                throw new RuntimeException(
+                    "Partition columns cannot differ for files in the same checkpointId. " +
+                        "checkpointId=" + checkpointId + " " +
```
tdas, line 269:
```suggestion
                        "checkpointId = " + checkpointId + " " +
```

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -245,26 +248,49 @@ private void doCommit(OptimisticTransaction transaction,
...
+                        "Partition spec " + deltaPendingFile.getPartitionSpec() +
```
tdas, line 270:
```suggestion
                        "partition spec = " + deltaPendingFile.getPartitionSpec() +
```

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -416,6 +457,24 @@ private Operation prepareDeltaLogOperation(Map<String, String> operationMetrics)
...
     }
 
+    /**
+     * Simple method for comparing the order and equality of keys in two linked sets
+     *
+     * @param first  instance of linked set to be compared
+     * @param second instance of linked set to be compared
+     * @return result of the comparison on order and equality of provided sets
+     */
+    private boolean compareKeysOfLinkedSets(Set<String> first, Set<String> second) {
```
tdas, line 467: why do you need this? doesnt Set.equals work? and why do you need the order to be same? As long as the contents are same .. it should be fine.. isnt it?

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -288,13 +314,23 @@ private void doCommit(OptimisticTransaction transaction,
...
      */
     private void handleMetadataUpdate(boolean tableExists,
-                                      OptimisticTransaction transaction) {
+                                      OptimisticTransaction transaction,
+                                      List<String> partitionColumns) {
         Metadata currentMetadata = transaction.metadata();
+        if (tableExists && (!partitionColumns.equals(currentMetadata.getPartitionColumns()))) {
+            throw new RuntimeException(
+                "Stream's partition columns are different from table's partitions columns. \n" +
+                    "provided: " + Arrays.toString(partitionColumns.toArray()) + "\n" +
```
tdas, line 329:
```suggestion
                    "Columns in data files: " + Arrays.toString(partitionColumns.toArray()) + "\n" +
```

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -288,13 +314,23 @@ private void doCommit(OptimisticTransaction transaction,
...
+                    "is different from: " +
```
tdas, line 330:
```suggestion
                    "Columns in table: " +
```

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -353,14 +388,20 @@ private boolean areSchemasEqual(@Nullable StructType first,
...
      */
-    private Operation prepareDeltaLogOperation(Map<String, String> operationMetrics) {
+    private Operation prepareDeltaLogOperation(List<String> partitionColumns,
+                                               Map<String, String> operationMetrics) {
         Map<String, String> operationParameters = new HashMap<>();
         try {
             ObjectMapper objectMapper = new ObjectMapper();
             operationParameters.put("mode", objectMapper.writeValueAsString(APPEND_MODE));
+            // we need to perform mapping to JSON object twice for partition columns. First to map
+            // the list to string type and then again to make this string JSON encoded
```
tdas, line 402: can write an example of the string produced here to make it more clear

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaTablePartitionAssigner.java
pkubit-g: because the stream itself is generic and we cannot determine which events goes into which partition during creation of the sink. Also, Flink's DataStream API also does not carry information about partitioning (in contrast to Flink's Table API).

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSink.java
```diff
@@ -230,6 +231,31 @@ private long restoreOrGetNextCheckpointId(List<DeltaWriterBucketState> states) {
...
+     *                 It is advised to use {@link DeltaTablePartitionAssigner} for most of the
+     *                 cases, however for advanced partition assignment behaviour users can provide
+     *                 their own implementation of {@link BucketAssigner}.
+     * @return builder for the DeltaSink
+     */
+    public static DeltaSinkBuilder<RowData> forDeltaFormat(
+        final Path basePath,
+        final Configuration conf,
+        final RowType rowType,
+        BucketAssigner<RowData, String> assigner
```
tdas, line 258: will the user have to explicitly create DeltaTablePartitionAssigner objects and pass it on here? thats a very verbose API. we need to make this common thing simple

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -245,26 +248,49 @@ private void doCommit(OptimisticTransaction transaction,
...
         for (DeltaCommittable deltaCommittable : committables) {
             DeltaPendingFile deltaPendingFile = deltaCommittable.getDeltaPendingFile();
             AddFile action = deltaPendingFile.toAddFile();
             addFileActions.add(action);
 
+            Set<String> currentPartitionCols = deltaPendingFile.getPartitionSpec().keySet();
+            if (partitionColumnsSet == null) {
+                partitionColumnsSet = currentPartitionCols;
+            }
+            boolean isPartitionColumnsMetadataRetained = compareKeysOfLinkedSets(
```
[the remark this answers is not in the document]
pkubit-g: it's a simple check whether provided bucket assigner does not have inconsistent behaviour, so by `metadata retained within   checkpoing interval` i mean that all of the provided committables (representing written files) have the same partition spec.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
pkubit-g: will do a re-check for `Set.equals` behaviour for LinkedSet implementation as lately i've been skipping between different languages and don't remember.
The order is important i guess because those keys are generated from the partition to value mappings, so `a=2/b=5` should generate `LinkedSet(a,b)` and `b=2/a=5`  will generate `LinkedSet(b,a)` which is not the same.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/DeltaSink.java
pkubit-g: i think that it relates to [this](https://github.com/delta-io/connectors/pull/233#discussion_r768782711) discussion so i'm not sure if we can make it less verbose for DataStream API

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
tdas: i dont get it. "a=2/b=5" should be same as "b=5/a=2" right? so the order of the key string "a" and "b" should not matter as long as both of them exists in the set and nothing else exists.

what am i missing.

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTest.java
```diff
@@ -319,6 +385,13 @@ private DeltaGlobalCommitter getTestGlobalCommitter(RowType schema) {
         );
     }
```
tdas, line 387: can you make sure that there is a test for partition column ordering not causing a problem?

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/utils/DeltaSinkTestUtils.java
```diff
@@ -112,13 +120,27 @@ public static RowType dropOneColumnFromSchema(RowType schema) {
         return new LinkedHashMap<>();
     }
 
+    public static LinkedHashMap<String, String> getTestPartitionSpec() {
+        return new LinkedHashMap<String, String>() {{
+                put("a", "b");
```
tdas, line 125: can these be named more intuitively? like "col1" -> "val1", "col2" -> "val2"

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -245,26 +248,49 @@ private void doCommit(OptimisticTransaction transaction,
...
+                currentPartitionCols,
+                partitionColumnsSet);
+            if (!isPartitionColumnsMetadataRetained) {
+                throw new RuntimeException(
+                    "Partition columns cannot differ for files in the same checkpointId. " +
+                        "checkpointId=" + checkpointId + " " +
+                        "Partition spec " + deltaPendingFile.getPartitionSpec() +
+                        " does not comply with partition columns from other committables: " +
+                        partitionColumnsSet
```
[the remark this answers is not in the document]
pkubit-g: well ... to much Python lately

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
pkubit-g: e.g.
if we a file_1 in path `/some/table/path/a=1/b=2/` then it will have partitionSpec = LinkedHashSet("a" -> 1, "b" ->2) which would not be the same as for file_2 in path `/some/table/path/b=2/a=1/`

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitterTest.java
pkubit-g: well, partition column ordering is relevant and if it wouldn't be implemented as `LinkedSet/LinkedMap` then many tests could be randomly failing. 
But i've added a tests that expects a runtime exception when the partition column ordering in committable is different from from the table's.
In table with have `test-data/test-partitioned-delta-table-initial-state/a=b/c=d` and for test i've used
```
LinkedHashMap<String, String> partitionSpec = new LinkedHashMap<String, String>() {{
            put("c", "d");
            put("a", "b");
        }};
```
which fails during commit.

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/utils/DeltaSinkTestUtils.java
pkubit-g: sure. But it will require to regenerate "test tables" from resources and may influence all the tests where we initialize test table, so if you don't mind then i will provide this change in a complementary PR or tomorrow as you prefer as i won't be manage to pull it today.

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
tdas: no you are right. i was confused and I went back to the spec to verify that the partition column ordering matters. I was confused a bit because in the AddFile's json representation, the parttition values are stored as an unordered map. I forgot that the schema stores the partition columns as an ordered array. so the order matters.

flink-connector/src/test/java/org/apache/flink/connector/delta/sink/utils/DeltaSinkTestUtils.java
tdas: sounds good to me. please make that PR

tdas (approved):

flink-connector/src/main/java/org/apache/flink/connector/delta/sink/committer/DeltaGlobalCommitter.java
```diff
@@ -288,27 +316,30 @@ private void doCommit(OptimisticTransaction transaction,
...
-                .id(currentMetadata.getId())
-                .name(currentMetadata.getName())
-                .description(currentMetadata.getDescription())
-                .format(currentMetadata.getFormat())
-                // below line will be changed in the next PR
-                .partitionColumns(currentMetadata.getPartitionColumns())
-                .configuration(currentMetadata.getConfiguration())
+            Metadata updatedMetadata = currentMetadata.copyBuilder()
                 .schema(streamSchema)
+                .partitionColumns(partitionColumns)
```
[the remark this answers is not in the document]
pkubit-g: what about very first commit to a table ?
When i remove this line then some of the tests are failing because of this.
I you prefer then i can provide a change as in the below snippet so will be adding partition cols to the metadata only when the table hasn't been initialized yet.
Screenshot from 2021-12-16 17-09-05

PR #234, opened 2021-12-14
OptimisticTransactionImpl.commit changes absolute paths to relative paths

### LINKED ISSUE #228: OptimisticTransactionImpl.commit should turn absolute paths to relative paths if possible
When users call `OptimisticTransactionImpl.commit`, they may use absolution paths in `AddFile/RemoveFile`. It would be great to change absolute paths to relative paths if these paths are under the table path.

allisonport-db:
Resolves #228

`DeltaLogSuite` already has a test to check that `RemoveFile` paths are not relativized.

Re the last test added: Should we be qualifying paths if they're not in the table path?
- `tryRelativizePath` qualifies any absolute path whether or not it's in the table

### DIFF
--- a/standalone/src/main/java/io/delta/standalone/OptimisticTransaction.java
+++ b/standalone/src/main/java/io/delta/standalone/OptimisticTransaction.java
@@ -36,8 +36,8 @@ public interface OptimisticTransaction {
      * concurrent writer this method will throw an exception.
      * <p>
      * Note: any {@link io.delta.standalone.actions.AddFile} with an absolute path within the table
-     * path will be updated to have a relative path (based off of the table path). Because of this, be
-     * sure to generate all {@link io.delta.standalone.actions.RemoveFile}s from
+     * path will be updated to have a relative path (based off of the table path). Because of this,
+     * be sure to generate all {@link io.delta.standalone.actions.RemoveFile}s from
      * {@link io.delta.standalone.actions.AddFile}s read from the Delta Log, and not the
      * {@link io.delta.standalone.actions.AddFile}s created pre-commit.
      *
--- a/standalone/src/main/java/io/delta/standalone/OptimisticTransaction.java
+++ b/standalone/src/main/java/io/delta/standalone/OptimisticTransaction.java
@@ -34,6 +34,12 @@ public interface OptimisticTransaction {
      * Modifies the state of the log by adding a new commit that is based on a read at the table's
      * latest version as of this transaction's instantiation. In the case of a conflict with a
      * concurrent writer this method will throw an exception.
+     * <p>
+     * Note: any {@link io.delta.standalone.actions.AddFile} with an absolute path within the table
+     * path will be updated to have a relative path (based off of the table path). Because of this, be
+     * sure to generate all {@link io.delta.standalone.actions.RemoveFile}s from
+     * {@link io.delta.standalone.actions.AddFile}s read from the Delta Log, and not the
+     * {@link io.delta.standalone.actions.AddFile}s created pre-commit.
      *
      * @param <T>  A derived class of {@link Action}. This allows, for example, both a
      *             {@code List<Action>} and a {@code List<AddFile>} to be accepted.
--- a/standalone/src/main/java/io/delta/standalone/OptimisticTransaction.java
+++ b/standalone/src/main/java/io/delta/standalone/OptimisticTransaction.java
@@ -37,9 +37,9 @@ public interface OptimisticTransaction {
      * <p>
      * Note: any {@link io.delta.standalone.actions.AddFile} with an absolute path within the table
      * path will be updated to have a relative path (based off of the table path). Because of this,
-     * be sure to generate all {@link io.delta.standalone.actions.RemoveFile}s from
-     * {@link io.delta.standalone.actions.AddFile}s read from the Delta Log, and not the
-     * {@link io.delta.standalone.actions.AddFile}s created pre-commit.
+     * be sure to generate all {@link io.delta.standalone.actions.RemoveFile}s using
+     * {@link io.delta.standalone.actions.AddFile}s read from the Delta Log (do not use the
+     * {@link io.delta.standalone.actions.AddFile}s created pre-commit.)
      *
      * @param <T>  A derived class of {@link Action}. This allows, for example, both a
      *             {@code List<Action>} and a {@code List<AddFile>} to be accepted.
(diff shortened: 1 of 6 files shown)


scottsand-db:
> Re the last test added: Should we be qualifying paths if they're not in the table path?

What you've currently done LGTM, but I'm sure Ryan can confirm. Nonetheless, you can also double check what Delta OSS does. i.e. you can write a temporary test in OssCompatibilitySuite

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
```diff
@@ -54,7 +54,7 @@ private[internal] class DeltaLogImpl private(
   lazy val store = createLogStore(hadoopConf)
 
   /** Direct access to the underlying storage system. */
-  protected lazy val fs = logPath.getFileSystem(hadoopConf)
+  protected[internal] lazy val fs = logPath.getFileSystem(hadoopConf)
```
scottsand-db, line 57: `protected[internal]` should only make this visible to child classes (and traits) in this package, and since `OptimisticTransactionImpl` isn't a child, I'm not sure why this is compiling. (do you?)

I think we can just remove `protected[internal]` and make this 'public' to all `internal` classes (since `DeltaLogImpl` is `private[internal]` anyways).

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
```diff
@@ -216,8 +219,20 @@ private[internal] class OptimisticTransactionImpl(
     val customCommitInfo = actions.exists(_.isInstanceOf[CommitInfo])
     assert(!customCommitInfo, "Cannot commit a custom CommitInfo in a transaction.")
 
+    // Convert AddFile paths to relative paths if they're in the table path
+    var finalActions = actions.map {
```
scottsand-db, line 223: why not RemoveFiles too?

scottsand-db: re-read the description. I see the "do not relativize paths in RemoveFiles" test in DeltaLogSuite.

This is a little strange given that the delta protocol states that [RemoveFile.path](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-file-and-remove-file) is:
> A relative path to a file from the root of the table or an absolute path to a file that should be removed from the table. The path is a URI as specified by RFC 2396 URI Generic Syntax, which needs to be decoded to get the file path.

cc @zsxwing

allisonport-db: My interpretation was since `RemoveFile`s should be instantiated from `AddFile`s we don't want to alter the path.

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
allisonport-db: I think protected[internal] is still just package scope (it doesn't compile with just `protected`). But you're right since it's inside `DeltaLogImpl` 'public' is equivalent.

allisonport-db:
> > Re the last test added: Should we be qualifying paths if they're not in the table path?
> 
> What you've currently done LGTM, but I'm sure Ryan can confirm. Nonetheless, you can also double check what Delta OSS does. i.e. you can write a temporary test in OssCompatibilitySuite

I can verify in OssCompatibilitySuite but I'm pretty sure commit in Delta OSS does nothing to the paths. (Do we even provide a public API to create an external table? I think just in `ConvertToDelta`.)

allisonport-db:
Should this behavior be documented in the `OptimisicTransaction::commit`?

Also doesn't this change mean `AddFile.remove` should only be done with `AddFile`s read from the delta log. (And not the `AddFiles` generated pre-commit) - Should we clarify this?

allisonport-db:
test github actions comment

standalone/src/main/java/io/delta/standalone/OptimisticTransaction.java
```diff
@@ -34,6 +34,12 @@
      * Modifies the state of the log by adding a new commit that is based on a read at the table's
      * latest version as of this transaction's instantiation. In the case of a conflict with a
      * concurrent writer this method will throw an exception.
+     * <p>
+     * Note: any {@link io.delta.standalone.actions.AddFile} with an absolute path within the table
+     * path will be updated to have a relative path (based off of the table path). Because of this,
+     * be sure to generate all {@link io.delta.standalone.actions.RemoveFile}s from
+     * {@link io.delta.standalone.actions.AddFile}s read from the Delta Log, and not the
+     * {@link io.delta.standalone.actions.AddFile}s created pre-commit.
```
zsxwing, line 42:
```suggestion
     * {@link io.delta.standalone.actions.RemoveFile}s created pre-commit.
```
Since you need to change this, could you update RemoveFile constructor like this:

```
    /**
     * Users should <b>not</b> construct {@link RemoveFile}s themselves, and should instead use one
     * of the various {@link AddFile#remove()} methods to instantiate the correct {@link RemoveFile}
     * for a given {@link AddFile} instance.
     *
     * @deprecated {@link RemoveFile} should be created from {@link AddFile#remove()} instead.
     */
    @Deprecated
    public RemoveFile(
```
This would show up in IDE at least.

zsxwing (approved):

standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionSuite.scala
```diff
@@ -263,4 +265,48 @@ class OptimisticTransactionSuite
...
+      val txn = log.startTransaction()
+      val addFile = AddFile("/absolute/path/to/file/test.parquet", Map(), 0, 0, true)
+      txn.commit(Metadata() :: addFile :: Nil, op, "test")
+
+      val committedPath = new Path(log.update().getAllFiles.asScala.head.getPath)
+      // Path is fully qualified
+      assert(committedPath.isAbsolute && !committedPath.isAbsoluteAndSchemeAuthorityNull)
+      // Path is unaltered
+      assert(Path.getPathWithoutSchemeAndAuthority(committedPath).toString ==
+        "/absolute/path/to/file/test.parquet")
```
zsxwing, line 309: nit: why not just use `file:///absolute/path/to/file/test.parquet` to check?

PR #236, opened 2021-12-16
Flink Delta Sink - Docs

pkubit-g:


flink-connector/README.md
```diff
@@ -0,0 +1,232 @@
...
+
+[![License](https://img.shields.io/badge/license-Apache%202-brightgreen.svg)](https://github.com/delta-io/connectors/blob/master/LICENSE.txt)
+
+Official Delta Lake connector for [Apache Flink](https://flink.apache.org/).
+
+# Introduction
+
+This is the repository for Apache Flink connector to Delta Lake. It includes
+
+- Flink Delta Sink for directly writing data from Apache Flink's applications to a Delta Lake's
```
tdas, line 11:
```suggestion
- Sink for directly writing data from Apache Flink's applications to a Delta tables
```

flink-connector/README.md
```diff
@@ -0,0 +1,232 @@
...
+  table
+- source for reading Delta Lake's table using Apache Flink
```
tdas, line 13:
```suggestion
- Source for reading Delta Lake's table using Apache Flink (still in progress)
```

flink-connector/README.md
```diff
@@ -0,0 +1,232 @@
...
+
+Please refer to the [official design doc](
+https://docs.google.com/document/d/19CU4eJuBXOwW7FC58uSqyCbcLTsgvQ5P1zoPOPgUSpI/edit)
+if you want to learn more about the internals of the connector.
```
tdas, line 17: No need to have this here. Very few people will need this. i added the link in the flink issue. so we can point to the issue. but do it at the end of the doc. maybe have a "Additional Resources" section

flink-connector/README.md
```diff
@@ -0,0 +1,232 @@
...
+
+<project>
+    <dependencies>
+        <dependency>
+            <groupId>io.delta</groupId>
+            <artifactId>flink-connector</artifactId>
+            <version>0.2.1-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>io.delta</groupId>
```
tdas, line 64: why do you need to explicitly add standalone? that is transitive dependency and should be automatically added.

flink-connector/README.md
```diff
@@ -0,0 +1,232 @@
...
+            <artifactId>delta-standalone_2.12</artifactId>
+            <version>0.2.1-SNAPSHOT</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.flink</groupId>
```
tdas, line 69: again .. these are compile time dependencies .. isnt it? user's shouldnt have to add this unless they want to use a specific version of flink in their application

flink-connector/README.md
```diff
@@ -0,0 +1,232 @@
...
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-client</artifactId>
+            <version>${hadoop-version}</version>
+        </dependency>
+    </dependencies>
+</project>
+```
+
+Scala 2.11:
```
tdas, line 97: wait... if flink depends on standalone which is scala-specific....then even if we release a single jar of flink ... we still have to have different sbt configurations for different versions of scala to bring in the necessary scala-specific standalone...isnt it?

flink-connector/README.md
```diff
@@ -0,0 +1,137 @@
...
+- Sink for directly writing data from Apache Flink's applications to a Delta tables
+- Source for reading Delta Lake's table using Apache Flink (still in progress)
+
+# Building
+
+The project is compiled using [SBT](https://www.scala-sbt.org/1.x/docs/Command-Line-Reference.html).
+
+## Flink connector
+
+Flink Delta Lake Connector is a JVM library to read and write data from Apache Flink applications' to a Delta Lake
```
tdas, line 20: this is repeating the intro once again.. consolidate this in the intro section and simplify the building section

flink-connector/README.md
```diff
@@ -0,0 +1,137 @@
...
+This is the repository for Apache Flink connector to Delta Lake. It includes
+
+- Sink for directly writing data from Apache Flink's applications to a Delta tables
+- Source for reading Delta Lake's table using Apache Flink (still in progress)
+
+# Building
+
+The project is compiled using [SBT](https://www.scala-sbt.org/1.x/docs/Command-Line-Reference.html).
+
+## Flink connector
```
tdas, line 18: this section header is superfluous. this readme is only for Flink connector so .. why have section here

flink-connector/README.md
```diff
@@ -0,0 +1,137 @@
...
+- [Delta Standalone Reader](https://github.com/delta-io/connectors/wiki/Delta-Standalone-Reader)
+- [Delta Standalone Writer](https://github.com/delta-io/connectors/wiki/Delta-Standalone-Writer) <TODO>
+
+### Build commands
+
+- To compile the project, run `build/sbt flinkConnector/compile`
+- To test the project, run `build/sbt flinkConnector/test`
+- To publish the JAR, run `build/sbt flinkConnector/publishM2`
+
+### How to use it
```
tdas, line 40: How to use it should not be a under "Building". It should be a top-level section likee
- Intro
- Usage
- Building

flink-connector/delta_sink_wiki.md
```diff
@@ -0,0 +1,275 @@
...
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-client</artifactId>
+            <version>${hadoop-version}</version>
+        </dependency>
+    </dependencies>
+</project>
+```
+
+### SBT
```
tdas, line 106: This is repeated from the README

examples/flink-connector-example-project/pom.xml
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
tdas, line 1: name the directory `flink-example` to reduce length

flink-connector/README.md
```diff
@@ -0,0 +1,237 @@
+# Flink Delta Lake Connector
```
tdas, line 1: why do you need the above empty directories?

examples/flink-connector-example-project/pom.xml
pkubit-g: done

flink-connector/README.md
pkubit-g: I assume you mean `example_table` and `example_partitioned_table` ?
Those are the directories that the example project uses as target paths for its Delta's tables. 
With every run of the example project (before starting the in-memory Flink job) those dirs are being emptied and recreated so we can easily remove them, but i thought that we may as well leave them in the example dir to explicitly indicate for the users where they should look for created tables and written data/delta log.

flink-connector/README.md
```diff
@@ -0,0 +1,237 @@
...
+### Frequently asked questions (FAQ)
+
+#### Can I use this connector to read data from a Delta Lake table?
+
+No, currently we are supporting only writing to a Delta Lake table. `DeltaSource` with the support for reading data from
+Delta's tables will be added in future releases.
+
+#### Can I use this connector to append data to a Delta Lake Table?
+
+Yes, you can use this connector to append data to either an existing or a new Delta Lake Table (if there is no existing
```
scottsand-db, line 206: `Delta Lake table` (lowercase `t`)

flink-connector/README.md
```diff
@@ -0,0 +1,237 @@
...
+#### Can I use this connector with other modes (overwrite, upsert etc.) ?
+
+No, currently only append is supported, other modes may be added in future releases.
+
+#### Do I need to specify the partition columns when creating a Delta table?
+
+If you are using DataStream API then you have to provide a
+`org.apache.flink.streaming.api.functions.sink.filesystem.BucketAssigner` instance while building the sink instace. You
+are free to roll out your own implementation of bucket assigner or use utility one provided by the connector
+as `delta.io.flink.DeltaTablePartitionAssigner`.
```
scottsand-db, line 218: can you include a link to the example showing this?

flink-connector/README.md
```diff
@@ -0,0 +1,237 @@
...
+#### Why do I need to specify the table schema? Shouldn’t it exist in the underlying Delta table metadata or cannot be extracted from the stream's metadata?
+
+Unfortunately we cannot extract schema information from a generic DataStream and it is also required for interacting
+with DeltaLog. The sink must be aware of both Delta table's schema and the structure of the events in the stream in
+order not to violate the integrity of the table.
+
+#### What if I change the underlying Delta table schema ?
+
+Next commit (after mentioned schema change) performed from the DeltaSink to a DeltaLog will fail unless you will
+set `shouldTryUpdateSchema` param to true. In such case DeltaStandaloneWriter will try to merge both schemas and check
```
scottsand-db, line 229: `Delta Standalone`

flink-connector/README.md
```diff
@@ -0,0 +1,237 @@
...
+for their compatibility. If this check will fail (e.g. the change consisted of removing a column) so will the DeltaLog's
+commit which will cause failure of the Flink job.
```
scottsand-db, line 231: then what? how would users recover from this state?

scottsand-db (changes requested):
Minor changes requested

scottsand-db:
@tdas since this README.md is going into the `flink-connector/` folder, then an action item for us should be to refactor the root `README.md` for hive and standalone into similar folders.

Then the root README.md should have more general information, with links to WIKIs / docs / other READMEs / Javadocs, eh?

pkubit-g:
hi @scottsand-db so i was trying to generate the javadocs but it seems like the exclusion rule works in other way than expected. 
My understating was that it should exclude javadocs for the `internal` classes, whereas it seems that it excludes the internal classes and only then tries to compile the code, what fails because included classes try to import those not existing (excluded ones).
Screenshot from 2022-01-14 23-52-06

Please let me know if've misintepreted or overlooked something. 
I'm trying to find another solution for this exclusion rule but with no luck yet since the docs for this `unidoc` are very scarce. Or maybe there are sbt tricks that we could use to workaround this ?

examples/flink-example/src/main/java/io/delta/flink/example/sink/DeltaSinkExample.java
```diff
@@ -0,0 +1,31 @@
...
+import org.apache.flink.table.data.RowData;
+import org.apache.hadoop.conf.Configuration;
+
+import java.nio.file.Paths;
+
+/**
+ * Demonstrates how the Flink Delta Sink can be used to write data to Delta table.
+ *
+ * If you run this example then application will spawn example local Flink job generating data to the underlying
+ * Delta table under directory of "/src/main/resources/example_table". The job will be run in a daemon thread while
```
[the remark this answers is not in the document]
pkubit-g: yep, my bad. Corrected. Thanks!

examples/flink-example/src/main/java/io/delta/flink/example/sink/DeltaSinkExample.java
```diff
@@ -0,0 +1,31 @@
...
+ * in the main app's thread there will Delta Standalone Reader application reading and printing all the data to the
```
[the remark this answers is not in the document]
pkubit-g: corrected

examples/flink-example/src/main/java/io/delta/flink/example/sink/DeltaSinkExampleBase.java
```diff
@@ -0,0 +1,100 @@
...
+        if (tableDir.list().length > 0) {
+            FileUtils.cleanDirectory(tableDir);
+        }
+        StreamExecutionEnvironment env = getFlinkStreamExecutionEnvironment(tablePath);
+        runFlinkJobInBackground(env);
+        printDeltaTableRows(tablePath);
+    }
+
+    DeltaSink<RowData> getDeltaSink(String tablePath) {
+        throw new RuntimeException("Not implemented");
```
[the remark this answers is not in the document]
pkubit-g: my bad, have been switching between different languages lately and i missed the fact that when having abstract class in Java i should not be throwing anything, just make it abstract. Corrected.

examples/flink-example/src/main/java/io/delta/flink/example/sink/DeltaSinkExampleSourceFunction.java
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.flink.streaming.api.functions.source.RichParallelSourceFunction;
+import org.apache.flink.table.data.RowData;
+import org.apache.flink.table.data.util.DataFormatConverters;
+import org.apache.flink.table.types.utils.TypeConversions;
+import org.apache.flink.types.Row;
+
+import java.util.concurrent.ThreadLocalRandom;
+
+/**
+ * Internal class providing mock implementation for example stream source.
```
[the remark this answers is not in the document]
pkubit-g: corrected

examples/flink-example/src/main/java/io/delta/flink/example/sink/DeltaSinkPartitionedTableExample.java
```diff
@@ -0,0 +1,56 @@
...
+import org.apache.hadoop.conf.Configuration;
+
+import java.nio.file.Paths;
+import java.util.LinkedHashMap;
+
+/**
+ * Demonstrates how the Flink Delta Sink can be used to write data to a partitioned Delta table.
+ *
+ * If you run this example then application will spawn example local Flink job generating data to the underlying
+ * Delta table under directory of "/src/main/resources/example_table". The job will be run in a daemon thread while
```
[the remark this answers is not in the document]
pkubit-g: corrected

examples/flink-example/src/main/java/io/delta/flink/example/sink/DeltaSinkPartitionedTableExample.java
```diff
@@ -0,0 +1,56 @@
...
+ * in the main app's thread there will Delta Standalone Reader application reading and printing all the data to the
```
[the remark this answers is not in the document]
pkubit-g: corrected

flink-connector/README.md
```diff
@@ -0,0 +1,237 @@
+# Flink Delta Lake Connector
+
+[![License](https://img.shields.io/badge/license-Apache%202-brightgreen.svg)](https://github.com/delta-io/connectors/blob/master/LICENSE.txt)
+
+Official Delta Lake connector for [Apache Flink](https://flink.apache.org/).
+
+# Introduction
+
+Flink Delta Lake Connector is a JVM library to read and write data from Apache Flink applications' to a Delta Lake
+tables utilizing [Delta Standalone JVM library](https://github.com/delta-io/connectors). It includes
```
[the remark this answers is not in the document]
pkubit-g: corrected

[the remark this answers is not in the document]
pkubit-g: corrected

flink-connector/README.md
```diff
@@ -0,0 +1,237 @@
...
+
+- Sink for writing data from Apache Flink to a Delta tables
```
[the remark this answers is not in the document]
pkubit-g: corrected

flink-connector/README.md
```diff
@@ -0,0 +1,237 @@
...
+- Source for reading Delta Lake's table using Apache Flink (still in progress)
+
+NOTE:
+
+- currently only sink is supported which means that this connectors supports only writing to a Delta table and is not
+  able to read from it,
```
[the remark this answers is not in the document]
pkubit-g: corrected

flink-connector/README.md
pkubit-g: they don't, at least automatically. i believe this is what we have specified and agreed on within the design doc - if the DeltaLog's commit fails because of some sort of schema incompatibility then it cannot be automatically resolved by the connector and we should fail the job to let the user perform a validation.

Are you suggesting that we should put some comment like `such situation may require some manual user's validation in order to retain the table's consistency` ?

pkubit-g:
hi @tdas @scottsand-db 
so i've corrected the linkage in java docs from root API classes to `internal` package. There where only 4 such cases, all of the other errors are caused by removed sources and it seems like the unidoc plugin tries to compile the module aftere excluding the internal package. Funnily enough it's not the case for the `standalone` module. I've had a quick chat on this with @scottsand-db and he has some smart ideas how to workaround this so i will wait for your expertise on this as i;ve failed when trying to resolve it on my own.

flink-connector/src/main/java/io/delta/flink/sink/DeltaSink.java
```diff
@@ -164,11 +167,11 @@ private String restoreOrCreateAppId(List<DeltaWriterBucketState> states) {
      *               appId
      * @return value of the nextCheckpointId to be passed to the new writer's instance
      */
-    private long restoreOrGetNextCheckpointId(List<DeltaWriterBucketState> states) {
+    private long restoreOrGetNextCheckpointId(List<AbstractDeltaWriterBucketState> states) {
```
scottsand-db, line 170: do we need to do this for private methods? those won't be exposed in the javadocs right?

flink-connector/src/main/java/io/delta/flink/sink/internal/writer/DeltaWriter.java
```diff
@@ -290,7 +286,7 @@ public void write(IN element, Context context) throws IOException {
         }
 
         incrementNextCheckpointId();
-        return committables;
+        return (List<AbstractDeltaCommittable>) (List<?>) committables;
```
scottsand-db, line 289: Can we avoid doing this by doing

```java
List<AbstractDeltaCommittable> committables = new ArrayList<>();
...
committables.addAll(...);
...
return committables;
```

flink-connector/src/test/java/io/delta/flink/sink/utils/DeltaSinkTestUtils.java
```diff
@@ -330,8 +344,8 @@ private static LocalRecoverableWriter getTestBucketRecoverableWriter(Path path)
     // IT case utils
     ///////////////////////////////////////////////////////////////////////////
 
-    public static DeltaSink<RowData> createDeltaSink(String deltaTablePath,
-                                                     boolean isTablePartitioned) {
+    public static Sink<RowData, ?, ?, ?> createDeltaSink(String deltaTablePath,
```
scottsand-db, line 347: just curious, why doesn't `DeltaSink<RowData>` work?

flink-connector/src/main/java/io/delta/flink/sink/DeltaSink.java
pkubit-g: you're right. I've simply went too far, thanks !

flink-connector/src/main/java/io/delta/flink/sink/internal/writer/DeltaWriter.java
pkubit-g: yeah, should work. 
Tbh i do not have a preference which is better - casting or creating duplicated list so can change this.
Out of curiosity - is there any specific reason for proposing this besides code style ?

flink-connector/src/test/java/io/delta/flink/sink/utils/DeltaSinkTestUtils.java
pkubit-g: it's a leftover from previous prototyping. 
I haven't done a second pass on all the changes in the PR yet, will do it tomorrow.

pkubit-g:
@scottsand-db 
you're quick :) In between i got stuck with rebasing this branch on top of current master as there were some tricky conflicts and only now i'm pushing rebased version. You can take a look at this once again if you wish, however i do need to take a detailed pass on the whole PR once again tomorrow to make sure that i've resolved all the conflicts correctly.

pkubit-g:
@cheleb could you take a look at failing checks in this PR ? I've rebased it on top of master and now the build is failing with some weird error and it looks like [this commit](https://github.com/delta-io/connectors/commit/416a3d593d254e11d444f9c60ba71cec93d6739f) may have sth to do with it.

flink-connector/src/main/java/io/delta/flink/sink/internal/writer/DeltaWriter.java
pkubit-g: @scottsand-db after giving it a second thought i would say that casting here is better than rewritting whole collection, isn't it?

scottsand-db: Where is the duplicate collection? We create `committables` .... then we add to it .... then we return it.

Let me know what I am missing.

pkubit-g: with this solution we need to allocate new ArrayList instance which will contain the same elements as the initial one, so the casting will come with a slightly smaller cost. 
I'm not saying that this "cost" is an argument here, however i do not see what we would gain if we go for new ArrayList allocation instead of explicit casting.

cheleb:
> @cheleb could you take a look at failing checks in this PR ? I've rebased it on top of master and now the build is failing with some weird error and it looks like [this commit](https://github.com/delta-io/connectors/commit/416a3d593d254e11d444f9c60ba71cec93d6739f) may have sth to do with it.

genjavadoc plugin should be updated ?

flink-connector/src/main/java/io/delta/flink/sink/internal/writer/DeltaWriter.java
kristoffSC: I strongly suggest casting over creating a new array. 

Actually I would say that casting here has almost zero cost.
I believe we do Implicit casting, meaning casting from a type to a wider type, which is done automatically and there is no overhead.

DeltaCommittable -> AbstractDeltaCommittable, plus this is all Generics so in the runtime List<AbstractDeltaCommittable> becomes List<Object> anyways.

scottsand-db: Sorry, I don't think you've answered my root question of "where is this duplicate collection?". How does my approach "Create a new array"?

From what I understand, your approach does
```java
List<DeltaCommittable> committables = new ArrayList<>();
...
while (activeBucketIter.hasNext()) {
    ...
        committables.addAll(entry.getValue().prepareCommit(flush, appId, nextCheckpointId));
    ...
}
...
return (List<AbstractDeltaCommittable>) (List<?>) committables;
```

and my approach does this
```java
List<AbstractDeltaCommittable> committables = new ArrayList<>(); // ****** use Abstract ******
...
while (activeBucketIter.hasNext()) {
    ...
        committables.addAll(entry.getValue().prepareCommit(flush, appId, nextCheckpointId));
    ...
}
...
return committables; // ****** no casting ******
```

I just don't see where the duplicate array is coming from. There's still only one `ArrayList`. Let me know.

scottsand-db: > Out of curiosity - is there any specific reason for proposing this besides code style ?
This is purely a code style suggestion

kristoffSC: Hah,
I think there was a misunderstanding :)

In that code snipped you added to your comment, it looked like you would like to create a new Array, 
where in fact you just changed the generic type in already existing Array.

So in that case, I would be voting to chanign the original, existing array declaration from:
`List<DeltaCommittable> committables = new ArrayList<>();`
to
`List<AbstractDeltaCommittable> committables = new ArrayList<>(); // ****** use Abstract ******`

scottsand-db: Ah, that makes more sense.
So yeah, I prefer `List<AbstractDeltaCommittable> committables = new ArrayList<>(); // ****** use Abstract ******` instead of casting.

pkubit-g: oh my ...
@scottsand-db you're absolutely right and thank you for spotting this as i've missed this detail during refactor.

And yeah, the misunderstanding was that creating a separate array is an alternative solution to casting, but obviously only when the collection is passed as an method's arguments, eg.
```
def someMethod(List<Committable> committables) {

    List<AbstractDeltaCommittable> abstractCommittables = new ArrayList<>();
    abstractCommittables.addAll(committables);
    
    // same effect:
    (List<AbstractDeltaCommittable>) (List<?>) committables
```    
So that's why i thought you are suggesting the former instead of the latter. But obviously both solutions doesn't apply when we are creating the collection inside the method, so thank you for "finger-pointing" the suggestion :) 


FYI i'm moving this part to a separate PR as discussed and will include this fix in there.

pkubit-g:
@cheleb this has done the trick, thank you !

PR #237, opened 2021-12-16
Clean up public APIs

zsxwing:
Fix some minor issues in public APIs

### DIFF
--- a/standalone/src/main/java/io/delta/standalone/actions/Action.java
+++ b/standalone/src/main/java/io/delta/standalone/actions/Action.java
@@ -35,13 +35,4 @@ package io.delta.standalone.actions;
  * }</pre>
  */
 public interface Action {
-
-    /** The maximum reader version of the protocol that this version of Delta Standalone
-     * understands.
-     */
-    int readerVersion = 1;
-    /** The maximum writer version of the protocol that this version of Delta Standalone
-     * understands.
-     */
-    int writerVersion = 2;
 }
--- a/standalone/src/main/java/io/delta/standalone/actions/Protocol.java
+++ b/standalone/src/main/java/io/delta/standalone/actions/Protocol.java
@@ -30,14 +30,9 @@ import java.util.Objects;
  *
  * @see  <a href="https://github.com/delta-io/delta/blob/master/PROTOCOL.md#protocol-evolution">Delta Transaction Log Protocol: Protocol Evolution</a>
  */
-public class Protocol implements Action {
+public final class Protocol implements Action {
     private final int minReaderVersion;
     private final int minWriterVersion;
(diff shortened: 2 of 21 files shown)


standalone/src/main/java/io/delta/standalone/actions/Action.java
```diff
@@ -35,13 +35,4 @@
  * }</pre>
  */
 public interface Action {
-
-    /** The maximum reader version of the protocol that this version of Delta Standalone
```
zsxwing, line 39: Remove them as this should be checked internally by DSR/DSW and we don't need to expose them to the users right now

standalone/src/main/java/io/delta/standalone/actions/Protocol.java
```diff
@@ -30,15 +30,10 @@
  *
  * @see  <a href="https://github.com/delta-io/delta/blob/master/PROTOCOL.md#protocol-evolution">Delta Transaction Log Protocol: Protocol Evolution</a>
  */
-public class Protocol implements Action {
+public final class Protocol implements Action {
     private final int minReaderVersion;
     private final int minWriterVersion;
 
-    public Protocol() {
```
zsxwing, line 37: Remove this as we remove Action.readerVersion/writerVersion.

standalone/src/main/java/io/delta/standalone/actions/RemoveFile.java
```diff
@@ -53,7 +53,8 @@
     @Nullable
     private final Map<String, String> partitionValues;
 
-    private final long size;
+    @Nonnull
+    private final Optional<Long> size;
```
zsxwing, line 57: Changing this to an Optional to reflect the fact that this may be empty.

standalone/src/main/java/io/delta/standalone/util/ParquetSchemaConverter.java
```diff
@@ -61,7 +61,7 @@
      * @return {@code schema} as a Parquet {@link MessageType}
      * @throws IllegalArgumentException if a {@code StructField} name contains invalid character(s)
      */
-    public static MessageType sparkToParquet(StructType schema) {
+    public static MessageType deltaToParquet(StructType schema) {
```
zsxwing, line 64: rename `sparkToParquet` to `deltaToParquet` as `StructType` is DS's own schema.

allisonport-db (approved):
Should all actions be `final`? Missing `CommitInfo` (and `JobInfo` and `NotebookInfo`). Otherwise LGTM

zsxwing:
Good question. Those APIs have been released so we cannot change them.

PR #239, opened 2021-12-16
Flink Delta Sink - PR 7

pkubit-g:
In this PR:
- `flink-connector`'s dependencies are being marked as "provided",
- `flink-connector`'s packages are being renamed `org.apache.flink.connector.delta.sink` -> `io.delta.flink.sink`


tdas (changes requested):
I think we should have separate packages for the public API classes (like Builder and Sink) and the internal classes (committers and committables). This would allows us to only put the public class package in Java docs.... and hide all the internal classes by hiding the entire package from the java docs.

pkubit-g:
@tdas do you have a preference whether we should use exclude or include condition for docs generator plugin ?
If no then i can simply put all of nested packages under `internal` and you will be able to put exlusion for `io.delta.flink.sink.internal` package. Sounds good ? I'm committing this right know for reference so take a look and let me know if you see it differently. Also i'm not sure about `SchemaConverter` class so leaving it as is which means top-level, but if you think that this should go to internal as well then let me know.

tdas:
that is usually how we do it in other places. internal as the package name sounds good. that makes it very obvious independent of the java docs that these classes are not public.

the schema converter is not something that the users have to directly use in any way.... isnt it? if not, then make it internal.

pkubit-g:
@tdas yeah, given a second thought i agree with "hiding" SchemaConverter as well. Done && committed.
Screenshot from 2021-12-17 16-43-26

build.sbt
```diff
@@ -649,7 +647,7 @@ lazy val flinkConnector = (project in file("flink-connector"))
     sourceGenerators in Compile += Def.task {
       val file = (sourceManaged in Compile).value / "meta" / "Meta.java"
       IO.write(file,
-        s"""package org.apache.flink.connector.delta.sink;
+        s"""package io.delta.flink.sink;
```
tdas, line 650: this does not need to be in thee public package.. isnt it?

pkubit-g: well, that's a good catch. Changed.

build.sbt
```diff
@@ -632,13 +632,11 @@ lazy val flinkConnector = (project in file("flink-connector"))
...
     libraryDependencies ++= Seq(
-      "org.apache.flink" % "flink-core" % flinkVersion,
-      "org.apache.flink" % "flink-connector-files" % flinkVersion,
-      "org.apache.flink" % "flink-table-common" % flinkVersion,
-      "org.apache.flink" %% "flink-parquet" % flinkVersion,
-      "org.apache.flink" %% "flink-runtime" % flinkVersion,
-      "org.apache.flink" %% "flink-table-runtime-blink" % flinkVersion,
-      "org.apache.hadoop" % "hadoop-client" % hadoopVersion,
+      "org.apache.flink" %% "flink-runtime" % flinkVersion % "provided",
+      "org.apache.flink" %% "flink-table-runtime-blink" % flinkVersion % "provided",
```
tdas, line 636: do we need all of these for compile time?

pkubit-g: that's my bad. I thought i verified this some time ago but i may have some protyped code in the project that time, sorry for this.

I have narrowed it down and removed `flink-runtime`  and `flink-streaming-java % test`. Also i've changed scope for `flink-table-runtime-blink` from `provided` to `test`.

I will also adjust the docs and readme branch accordingly.

pkubit-g: @tdas tests went good, can you double check if the current state of build.sbt resolves your concerns ?

tdas (approved):

PR #240, opened 2021-12-16
Update build.sbt

allisonport-db:
- Move `releaseProcess` to root project

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -78,6 +78,9 @@ lazy val commonSettings = Seq(
 
 lazy val releaseSettings = Seq(
   publishMavenStyle := true,
+  publishArtifact := true,
+  (publishArtifact in Test) := false,
+  releasePublishArtifactsAction := PgpKeys.publishSigned.value,
   releaseCrossBuild := true,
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -411,7 +413,6 @@ lazy val standalone = (project in file("standalone"))
   .enablePlugins(GenJavadocPlugin, JavaUnidocPlugin)
   .settings(
     name := "delta-standalone-original",
-    skip in publish := true,
```
zsxwing, line 414: This should be kept. `standaloneCosmetic` is the project we want to publish

allisonport-db: Shouldn't it be in `skipReleaseSettings`?

allisonport-db: This can be updated: 
```
lazy val skipReleaseSettings = Seq(
  publishArtifact := false,
  skip in publish := true
)
```

zsxwing: You are right. Didn't know you were removing a duplicate config

zsxwing (approved):

PR #242, opened 2021-12-20
Fail when committing metadata with incorrect protocol update

### LINKED ISSUE #229: OptimisticTransactionImpl: add extra warning when incorrectly updating the table Protocol
Currently, since DSW 0.3.0 only supports a single protocol version (1, 2), when users try to [change the protocol using metadata.configuration in OptimisticTransaction](https://github.com/delta-io/connectors/blob/master/standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala#L181), we ignore those changes.

In addition to ignoring those changes, we should log a warning, too.

allisonport-db:
Resolves #229 

Should we throw an error if protocol configuration properties are provided, but they are correct?

### DIFF
--- a/standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
@@ -64,8 +64,6 @@ private[internal] class OptimisticTransactionImpl(
 
   /** Stores the updated metadata (if any) that will result from this txn. */
   private var newMetadata: Option[Metadata] = None
-  /** Stores the updated Java metadata (if any) that will result from this txn. */
-  private var newMetadataJ: Option[MetadataJ] = None
 
   /** Stores the updated protocol (if any) that will result from this txn. */
   private var newProtocol: Option[Protocol] = None
@@ -174,17 +172,29 @@ private[internal] class OptimisticTransactionImpl(
     scan
   }
 
+  /**
+   * All [[Metadata]] actions must go through this function, and be added to the committed actions
+   * via `newMetadata` (they shouldn't ever be passed into `prepareCommit`.)
+   * This function enforces:
+   * - At most one unique [[Metadata]] is committed in a single transaction.
+   * - If this is the first commit, the committed metadata configuration includes global Delta
+   *   configuration defaults.
+   * - Checks for unenforceable NOT NULL constraints in the table schema.
+   *  - Checks for column name duplication.
+   *    - Verifies column names are parquet compatible.
+   *    - Enforces that protocol versions are not part of the table properties.
+   */
   override def updateMetadata(metadataJ: MetadataJ): Unit = {
 
+    var latestMetadata = ConversionUtils.convertMetadataJ(metadataJ)
+
     // this Metadata instance was previously added
-    if (newMetadataJ.exists(_ eq metadataJ)) {
+    if (newMetadata.contains(latestMetadata)) {
       return
     }
 
     assert(newMetadata.isEmpty,
       "Cannot change the metadata more than once in a transaction.")
-
-    var latestMetadata = ConversionUtils.convertMetadataJ(metadataJ)
 
     if (readVersion == -1 || isCreatingNewTable) {
       latestMetadata = withGlobalConfigDefaults(latestMetadata)
@@ -200,7 +210,6 @@ private[internal] class OptimisticTransactionImpl(
     logInfo(s"Updated metadata from ${newMetadata.getOrElse("-")} to $latestMetadata")
 
     newMetadata = Some(latestMetadata)
-    newMetadataJ = Some(metadataJ)
   }
 
   override def readWholeTable(): Unit = {
--- a/standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
@@ -64,6 +64,7 @@ private[internal] class OptimisticTransactionImpl(
 
   /** Stores the updated metadata (if any) that will result from this txn. */
   private var newMetadata: Option[Metadata] = None
+  private var newMetadataJ: Option[MetadataJ] = None // TODO: document
 
   /** Stores the updated protocol (if any) that will result from this txn. */
   private var newProtocol: Option[Protocol] = None
@@ -99,7 +100,15 @@ private[internal] class OptimisticTransactionImpl(
       actionsJ: java.lang.Iterable[T],
       op: Operation,
       engineInfo: String): CommitResult = {
-    val actions = actionsJ.asScala.map(ConversionUtils.convertActionJ).toSeq
+
+    actionsJ.asScala.collect { case m: MetadataJ => m }.foreach { m =>
+      updateMetadata(m)
+    }
+
+    val actions = actionsJ.asScala
+      .map(ConversionUtils.convertActionJ)
+      .filter(!_.isInstanceOf[Metadata])
+      .toSeq
 
     // Try to commit at the next version.
     var preparedActions = prepareCommit(actions)
@@ -165,6 +174,12 @@ private[internal] class OptimisticTransactionImpl(
   }
 
   override def updateMetadata(metadataJ: MetadataJ): Unit = {
+
+    // this Metadata instance was previously added
+    if (newMetadataJ.exists(_ eq metadataJ)) {
+      return
+    }
+
     assert(newMetadata.isEmpty,
       "Cannot change the metadata more than once in a transaction.")
 
@@ -185,6 +200,7 @@ private[internal] class OptimisticTransactionImpl(
     logInfo(s"Updated metadata from ${newMetadata.getOrElse("-")} to $latestMetadata")
 
     newMetadata = Some(latestMetadata)
+    newMetadataJ = Some(metadataJ)
   }
 
   override def readWholeTable(): Unit = {
@@ -203,7 +219,6 @@ private[internal] class OptimisticTransactionImpl(
 
   /**
    * Prepare for a commit by doing all necessary pre-commit checks and modifications to the actions.
-   *
    * @return The finalized set of actions.
    */
   private def prepareCommit(actions: Seq[Action]): Seq[Action] = {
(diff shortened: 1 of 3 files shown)


scottsand-db:
> Should we throw an error if protocol configuration properties are provided, but they are correct?

Good question. We shouldn't throw an error in this case.

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
```diff
@@ -263,6 +265,8 @@ private[internal] class OptimisticTransactionImpl(
     if (protocolOpt.isDefined) {
       assert(protocolOpt.get == Protocol(), s"Invalid Protocol ${protocolOpt.get.simpleString}. " +
         s"Currently only Protocol readerVersion 1 and writerVersion 2 is supported.")
+    } else {
+      finalActions = newProtocol.toSeq ++ finalActions
```
scottsand-db, line 269: doesn't this mean we are double-adding the Protocol action?

```scala
val protocolOpt = finalActions.collectFirst{ case p: Protocol => p } // extract Protocol from finalActions list
...
finalActions = newProtocol.toSeq ++ finalActions // add Protocol back into the finalActions list
```

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
```diff
@@ -231,19 +240,12 @@ private[internal] class OptimisticTransactionImpl(
...
-    // If the metadata has changed, add that to the set of actions
-    finalActions = newMetadata.toSeq ++ finalActions
-
-    val metadataChanges = finalActions.collect { case m: Metadata => m }
-    assert(metadataChanges.length <= 1,
-      "Cannot change the metadata more than once in a transaction.")
-
-    metadataChanges.foreach { m =>
-      verifyNewMetadata(m)
+    newMetadata.foreach{ m =>
```
scottsand-db, line 243: nit: ` {` (spacing)

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
```diff
@@ -231,19 +240,12 @@ private[internal] class OptimisticTransactionImpl(
       case a: Action => a
     }
 
-    // If the metadata has changed, add that to the set of actions
-    finalActions = newMetadata.toSeq ++ finalActions
-
-    val metadataChanges = finalActions.collect { case m: Metadata => m }
-    assert(metadataChanges.length <= 1,
-      "Cannot change the metadata more than once in a transaction.")
```
scottsand-db, line 239: since we are removing this assertion, can you please update the scaladoc of this method stating that the caller must ensure no Metadata action is being passed inside of `actions` ?

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
```diff
@@ -165,6 +175,12 @@ private[internal] class OptimisticTransactionImpl(
   }
 
   override def updateMetadata(metadataJ: MetadataJ): Unit = {
+
+    // this Metadata instance was previously added
+    if (newMetadataJ.exists(_ eq metadataJ)) {
+      return
```
scottsand-db, line 181: so if the user passes in two identical Metadata, we will take the first one only and not throw an exception.

can you add a test for this? (this is a weird case, but a test case makes it explicit)

For example, at first pass I was confused why the test case copied here was passing, but it is because each Metadata() has a different created at time.

```
test("commits shouldn't have more than one Metadata") {
    withTempDir { dir =>
      val log = DeltaLog.forTable(new Configuration(), dir.getCanonicalPath)
      val e = intercept[AssertionError] {
        txn.commit(Metadata() :: Metadata() :: Nil, manualUpdate, engineInfo)
      }
      assert(e.getMessage.contains("Cannot change the metadata more than once in a transaction."))
    }
  }
```

Can you also please update this test case name since it is no longer accurate

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
```diff
@@ -180,19 +196,12 @@ private[internal] class OptimisticTransactionImpl(
       SchemaUtils.checkUnenforceableNotNullConstraints(latestMetadata.schema)
     }
 
-    // Remove the protocol version properties
-    val noProtocolVersionConfig = latestMetadata.configuration.filter {
-      case (Protocol.MIN_READER_VERSION_PROP, _) => false
```
scottsand-db, line 185: can you remind me, what are the consequences of not removing this configuration?

we will still fail later, right? so it won't be committed?

I'm worried that removing this code can mean that we forget to add it back later to Delta Standalone, when we do support multiple protocol versions.

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
```diff
@@ -64,6 +64,8 @@ private[internal] class OptimisticTransactionImpl(
 
   /** Stores the updated metadata (if any) that will result from this txn. */
   private var newMetadata: Option[Metadata] = None
+  /** Stores the updated Java metadata (if any) that will result from this txn. */
+  private var newMetadataJ: Option[MetadataJ] = None
```
scottsand-db, line 68: This seems very coupled to `newMetadata`. i.e. if one is None, the other should be, too.

also, i think we can get rid of it? inside of `updateMetadata` we already convert MetadataJ to Metadata, right? `var latestMetadata = ConversionUtils.convertMetadataJ(metadataJ)`

Couldn't we just perform the Metadata checks then? Let me know if I'm missing something.

scottsand-db (changes requested):
Left a couple of comments + questions

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
allisonport-db: Err `newProtocol` is only defined if it's a new table (see `updateMetadata`). But this is actually repetitive with 250-262. I think it makes sense to remove the protocol creation in `updateMetadata` and not use `newProtocol` at all.

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
allisonport-db: This assertion is enforced by multiple calls to `updateMetadata` (it will fail the second time it's called.) Re: #229  the issue conversation we decided to accept `Metadata`s in `commit`

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
allisonport-db: We won't fail later. The check is in `verifyNewMetadata`, if we remove these it won't fail. I will update that method to accept the correct protocol configurations, and then that would be where later updates would go for future version support.

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
allisonport-db: So right now it's only accepting the **same** metadata instance (it's checking for reference equality, which is why we're saving the java metadata). Do you think it's more correct to accept equivalent Metadata actions?

For context, this was in theory to accommodate for something like
```
Metadata newMetadata = .....
txn.updateMetadata(newMetadata)
List<Action> actions = .... // generate actions including newMetadata
txn.commit(actions, ....)
```

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
allisonport-db: See below (for reference vs object equality)

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
scottsand-db: Yup, the assertion enforcement is fine, but I'm wondering if it is clear to future developers how it is that at most one metadata can be committed.

Perhaps add a comment to `updateMetadata` that all metadatas (either directly or extracted from commit actions) must go through that method?

But if you think it's fine, then SGTM.

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
scottsand-db: I see. I guess my question is, why check for reference equality instead value equality? If two different Metadata actions are value-equivalent, then the same JSON will be written to the delta log, which is all that matters.

So I think that value-equality is fine here.

And this will still work for your code example given above

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
scottsand-db: > that would be where later updates would go for future version support

But why shift this check from here to that function? If Delta OSS and DBR both perform similar checks in this method, what do we gain from moving checks there?

scottsand-db (changes requested):
Responded to comments + asked some questions

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
allisonport-db: This isn't a check though, it's filtering out those properties. The check has always been in `verifyNewMetadata` and our issue with not failing was because we were filtering out those properties before. In delta OSS all metadata goes through `verifyNewMetadata` in both `prepareCommit` and `updateMetadata`

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
allisonport-db: I think that's fair then sg

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
allisonport-db: I'll add a comment

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
```diff
@@ -164,30 +172,39 @@ private[internal] class OptimisticTransactionImpl(
...
   }
 
+  /**
+   * All [[Metadata]] actions must go through this function, and be added to the committed actions
+   * via `newMetadata` (they shouldn't ever be passed into `prepareCommit`.)
+   * This function enforces:
+   * - At most one unique [[Metadata]] is committed in a single transaction.
+   * - If this is the first commit, the committed metadata configuration includes global Delta
+   *   configuration defaults.
+   * - Checks for unenforceable NOT NULL constraints in the table schema.
```
scottsand-db, line 182: Do you mean to have 3 levels of indentation?
```
* - ...
*  - ...
*    - ...
```

scottsand-db (approved):
LGTM with 1 indentation nit / comment

PR #243, opened 2021-12-20
DeltaLog::getChanges javadoc clarification

scottsand-db:
The prior `DeltaLog::getChanges` API didn't clearly indicate that the versions returned were in increasing order. This PR fixes that.


allisonport-db (approved):
LGTM

PR #244, opened 2021-12-20
Flink Delta Sink - PR 8

pkubit-g:
Changes included in this PR:
- correct `build.sbt` to mark `delta-standalone` dependency as "provided",
- rename utility method for creating the sink for `RowData` and remove utlity method with `bucketAssigner` param (defining `BucketAssigner` is now possible only via `DeltaSinkBuilder`),
- minor style and code corrections (e.g. rephrasing error message or remove @Internal annotation which is no longer needed after moving some classes under `io.delta.flink.sink.internal` package,
- add IT case for testing if data written to a Delta table with `flink-connector` can be read back by `delta-standalone`,
- rename partition values and specs used in tests (Map("a"->"b") => Map("col1"->"val1")


flink-connector/src/test/java/io/delta/flink/sink/DeltaSinkWriteReadITCase.java
```diff
@@ -0,0 +1,269 @@
...
+            row.getShort("f7");
+            row.getLong("f8");
+            row.getBinary("f9");
+            row.getBinary("f10");
+            row.getTimestamp("f11");
+            row.getTimestamp("f12");
+            row.getDate("f13");
+            row.getString("f14");
+            row.getBigDecimal("f15");
+            row.getBigDecimal("f16");
```
tdas, line 265: you are not testing the values of each col

tdas:
please add in the description what all changes you are making.

pkubit-g:
@tdas FYI i've provided PR's description and added the validation for the values read back in the `DeltaSinkWriteReadITCase`

tdas (approved):

PR #252, opened 2022-01-11
:arrow_up: Scala 2.13.8

cheleb:

PR #253, opened 2022-01-12
:arrow_up: sbt 1.6.1 (was 0.13.x)

cheleb:


cheleb:
@zsxwing a test is failing but I don't think it is related to this sbt upgrade.

build/sbt-launch-lib.bash
```diff
@@ -19,57 +19,65 @@ declare -a sbt_commands
 declare -a maven_profiles
 
 if test -x "$JAVA_HOME/bin/java"; then
-    echo -e "Using $JAVA_HOME as default JAVA_HOME."
-    echo "Note, this will be overridden by -java-home if it is set."
-    declare java_cmd="$JAVA_HOME/bin/java"
+  echo -e "Using $JAVA_HOME as default JAVA_HOME."
```
zsxwing, line 22: Could you clarify how you got a new version of this file?

build.sbt
```diff
@@ -25,9 +25,14 @@ concurrentRestrictions in Global := {
   Tags.limitAll(1) :: Nil
 }
 
-parallelExecution in ThisBuild := false
-scalastyleConfig in ThisBuild := baseDirectory.value / "scalastyle-config.xml"
-crossScalaVersions in ThisBuild := Seq("2.12.8", "2.11.12")
+inThisBuild(
+  Seq(
+    parallelExecution := false,
+    //scalastyleConfig := baseDirectory.value / "scalastyle-config.xml",
```
zsxwing, line 31: Did you hit any issue with this?

build/sbt-launch-lib.bash
cheleb: To be honest I've c&p from the PR https://github.com/delta-io/connectors/pull/252#issuecomment-1011291266 you linked me.

build.sbt
cheleb: sbt warn that this property is not used.

build/sbt-launch-lib.bash
zsxwing: Did you re-format the file? They are not the same when I compared them. If so, could you just copy the file without any change?

build.sbt
zsxwing: I see. I can still see it picks up the correct `scalastyle-config.xml` without this line in SBT output. Could you just remove it since it's not needed?

zsxwing:
For the test failure:
```
[info] - data reader can read partition values *** FAILED ***
[info]   java.io.FileNotFoundException: File file:/home/runner/work/connectors/connectors/standalone/target/scala-2.12/test-classes/golden/data-reader-partition-values/as_int=0/as_long=0/as_byte=0/as_short=0/as_boolean=true/as_float=0.0/as_double=0.0/as_string=0/as_string_lit_null=null/as_date=2021-09-08/as_timestamp=2021-09-08 11%3A11%3A11/as_big_decimal=0/part-00000-b9dc86ae-0134-4363-bd87-19cfb3403e9a.c000.snappy.parquet does not exist
[info]   at org.apache.hadoop.fs.RawLocalFileSystem.deprecatedGetFileStatus(RawLocalFileSystem.java:641)
[info]   at org.apache.hadoop.fs.RawLocalFileSystem.getFileLinkStatusInternal(RawLocalFileSystem.java:930)
[info]   at org.apache.hadoop.fs.RawLocalFileSystem.getFileStatus(RawLocalFileSystem.java:631)
[info]   at org.apache.hadoop.fs.FilterFileSystem.getFileStatus(FilterFileSystem.java:454)
[info]   at org.apache.parquet.hadoop.ParquetReader$Builder.build(ParquetReader.java:300)
[info]   at com.github.mjakubowski84.parquet4s.ParquetIterableImpl$$anon$2.<init>(ParquetReader.scala:103)
[info]   at com.github.mjakubowski84.parquet4s.ParquetIterableImpl.iterator(ParquetReader.scala:100)
[info]   at io.delta.standalone.internal.data.CloseableParquetDataIterator.<init>(CloseableParquetDataIterator.scala:69)
[info]   at io.delta.standalone.internal.SnapshotImpl.open(SnapshotImpl.scala:87)
[info]   at io.delta.standalone.internal.DeltaDataReaderSuite.$anonfun$new$36(DeltaDataReaderSuite.scala:311)
[info]   ...
```
I feel there is some bug in SBT when it copies this file from `golden-tables/src/test/resources/golden/data-reader-partition-values` to `standalone/target/scala-2.12/test-classes/golden/data-reader-partition-values`. Could you disable this test to unblock the PR? We can continue to debug this after merging this PR.

build/sbt-launch-lib.bash
```diff
@@ -37,17 +37,25 @@ dlog () {
 }
 
 acquire_sbt_jar () {
-  SBT_VERSION=`awk -F "=" '/sbt\.version/ {print $2}' ./project/build.properties`
-  URL1=https://repo.typesafe.com/typesafe/ivy-releases/org.scala-sbt/sbt-launch/${SBT_VERSION}/sbt-launch.jar
-  JAR=build/sbt-launch-${SBT_VERSION}.jar
+  SBT_VERSION=$(awk -F "=" '/sbt\.version/ {print $2}' ./project/build.properties)
```
zsxwing, line 40: nit: could you revert it to make it consistent with the file in delta-io/delta?
```suggestion
  SBT_VERSION=`awk -F "=" '/sbt\.version/ {print $2}' ./project/build.properties`
```

cheleb:
Here we are.

zsxwing (approved):
LGTM

PR #254, opened 2022-01-12
Flink Delta Sink - Logging 

pkubit-g:


flink-connector/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
```diff
@@ -252,6 +256,11 @@ private void doCommit(OptimisticTransaction transaction,
         long numOutputRows = 0;
         long numOutputBytes = 0;
         for (DeltaCommittable deltaCommittable : committables) {
+            LOG.info("Committing file to the Delta table: " +
+                "appId=" + deltaCommittable.getAppId() +
+                " checkpointId=" + deltaCommittable.getCheckpointId() +
+                " deltaPendingFile=" + deltaCommittable.getDeltaPendingFile()
+            );
```
tdas, line 263: since all of this going to printed on a single jvm instance, i wonder whether this is going to be too verbose at large scale. i guess this is okay for now, we can make this debug level later

tdas: this can be a bit optimized though.. rather than printing the appId and checkpointId 100 times for 100 files... why not print 
```
appId=X chechkpointId=Y files: 
  file1
  file2
```

flink-connector/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
```diff
@@ -252,6 +256,11 @@ private void doCommit(OptimisticTransaction transaction,
...
             DeltaPendingFile deltaPendingFile = deltaCommittable.getDeltaPendingFile();
             AddFile action = deltaPendingFile.toAddFile();
             addFileActions.add(action);
```
tdas, line 266: there is no logging on successful commit? that is absolutely a critical debug info isnt it? whether commit was ignored / attempted / failed / succeeded should be there.

flink-connector/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
```diff
@@ -78,6 +80,8 @@
 public class DeltaGlobalCommitter
     implements GlobalCommitter<DeltaCommittable, DeltaGlobalCommittable> {
 
+    private static final Logger LOG = LoggerFactory.getLogger(DeltaGlobalCommitter.class);
+
     private static final String APPEND_MODE = "Append";
     private static final String ENGINE_INFO = "flink-delta-connector/" + Meta.VERSION;
```
tdas, line 86: I realized that we are not putting enough metadata here. We want to record both our connector version as well as the flink engine version. can you make that happen in a separate PR?

flink-connector/src/main/java/io/delta/flink/sink/DeltaSink.java
```diff
@@ -87,6 +89,8 @@
 public class DeltaSink<IN>
     implements Sink<IN, DeltaCommittable, DeltaWriterBucketState, DeltaGlobalCommittable> {
 
+    private static final Logger LOG = LoggerFactory.getLogger(DeltaSink.class);
```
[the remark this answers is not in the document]
pkubit-g: do you have in mind a non-static default method ?
Like this:
```
package io.delta.flink.sink.internal.logging;

import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

public interface Logging {

    default Logger getLogger() {
        return LoggerFactory.getLogger(this.getClass());
    }

}
```

because if we simply move above static `LOG` variable to an interface  then instead of:
```
22/01/15 17:15:49.822 main INFO DeltaCommitter: Committing delta committable locally: appId=1 checkpointId=1 deltaPendingFile=org.apache.flink.streaming.api.functions.sink.filesystem.DeltaPendingFile@32a068d1
```
we get
```
22/01/15 17:08:36.516 main INFO Logging: Committing delta committable locally: appId=1 checkpointId=1 deltaPendingFile=org.apache.flink.streaming.api.functions.sink.filesystem.DeltaPendingFile@32a068d1
``` 
which is less informative and not the goal here, right ?

flink-connector/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
pkubit-g: added

flink-connector/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
pkubit-g: [here](https://github.com/delta-io/connectors/pull/260)

flink-connector/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
pkubit-g: changed

flink-connector/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
```diff
@@ -251,6 +256,16 @@ private void doCommit(OptimisticTransaction transaction,
         Set<String> partitionColumnsSet = null;
         long numOutputRows = 0;
         long numOutputBytes = 0;
+
+        StringBuilder logFiles = new StringBuilder();
+        for (DeltaCommittable deltaCommittable : committables) {
+            logFiles.append("\ndeltaPendingFile=").append(deltaCommittable.getDeltaPendingFile());
```
tdas, line 262: unnecessarily verbose to print `deltaPendingFile=` in every line.. isnt it? please think this through. we want just enough verbosity that it is readable and interpretable, but anything more that than is clutter.

flink-connector/src/main/java/io/delta/flink/sink/internal/logging/Logging.java
```diff
@@ -0,0 +1,12 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public interface Logging {
+
+    default Logger getLogger() {
+        return LoggerFactory.getLogger(this.getClass());
+    }
+
```
tdas, line 11: nit: extra linee.

flink-connector/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
```diff
@@ -294,7 +309,13 @@ private void doCommit(OptimisticTransaction transaction,
             operationMetrics
         );
 
+        getLogger().info(String.format(
```
tdas, line 312: why not make `getLogging.info()` and trait method `logInfo()`?? significantly less verbose.

flink-connector/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
pkubit-g: changed `new line` to a whitespace

flink-connector/src/main/java/io/delta/flink/sink/internal/logging/Logging.java
```diff
@@ -0,0 +1,12 @@
+package io.delta.flink.sink.internal.logging;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public interface Logging {
```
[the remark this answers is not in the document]
pkubit-g: added

flink-connector/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
pkubit-g: done

tdas (approved):

PR #255, opened 2022-01-12
Remove "Reader" from Delta Standalone javadoc window title

scottsand-db:
Title is pretty self-explanatory.


allisonport-db (approved):
lgtm

PR #260, opened 2022-01-15
Flink Delta Sink - Update Commit Metadata 

pkubit-g:
Extends commit metadata in transaction log with Flink engine's version

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -653,7 +653,8 @@ lazy val flinkConnector = (project in file("flink-connector"))
         s"""package io.delta.flink.sink.internal;
            |
            |public final class Meta {
-           |  public static final String VERSION = "${version.value}";
+           |  public static final String FLINK_VERSION = "${flinkVersion}";
+           |  public static final String CONNECTOR_VERSION = "${version.value}";
            |}
            |""".stripMargin)
       Seq(file)
--- a/flink-connector/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
+++ b/flink-connector/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
@@ -79,7 +79,8 @@ public class DeltaGlobalCommitter
     implements GlobalCommitter<DeltaCommittable, DeltaGlobalCommittable> {
 
     private static final String APPEND_MODE = "Append";
-    private static final String ENGINE_INFO = "flink-delta-connector/" + Meta.VERSION;
+    private static final String ENGINE_INFO = "flink-engine/" + Meta.FLINK_VERSION +
+        " flink-delta-connector/" + Meta.CONNECTOR_VERSION;
 
     /**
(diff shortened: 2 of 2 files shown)


scottsand-db (approved):

tdas:
No unit test for this? Clearly no test changes here.

scottsand-db:
@tdas - there are unit tests in Delta Standalone that test that whatever the higher-level connector has set as their `engineInfo` is correctly set in the `_delta_log` Metadata.

Should connectors also test this? If we have specified in our Delta Standalone API what the `engineInfo` field is used for, and how it will be persisted, I'm not sure if more testing is necessary.

pkubit-g:
@tdas not sure how i should be testing this. The change is about adding Flink's versions to the engine info and we cannot specify explicitly in tests what value it should expect.

tdas:
no you can check in the test whether the non-version string parts are set.

scottsand-db:
@pkubit-g so it seems you can test it like this then:
- set the engine info and write some data to your flink table
- use Delta Standalone APIs directly to read the metadata and check if the `CommitInfo.engineInfo` field contains what you expect it to. (I don't think there's a way to do this without using low-level Delta Standalone APIs, right?)
- e.g. like this test [here](https://github.com/delta-io/connectors/blob/master/standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionLegacySuite.scala#L536)

pkubit-g:
@tdas @scottsand-db 
i've added a validation for expected string and some regex pattern to represent a version.
Also, i've found out that since i've added this `FLINK_VERSION` to the engine info and separated it with a whitespace from `CONNECTOR_VERSION` then the `standalone` code replaces this whitespace with a `-` sign, but i guess that this is expected ?

So the current full `engine info` string for the Flink Delta Sink is for instance:
`flink-engine/1.14.0-flink-delta-connector/0.2.1 Delta-Standalone/0.2.1-SNAPSHOT`.
And in the test i'm validating only it contains the sink part, that is `flink-engine/1.14.0-flink-delta-connector/0.2.1`

scottsand-db:
@pkubit-g yeah .... perhaps we shouldn't be replacing ` ` with `-` in Delta Standalone ....

For now, it is the way it is, so this test case is fine. If we change it in the future ... then this test case will fail and we will know to fix this test, too.

tdas (approved):

tdas:
@scottsand-db why do we replace whitespace with "-"?

scottsand-db:
The desired format is ~ `this-is-my-engine-name/a.b.c delta-standalone/0.3.0`.
We want spaces separating the different `engine1/version engine2/version`.
We replaced a space with a dash in case someone did `this is my engine name/a.b.c`
It seems we didn't think of the case of `flink-connector/a.b.c flink-version/x.y.z`.

I think it is fair to remove this logic (i.e. do not to that replacement).

PR #250, opened 2022-01-11
Flink Delta Sink - Table API 

pkubit-g:
Adds Table API support for the sink in `flink-connector`

PR #263, opened 2022-01-23
Flink Delta Sink - Rename shouldTryUpdateSchema -> mergeSchema

pkubit-g:
Changes contained in this PR:

- rename 'shouldTryUpdateSchema' -> 'mergeSchema'

### DIFF
--- a/flink-connector/src/main/java/io/delta/flink/sink/DeltaSink.java
+++ b/flink-connector/src/main/java/io/delta/flink/sink/DeltaSink.java
@@ -250,7 +250,7 @@ public class DeltaSink<IN>
             new BasePathBucketAssigner<>(),
             OnCheckpointRollingPolicy.build(),
             rowType,
-            false // shouldTryUpdateSchema
+            false // mergeSchema
         );
     }
 }
--- a/flink-connector/src/main/java/io/delta/flink/sink/DeltaSinkBuilder.java
+++ b/flink-connector/src/main/java/io/delta/flink/sink/DeltaSinkBuilder.java
@@ -92,7 +92,7 @@ public class DeltaSinkBuilder<IN> implements Serializable {
      * those will not match. The update is not guaranteed as there will be still some checks
      * performed whether the updates to the schema are compatible.
      */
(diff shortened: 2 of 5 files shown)


flink-connector/src/main/java/io/delta/flink/sink/DeltaTablePartitionAssigner.java
```diff
@@ -82,11 +87,85 @@ public String toString() {
...
+
+    /**
+     * Implementation of {@link DeltaPartitionComputer} for stream which elements are instances of
+     * {@link RowData}.
+     * <p>
+     * This partition computer resolves partition values by extracting them from element's fields
+     * by provided partitions' names. This behaviour can be overridden by providing static values
+     * for partitions' fields.
+     */
+    public static class DeltaRowDataPartitionComputer implements DeltaPartitionComputer<RowData> {
```
tdas, line 106: why is this class needed for mergeSchema name change??

pkubit-g: it's not. 
I've put the explanation on [slack](https://delta-users.slack.com/archives/C023YA69W6B/p1642980911011400), but maybe i should have put it in the PR's desc ? If yes then sorry for missing that.

Anyway, this is the change coming from the [Table API PR](https://github.com/delta-io/connectors/pull/250) but it may be helpful for DataStream API job's as well so i thought it would be good to merge it earlier and add it here.
But if in this PR you would want to focus only on renaming `mergeSchema` option then let me know and i will decouple those changes into separate PRs.

tdas: yeah. lets focus on renaming only. since the first release wont have any table api and sql support.

pkubit-g: @tdas decoupled

scottsand-db (approved):
LGTM

PR #264, opened 2022-01-25
Fix MiMa issue

allisonport-db:
MiMa compares the compiled class files with the released JAR. `build/sbt standalone/compile` doesn't include the shaded JARs so we ignore them. They will be included with the released `standaloneCosmetic` JAR.

[question] since the release was pushed to `branch-0.3`, should I create a PR to merge that branch into master? And then merge this? or some other combination of that (merge this into `branch-0.3` first?)
- to upgrade the version in master to `0.3.1-SNAPSHOT`


version.sbt
```diff
@@ -1 +1 @@
-ThisBuild / version := "0.2.1-SNAPSHOT"
+version in ThisBuild := "0.3.1-SNAPSHOT"
```
zsxwing, line 1: nit: let's use `0.4.0-SNAPSHOT`

zsxwing:
> [question] since the release was pushed to branch-0.3, should I create a PR to merge that branch into master? And then merge this? or some other combination of that (merge this into branch-0.3 first?)

No. You can just create a new PR for branch-0.3 to fix it separately (the version in branch-0.3 should be `0.3.1-SNAPSHOT`)

scottsand-db:
Yup, so master should have `0.4.0-SNAPSHOT` and branch-0.3 should have `0.3.1-SNAPSHOT`. LGTM.

scottsand-db (approved):
After Ryan's suggestions, LGTM.

zsxwing (approved):
LGTM

PR #265, opened 2022-01-25
Fix MiMa issue in branch-0.3

allisonport-db:
MiMa compares the compiled class files with the released JAR. `build/sbt standalone/compile` doesn't include the shaded JARs so we ignore them. They will be included with the released `standaloneCosmetic` JAR.


zsxwing (approved):
LGTM

PR #266, opened 2022-01-26
Access tables in golden-tables/ directly

zsxwing:
`data reader can read partition values` currently fails because SBT doesn't copy all of files in the table to the correct place. I tried with SBT with various changes and noticed that it won't copy files whose depth is high.

To workaround with the issue, I changed the code to read the tables in `golden-tables/` rather than asking SBT to copy files to `standalone/target/scala-2.12/test-classes/` and use files in this target folder.

As some of tests may need to modify files, I also add utility methods `withLogForWritableGoldenTable`, `withLogImplForWritableGoldenTable`, and `withWritableHiveGoldenTable` for these tests so that we can avoid modifying files in `golden-tables/` when running these tests.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -213,7 +213,6 @@ lazy val hiveMR = (project in file("hive-mr")) dependsOn(hiveTest % "test->test"
   name := "hive-mr",
   commonSettings,
   skipReleaseSettings,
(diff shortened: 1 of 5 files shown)


scottsand-db (approved):
LGTM

PR #267, opened 2022-01-26
Fix crossScalaVersions for Scala 2.11

zsxwing:
`crossScalaVersions` is set in a wrong place after upgrading SBT. For example, `build/sbt "++ 2.11.12 standalone/test"` still uses Scala 2.12.

This PR fixes it.


scottsand-db (approved):
LGTM

PR #269, opened 2022-01-27
Flink Delta Sink - RowData partition computer

pkubit-g:
Changes contained in this PR:

- add partition computer for Flink's RowData type

### DIFF
--- a/flink-connector/src/main/java/io/delta/flink/sink/DeltaTablePartitionAssigner.java
+++ b/flink-connector/src/main/java/io/delta/flink/sink/DeltaTablePartitionAssigner.java
@@ -20,10 +20,15 @@ package io.delta.flink.sink;
 
 import java.io.Serializable;
 import java.util.LinkedHashMap;
+import java.util.List;
 
 import org.apache.flink.core.io.SimpleVersionedSerializer;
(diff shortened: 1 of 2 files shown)


flink-connector/src/main/java/io/delta/flink/sink/DeltaTablePartitionAssigner.java
```diff
@@ -82,11 +87,85 @@ public String toString() {
...
+
+    /**
+     * Implementation of {@link DeltaPartitionComputer} for stream which elements are instances of
+     * {@link RowData}.
+     * <p>
+     * This partition computer resolves partition values by extracting them from element's fields
+     * by provided partitions' names. This behaviour can be overridden by providing static values
+     * for partitions' fields.
+     */
+    public static class DeltaRowDataPartitionComputer implements DeltaPartitionComputer<RowData> {
```
scottsand-db, line 106: All the code LGTM .. but how is the user supposed to use this (it's a public, non-internal class)? Shouldn't our documentation also be updated to include an example of this.

pkubit-g: Of course it should. However the docs and example project are not merged yet so i will update/rebase PRs accordingly to whichever PR will go first to master branch.

PR #270, opened 2022-01-31
:arrow_up: Add Scala 2.13 test runs

cheleb:


allisonport-db:
Thanks for adding this! I realized we should wait to merge this until right before the next release (when there will be a 2.13 version to test.)

allisonport-db (approved):
LGTM

PR #271, opened 2022-02-02
Existing Power BI function wrapped into a custom connector.

pichlerpa:
This PR contains a read-to-use Power BI Custom Connector incorporating the existing function developed with the Power Query SDK in Visual Studio.


.gitignore
```diff
@@ -102,3 +102,6 @@ spark-warehouse/
 
 # For SBT
 .jvmopts
+
+# For VS
+/.vs
```
gbrueckl, line 107: please also add the obj and debug folders from VS to the ignore list

pichlerpa:
@gbrueckl done, new version available.

dennyglee:
Thanks very much for the update!

PR #274, opened 2022-02-04
Document tables created by other systems cannot be read by the Hive connector directly

zsxwing:


README.md
```diff
@@ -128,6 +128,9 @@ No. The connector **must** be used with Apache Hive. It doesn't work in other sy
 #### If I create a table using the connector in Hive, can I query it in Apache Spark or Presto?
 No. The table created by this connector in Hive cannot be read in any other systems right now. We recommend to create different tables in different systems but point to the same path. Although you need to use different table names to query the same Delta table, the underlying data will be shared by all of systems.
 
+#### If a table in the Hive Metastore is created by other systems such as Apache Spark or Presto, can I use this connector to query it in Hive?
+No. If a table in the Hive Metastore is created by other systems such as Apache Spark or Presto, Hive cannot find the correct connector to read it. You can follow our instruction to [create a new table](#create-a-hive-table) with a different table name but point to the same path in Hive. Although it's a different table name, the underlying data will be shared by all of systems. We recommend to create different tables in different systems but point to the same path.
```
scottsand-db, line 132: > We recommend to create different tables in different systems but point to the same path.

To be clear, we do not recommend to create a table using the Hive Connector? Or do you mean, "if a table was created by another system, **then** what we recommend is to create a table (different name) using Hive connector that points to that same path"?

If that's what you mean, I think we can delete the last sentence `We recommend to create different tables in different systems but point to the same path.`

zsxwing: The point is:

- If you want to query it in Spark, create the table in Spark and point it to the table path.
- If you want to query it in Presto, create the table in Presto and point it to the table path.
- If you want to query it in Hive, create the table in Hive and point it to the table path.

scottsand-db: sgtm

scottsand-db (approved):

PR #261, opened 2022-01-19
Update readme to point to delta.io for Delta Standalone

allisonport-db:
- merge after delta.io is deployed
- todo: delete the wiki page

PR #277, opened 2022-02-08
Fix hive assembly JAR name

allisonport-db:
Currently creates `hive-assembly-assembly_2.11-0.3.0.jar`

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -179,7 +179,7 @@ lazy val hiveAssembly = (project in file("hive-assembly")) dependsOn(hive) setti
   skipReleaseSettings,
 
   assembly / logLevel := Level.Info,
-  assembly / assemblyJarName := s"${name.value}-assembly_${scalaBinaryVersion.value}-${version.value}.jar",
+  assembly / assemblyJarName := s"${name.value}_${scalaBinaryVersion.value}-${version.value}.jar",
   assembly / test := {},
   // Make the 'compile' invoke the 'assembly' task to generate the uber jar.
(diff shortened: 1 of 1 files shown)


scottsand-db:
Fixing the double `assembly` makes sense. But now we are prepending `delta-`, too. Shouldn't this PR be paired with a other PRs to update docs (e.g. README, delta.io website, etc)? And we try and merge them all around the same time ...

Let me know.

scottsand-db:
Or can we just keep it as `hive-` instead of `delta-hive-`

zsxwing:
Good catch. Did you fix the file names when uploading them to https://github.com/delta-io/connectors/releases/tag/v0.3.0 ?

zsxwing (approved):

allisonport-db:
What references in README, or delta.io website? It's currently `delta-hive-assembly` in the README, which is what it was previously.

#198 introduced `hiveAssembly` which changed the sbt command for generating the assembly JAR, and its name.
I already updated the README with the new sbt command in #261 but left the name since I meant to also update this.

allisonport-db:
> Good catch. Did you fix the file names manually when uploading them to https://github.com/delta-io/connectors/releases/tag/v0.3.0 ?

Yep, sorry meant to fix this sooner.

scottsand-db:
> It's currently delta-hive-assembly in the README, which is what it was previously.

Ok sweet, so this change is correct. just double checking :)

scottsand-db (approved):

PR #278, opened 2022-02-09
Flink Delta Sink - Docs Refactor 2

pkubit-g:


flink-connector/src/main/java/io/delta/flink/sink/DeltaSink.java
```diff
@@ -51,182 +35,60 @@
...
- * explicit which methods are directly copied from FileSink we use `FileSink-specific methods`
- * comment marker inside class files to decouple DeltaLake's specific code from parts borrowed
- * from FileSink.
  */
-public class DeltaSink<IN> implements Sink<IN, AbstractDeltaCommittable,
-                                              AbstractDeltaWriterBucketState,
-                                              AbstractDeltaGlobalCommittable>, Logging {
-
-    private final DeltaSinkBuilder<IN> sinkBuilder;
+public class DeltaSink<IN> extends DeltaSinkInternal<IN> {
```
scottsand-db, line 83: why does the public class extend the internal class?
shouldn't the internal class extend the public class?

pkubit-g:
@tdas @scottsand-db 
i have some interesting findings. 
I've done this second refactor as you wished but i forgot one important detail that in the main `DeltaSink` (and not its internal version) we would still have:
```
public class DeltaSink<IN> extends DeltaSinkInternal<IN>
```
which shouldn't work as we have tried it before.

However, i did forgot that fact and went through the refactor and ... realized that it's working. So i went further and removed all the abstract interfaces that we've created and ... it's still working and the docs are being generated. Not sure what exactly happened but the last time we've been trying it was before rebasing on master and from what i can see - in the second part of January (like in the last 2-3 weeks) there was some versions upgrades in the `connectors` repo (e,g, sbt major version upgrade, doc plugin upgrade etc) so this seems to have resolved our issues and the docs generation no longer fails when generating the docs even with some excluded classes in the method's or class's interface. 

Here's a dummy example (`DeltaWriterBucketState` comes from the excluded `internal` package):
Screenshot from 2022-02-10 16-13-30

So it seems that we went through all this abstract interfaces hell for nothing as all we needed was tools/versions upgrade :) 
But good it works now.
So please go again through the docses and let me know your remarks.

This is the current docs:
Screenshot from 2022-02-10 16-15-41

And the only disadvantage is this part:
```
public class DeltaSink<IN>
extends <any>
```
as seen here
Screenshot from 2022-02-10 16-16-35

but i believe that we can live with that ?

pkubit-g:
Also, there is another small downside - compilation during docs generation shows some errors (exactly where expected) but it doesn't influence the successfull generation of this docs.
Screenshot from 2022-02-10 16-55-28

flink-connector/src/main/java/io/delta/flink/sink/DeltaSinkBuilder.java
```diff
@@ -18,112 +18,56 @@
...
 import org.apache.flink.streaming.api.functions.sink.filesystem.OutputFileConfig;
 import org.apache.flink.streaming.api.functions.sink.filesystem.rollingpolicies.CheckpointRollingPolicy;
+import org.apache.flink.table.data.RowData;
 import org.apache.flink.table.types.logical.RowType;
 import org.apache.hadoop.conf.Configuration;
-import static org.apache.flink.util.Preconditions.checkNotNull;
 
 /**
  * A builder class for {@link DeltaSink}.
  * <p>
```
scottsand-db, line 37: Can you add: why would you use this builder class instead of using `DeltaSource.forRowData`?

flink-connector/src/main/java/io/delta/flink/sink/DeltaSink.java
```diff
@@ -51,182 +35,60 @@
  * Parquet format and commits those files to the {@link DeltaLog}. This sink achieves exactly-once
  * semantics for both {@code BATCH} and {@code STREAMING}.
  * <p>
+ * To create new instance of the sink for stream of {@link RowData}:
```
scottsand-db, line 38: Document that for most cases, users should user `DeltaSink.forRowData` instead of the builder, and why

flink-connector/src/main/java/io/delta/flink/sink/DeltaSinkBuilder.java
```diff
@@ -146,7 +90,34 @@ protected DeltaSinkBuilder(
         );
     }
 
-    protected DeltaSinkBuilder(
+    /**
+     * Creates instance of the builder for {@link DeltaSink}.
+     *
+     * @param basePath            path to a Delta table
+     * @param conf                Hadoop's conf object
+     * @param bucketCheckInterval interval for triggering {@link Sink.ProcessingTimeService} within
```
scottsand-db, line 98: seconds? milliseconds? unit

scottsand-db (approved):
LGTM. Left 2 minor comments.

scottsand-db:
[javaunidoc.zip](https://github.com/delta-io/connectors/files/8043278/javaunidoc.zip)

for commit 573270c2770874d35de12b109a1abf9ee0eb732b

flink-connector/src/main/java/io/delta/flink/sink/DeltaSinkBuilder.java
pkubit-g: added

flink-connector/src/main/java/io/delta/flink/sink/DeltaSink.java
pkubit-g: done

flink-connector/src/main/java/io/delta/flink/sink/DeltaSinkBuilder.java
pkubit-g: added

PR #289, opened 2022-03-07
Fix flink connector `publishM2` command

scottsand-db:
- `build/sbt flinkConnector/publishM2` command now works
- for some reason, when running `publishM2`, the javadocs generation seems to be using the values from `Compile / doc / javacOptions` instead of from `JavaUnidoc / unidoc / javacOptions`. Not sure why. `standalone` project doesn't have this issue
- nonetheless... the problem is now solved.


zsxwing (approved):

scottsand-db:
@pawel-kubit-cs can you verify on your end? after that I will merge.

pkubit-g:
@scottsand-db yes, it does work. Thank you !

pkubit-g:
@scottsand-db yes, it does work. Thank you !

PR #291, opened 2022-03-10
Flink Delta Sink - Update examples

pkubit-g:


examples/flink-example/src/main/java/io/delta/flink/example/sink/DeltaSinkExampleBase.java
```diff
@@ -13,19 +15,26 @@
 import org.apache.flink.table.types.logical.RowType;
 import org.apache.flink.table.types.logical.VarCharType;
 
-import java.io.File;
-import java.io.IOException;
-import java.util.Arrays;
-import java.util.Collections;
```
pkubit-g, line 19: sorry for that, but it wasn't properly formatted before (according to the project's style settings) so i'm correcting it now

examples/flink-example/src/main/java/io/delta/flink/example/sink/DeltaSinkExampleBase.java
```diff
@@ -13,19 +15,26 @@
...
+                   rootPath + "/src/main/resources/" + resourcesTableDir :
+                   rootPath + "/flink-example/src/main/resources/" + resourcesTableDir;
+    }
+
     static int PRINT_PAD_LENGTH = 4;
 
     public static final RowType ROW_TYPE = new RowType(Arrays.asList(
-            new RowType.RowField("f1", new VarCharType(VarCharType.MAX_LENGTH)),
-            new RowType.RowField("f2", new VarCharType(VarCharType.MAX_LENGTH)),
-            new RowType.RowField("f3", new IntType())
```
pkubit-g, line 28: sorry for that, but it wasn't properly formatted before (according to the project's style settings) so i'm correcting it now.

Same goes for all of the below

Issue #298, opened 2022-03-23
I was surprised to find parquet4s integration first, but than even more so how dated library is


MironAtHome:
Hey, if we are using Marcin Jakubowski's parquet4s for standalone parquet read / write IO, than why is it frozen at ancient version?
I think it's time to advance platform forward, since 2.3.0 is released more than a month ago and currently 2.4.0 is about to be completed.
@mjakubowski84

scottsand-db:
@MironAtHome - thanks for this issue. are you interested in doing that update + code changes?

vkorukanti:
This repo has been deprecated and the code is moved under `connectors` module in https://github.com/delta-io/delta repository. Please create the issue in repository https://github.com/delta-io/delta. See delta-io/connectors#556 for details.

PR #303, opened 2022-03-25
Update parquet4s library (Fix issue #298)

MironAtHome:
Update parquet4s library from 1.2.1 to 2.3.0


MironAtHome:
[parquet4s major library supporting delta DSW does not support Scala 2.11 since 2.3.0 version
It supports 2.12, 2.13 and 3.0

mjakubowski84:
Parquet4S depends on much newer version of Parquet. Delta DSW overrides it with a version 1.10.1 which is more than 3 years old: https://github.com/delta-io/connectors/blob/master/build.sbt#L46

standalone/src/main/scala/io/delta/standalone/internal/actions/MemoryOptimizedLogReplay.scala
```diff
@@ -56,7 +56,7 @@ private[internal] class MemoryOptimizedLogReplay(
           Some(new CustomJsonIterator(logStore.read(nextFile, hadoopConf)))
         } else if (nextFile.getName.endsWith(".parquet")) {
           val parquetIterable = ParquetReader.read[Parquet4sSingleActionWrapper](
-            nextFile.toString,
+            Parquet4sPath(nextFile.toString),
```
mjakubowski84, line 59: Redundant `toString`

MironAtHome:
> Parquet4S depends on much newer version of Parquet. Delta DSW overrides it with a version 1.10.1 which is more than 3 years old: https://github.com/delta-io/connectors/blob/master/build.sbt#L46

Updated to version matching parquet4s, 1.12.2
Similarly hadoop version leveled to 3.3.1
spark version 3.2.1

zsxwing:
@MironAtHome thanks for the contribution. As we need to support Flink, we have to support Scala 2.11 right now. Could you find a parquet4s version that still supports Scala 2.11? In addition, could you update the PR description to reflect the change in the PR, such as `Upgrade parquet4s to XXX`? You can refer to the original issue in the PR description using `Fixes #298`.

MironAtHome:
> Update parquet4s library from 1.2.1 to 2.3.0
I believe your ask "Upgrade parquet4s to XXX" is in the PR name
"Update parquet4s library from 1.2.1 to 2.3.0"
----
I can see logic in your request.
At the same time you have to see to the fact, that with introduction of Delta-DSW this is no longer sustainable in the format prior, since writing into Parquet using older edition of parquet4s involves difficulties, while writing through parquet-mr is incompatible with old version reading facilities.
So, bringing up version to 2.3.0 of parquet4s is the least that can be done to support move from Delta-DSR to Delta-DSW, while older edition can still be used to support Flink?
Now, if this is a hard ask, we can probably shade Flink or decouple it with its own version of parquet4s?
If you could please help me to split it out into its own project ( if it is not already ) I am happy to make it.

zsxwing:
> I believe your ask 'update the PR description to reflect the change in the PR' is in the PR's name 'update parquet4s version from 1.2.1 to 2.3.0'

Yep. I meant the title.

Take your time to work on the PR 😄

ckipp01:
Thanks for this pr! I actually just hit on this today as we would like to use this but hit on a version conflict due to `delta-standalone` using such an old version of `parquet4s-core` (1.2.1 vs 2.5.1). 

> As we need to support Flink, we have to support Scala 2.11 right now.

Just for context, if I'm not mistaken there is also 2.12 support for Flink, so why is support for 2.11 needed, especially seeing that 2.12.x was released in 2016, meaning that there was almost 6 full years for people to get off 2.11. I know much of the ecosystem has dropped support for 2.11 already, meaning that the pursuit to support 2.11 will continually become more and more challenging causing really old version of dependencies to have to stick around.

> @MironAtHome Any luck to find a version that supports Scala 2.11?

From looking around it seems that 1.9.4 is that last version that supported 2.11. It look like with the 2.0.x series they dropped support for 2.11 and introduced support for Scala 3.

scottsand-db:
@zsxwing we may eventually upgrade to flink 1.15 when we add SQL/Table API/Catalog support. But we don't plan on doing that soon (i.e. this month).

scottsand-db:
@zsxwing - edit, Table API support will use Flink 1.13.

scottsand-db:
As @ckipp01 wrote:
> From looking around it seems that 1.9.4 is that last version that supported 2.11.

@MironAtHome - can you change this PR to support parquet4s 1.9.4?

PR #308, opened 2022-03-29
Fix `build.sbt` for flink connector release

allisonport-db:
- fixes `build.sbt` publishing the flink connector for two versions of scala, which caused a double publishing error
- adds `autoScalaLibrary := false` for flink connector which excludes `scala-lang` from the dependenceis
- halts release for `sql-delta-import` which will be waiting to release until next version
    - also needs `releaseSettings` moved up before updating any `releaseSettings` variables (was previously publishing for all 3 versions)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -669,7 +669,7 @@ lazy val flinkConnector = (project in file("flink-connector"))
     name := "flink-connector",
     commonSettings,
     releaseSettings,
-    publishArtifact := scalaBinaryVersion.value == "2.12",
+    publishArtifact := scalaBinaryVersion.value == "2.12", // only publish once
     autoScalaLibrary := false, // exclude scala-library from dependencies
     Test / publishArtifact := false,
     pomExtra :=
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -667,11 +667,11 @@ lazy val flinkConnector = (project in file("flink-connector"))
   .enablePlugins(GenJavadocPlugin, JavaUnidocPlugin)
   .settings (
     name := "flink-connector",
-    scalaVersion := "2.12.8",
     commonSettings,
-    publishArtifact := scalaBinaryVersion.value != "2.13",
-    Test / publishArtifact := false,
     releaseSettings,
+    publishArtifact := scalaBinaryVersion.value == "2.12",
```
scottsand-db, line 672: nit: can you add a comment like `// we only want to publish once` ? if you think it is unnecessary, then sounds good to me

scottsand-db (approved):
LGTM with one comment

PR #305, opened 2022-03-26
Update integration tests for 0.4.0 release

allisonport-db:
- Minor refactoring to existing tests
- Update flink integration tests to allow for specifying version & maven repository for release
- Add copyrights

PR #309, opened 2022-03-29
Flink integration tests check if table directory exists

allisonport-db:
Thought I accidentally deleted necessary folder structure in #305 by removing the `.gitkeep`s in `examples/flink-example/src/main/resources/...` but realized the integration test actually deletes them when running. Without them, we get a `NullPointerException` because the folders don't exist.

Instead updated the test to first check if the directory exists, and then either clear it or create it accordingly.


scottsand-db:
ya that's my bad, we needed the `.gitkeep`. i missed that in the PR review.

but this works too.

scottsand-db (approved):

PR #312, opened 2022-03-31
Rename Flink connector artifact & folder

allisonport-db:
- Move `flink-connector` to `flink` and update all corresponding instances.
- Change artifact name to `delta-flink` and update necessary docs & other instances

Remaining to update:
- [ ] #310 move documentation to folder `delta-flink` instead of `flink-connector`


build.sbt
```diff
@@ -530,8 +530,8 @@ lazy val standalone = (project in file("standalone"))
         .map(_.filterNot(_.getCanonicalPath.contains("/internal/")))
```
scottsand-db, line 530: tbh we should update this in another PR to use `filter` instead of `filterNot` to just check that it does contain `standalone` ... instead of checking what it does _not_ contain

scottsand-db (approved):
LGTM

build.sbt
allisonport-db: yeah agreed I can do this later

PR #323, opened 2022-04-20
use dummy_col to simply the creating-table syntax

YannByron:
when we create external delta table in Hive, we have to provide the all columns with the right name and the right data type.
here, we can just use `dummy_col array<string>` to simply and make it work. The SQL like: 

`
CREATE TABLE table_name (dummy_col array<string>) STORED BY 'io.delta.hive.DeltaStorageHandler' LOCATION "/path/to/delta";
`

### DIFF
--- a/hive-test/src/test/scala/io/delta/hive/HiveConnectorTest.scala
+++ b/hive-test/src/test/scala/io/delta/hive/HiveConnectorTest.scala
@@ -462,9 +462,9 @@ abstract class HiveConnectorTest extends HiveTest with BeforeAndAfterEach {
             "dummy_col array<string>"
           } else {
             s"""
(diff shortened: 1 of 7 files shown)


YannByron:
@zsxwing PTAL.

zsxwing:
The syntax is weird. Have you seen any issues when trying Hive SerDe? The major work would be converting Delta's schema to Hive schema. That's some work but should be doable.

YannByron:
> The syntax is weird. Have you seen any issues when trying Hive SerDe? The major work would be converting Delta's schema to Hive schema. That's some work but should be doable.

yep, it's weird, i also feel that. i will try to check out the Serde usage. Then close it.

PR #335, opened 2022-05-04
[324] Fix examples

### LINKED ISSUE #324: None of the Java example project can be run due to compilation error
All Java example projects under https://github.com/delta-io/connectors/tree/master/examples
are facing a compilation error preventing them to build and run.

The issue is that class package declaration does not reflects an actual folder structure.
image

To match the package declaration which is `package example` the folder structure should be like this:
`src/main/java/example`

This problem occurs in every Java example project:
1. Convert-to-delta
2. flink-example (originally was ok but was changed in PR https://github.com/delta-io/connectors/pull/305/files)
3. hello-world

To load for example a "hello-world" module as a java project you need to:
1. right click on java folder under "hello-world" module and mark it as a source root folder:
image
2. Right click on pom.xml -> Add as maven project
image

scottsand-db:
- resolves #324 
- refactor some example classes (e.g. to `example` folder)
- update class comments for how to run in IntelliJ
- update `run_examples.py` with how to run locally vs with staged jar
- fix flink `publishM2` javadoc error due to missing javadoc annotation

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -734,6 +734,7 @@ lazy val flink = (project in file("flink"))
       "-windowtitle", "Flink/Delta Connector " + version.value.replaceAll("-SNAPSHOT", "") + " JavaDoc",
       "-noqualifier", "java.lang",
       "-tag", "implNote:a:Implementation Note:",
+      "-tag", "apiNote:a:API Note:",
       "-Xdoclint:all"
     ),
     Compile / doc / javacOptions := (JavaUnidoc / unidoc / javacOptions).value,
--- a/examples/run_examples.py
+++ b/examples/run_examples.py
@@ -96,7 +96,8 @@ class WorkingDirectory(object):
 if __name__ == "__main__":
     """
     Script to run integration tests which are located in the examples directory.
-    Call this by running "python3 run-examples.py".
+    Call this by running "python3 run-examples.py --version <version>", where <version> is the
+    Delta Connectors repo version to use.
     
     There are two version 'modes' you should use to run this file.
     1. using published or staged jar: explicitly pass in the --version argument.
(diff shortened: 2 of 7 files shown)


examples/run_examples.py
```diff
@@ -96,17 +96,29 @@ def __exit__(self, tpe, value, traceback):
...
+    
+    There are two version 'modes' you should use to run this file.
+    1. using published or staged jar: explicitly pass in the --version argument.
+    2. using locally-generated jar (e.g. 0.4.0-SNAPSHOT): explicitly pass in the --version argument
+       and --use-local-cache argument.
+       
+       In this mode, ensure that the local jar exists for all scala versions. You can generate it
+       by running the following commands in the root connectors folder.
+       
+       build/sbt '++2.11.12 publishM2'
```
scottsand-db, line 110: todo double check this is needed

scottsand-db: double checked. it is needed.

build.sbt
```diff
@@ -734,6 +734,7 @@ lazy val flink = (project in file("flink"))
       "-windowtitle", "Flink/Delta Connector " + version.value.replaceAll("-SNAPSHOT", "") + " JavaDoc",
       "-noqualifier", "java.lang",
       "-tag", "implNote:a:Implementation Note:",
+      "-tag", "apiNote:a:API Note:",
```
zsxwing, line 737: What does this do?

zsxwing (approved):
LGTM. Left one question

build.sbt
scottsand-db: the flink connector uses an annotation for one of its classes. these are custom javadoc annotations. we need to explicitly allow them here.

PR #337, opened 2022-05-09
Automatically select LogStore based on path scheme by adding DelegatingLogStore

allisonport-db:
• Adds `DelegatingLogStore` which automatically selects LogStore based on path scheme (largely copied over from delta)
• Updates `LogStoreProvider` to default to `DelegatingLogStore`
• Adds `DelegatingLogStoreSuite` and `LogStoreProviderSuite` (somewhat copied over from delta `DelegatingLogStoreSuite`)

PR #338, opened 2022-05-11
Use `delta-storage` in Delta Standalone

allisonport-db:
Based off of #337 (merge that first to view changes, or view last 2 commits).

## Changes Introduced

- Delete Standalone's java LogStore API, and instead use the one in `delta-storage` (this consolidates our public Java API with https://github.com/delta-io/delta)
- Change the default `LogStore`s for `DelegatingLogStore` to be the `delta-storage` Java implementations instead of standalone's internal scala implementations
    - This also adds support for GCS

#### Note

In a later release we will remove these deprecated scala implementations, but we keep them for now as a fall-back.

## Testing

- Adds GCS scheme and the `delta-storage` LogStore classes to `DelegatingLogStoreSuite`
- Adds "public" LogStore suites to `LogStoreSuite` for each of the `delta-storage` LogStores. These tests are addition to more extensive tests in https://github.com/delta-io/delta/blob/master/core/src/test/scala/org/apache/spark/sql/delta/LogStoreSuite.scala, and I feel are sufficient, but I am open to other opinions here.

- Coming in a follow up PR: verification that these changes make Delta Standalone compatible with `delta-storage-s3-dynamodb`
    - Update: verified with an example, will clean up and open a PR to add it to `connectors/examples`


allisonport-db:
Will update `MiMa` to filter out missing class error with `io.delta.standalone.storage.LogStore` tomorrow :)

standalone/src/main/scala/io/delta/standalone/internal/storage/LogStoreProvider.scala
```diff
@@ -16,36 +16,59 @@
...
 private[internal] object LogStoreProvider extends LogStoreProvider
 
 private[internal] trait LogStoreProvider {
 
-  val defaultLogStoreClassName: String = classOf[HDFSLogStore].getName
+  val logStoreClassConfKey: String = StandaloneHadoopConf.LOG_STORE_CLASS_KEY
+  val defaultLogStoreClass: String = classOf[DelegatingLogStore].getName
+
+  // The conf key for setting the LogStore implementation for `scheme`.
+  def logStoreSchemeConfKey(scheme: String): String = s"io.delta.standalone.logStore.${scheme}.impl"
```
scottsand-db, line 37: what is the verdict on this?

PR #341, opened 2022-05-12
[340] Fix  java.io.IOException when writing delta log to non local path

### LINKED ISSUE #340: Getting java.io.IOException: Cannot create file `file://<tablePath>` when using Flink/Delta Connector with S3
Hi! I'm getting  a `java.io.IOException: Cannot create file file:/<tablePath>` when the delta connector for flink tries to write the DeltaLog for a DataLake hosted in an S3 bucket. I'm using the delta-flink v0.4.0 connector.

Two things came to my attention: 

1. The connector is trying to write the delta log in a local directory instead of the S3 bucket
2. The delta table path is different than the one I specified. I'm using a table path like `s3a://<bucket>/<some>/<path>` and the path that the DeltaLog is trying to access is something like `file:/<some>/<path>`

I'm using something similar to the following code to initialize the DeltaSink

```scala
private val schema = new RowType(
    Seq(new RowType.RowField("text", new VarCharType(VarCharType.MAX_LENGTH)),
      new RowType.RowField("date", new VarCharType(VarCharType.MAX_LENGTH)),
    ).asJava
  )
def createDeltaSink: DeltaSink[RowData] = {
    val hadoopConf = new Configuration()
    hadoopConf.set("io.delta.standalone.LOG_STORE_CLASS_KEY", "io.delta.standalone.internal.storage.S3SingleDriverLogStore")
    DeltaSink.forRowData(new Path("s3a://myBucket/myTable"),
      hadoopConf,
      schema)
      .withPartitionColumns("date")
      .build()
  }
```

A redacted full stack trace is like the following
```
java.io.IOException: Cannot create file:/myTable/_delta_log at io.delta.standalone.internal.DeltaLogImpl.ensureLogDirectoryExist(DeltaLogImpl.scala:165) ~[?:?]at io.delta.standalone.internal.OptimisticTransactionImpl.prepareCommit(OptimisticTransactionImpl.scala:259) ~[?:?]at io.delta.standalone.internal.OptimisticTransactionImpl.commit(OptimisticTransactionImpl.scala:113) ~[?:?]at io.delta.flink.sink.internal.committer.DeltaGlobalCommitter.doCommit(DeltaGlobalCommitter.java:318) ~[?:?]at io.delta.flink.sink.internal.committer.DeltaGlobalCommitter.commit(DeltaGlobalCommitter.java:221) ~[?:?]at org.apache.flink.streaming.runtime.operators.sink.StreamingGlobalCommitterOperator.commit(StreamingGlobalCommitterOperator.java:88) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at org.apache.flink.streaming.runtime.operators.sink.AbstractStreamingCommitterOperator.commitUpTo(AbstractStreamingCommitterOperator.java:155) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at org.apache.flink.streaming.runtime.operators.sink.AbstractStreamingCommitterOperator.notifyCheckpointComplete(AbstractStreamingCommitterOperator.java:137) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at org.apache.flink.streaming.runtime.operators.sink.StreamingGlobalCommitterOperator.notifyCheckpointComplete(StreamingGlobalCommitterOperator.java:104) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at org.apache.flink.streaming.runtime.tasks.StreamOperatorWrapper.notifyCheckpointComplete(StreamOperatorWrapper.java:99) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at org.apache.flink.streaming.runtime.tasks.SubtaskCheckpointCoordinatorImpl.notifyCheckpointComplete(SubtaskCheckpointCoordinatorImpl.java:334) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at org.apache.flink.streaming.runtime.tasks.StreamTask.notifyCheckpointComplete(StreamTask.java:1178) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at org.apache.flink.streaming.runtime.tasks.StreamTask.lambda$notifyCheckpointCompleteAsync$10(StreamTask.java:1143) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at org.apache.flink.streaming.runtime.tasks.StreamTask.lambda$notifyCheckpointOperation$12(StreamTask.java:1166) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at org.apache.flink.streaming.runtime.tasks.StreamTaskActionExecutor$SynchronizedStreamTaskActionExecutor.runThrowing(StreamTaskActionExecutor.java:93) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at org.apache.flink.streaming.runtime.tasks.mailbox.Mail.run(Mail.java:90) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at org.apache.flink.streaming.runtime.tasks.mailbox.MailboxProcessor.processMailsWhenDefaultActionUnavailable(MailboxProcessor.java:344) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at org.apache.flink.streaming.runtime.tasks.mailbox.MailboxProcessor.processMail(MailboxProcessor.java:330) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at org.apache.flink.streaming.runtime.tasks.mailbox.MailboxProcessor.runMailboxLoop(MailboxProcessor.java:202) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at org.apache.flink.streaming.runtime.tasks.StreamTask.runMailboxLoop(StreamTask.java:681) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at org.apache.flink.streaming.runtime.tasks.StreamTask.executeInvoke(StreamTask.java:636) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at org.apache.flink.streaming.runtime.tasks.StreamTask.runWithCleanUpOnFail(StreamTask.java:647) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at org.apache.flink.streaming.runtime.tasks.StreamTask.invoke(StreamTask.java:620) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at org.apache.flink.runtime.taskmanager.Task.doRun(Task.java:784) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at org.apache.flink.runtime.taskmanager.Task.run(Task.java:571) ~[flink-dist_2.12-1.13.2.jar:1.13.2]at java.lang.Thread.run(Thread.java:829) ~[?:?]
--
```
After some tracing, I believe that the problem may be in the [DeltaGlobalCommiter](https://github.com/delta-io/connectors/blob/925bd79bcae9b0c85e7cd7342e4686e7be2a21d5/flink/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java#L215). I tested by changing the initialization of the DeltaLog to  `DeltaLog.forTable(conf, basePath.toString());` and seems to be working.

Let me know if you need any other information that could help resolving this issue.

Thanks in advance!

pabloFloresHdz:
This PR Fixes #340 by passing the full table path when calling `DeltaLog.forTable`

### DIFF
--- a/flink/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
+++ b/flink/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
@@ -212,7 +212,7 @@ public class DeltaGlobalCommitter
         if (appId != null) { // means there are committables to process
             SortedMap<Long, List<DeltaCommittable>> committablesPerCheckpoint =
                 groupCommittablesByCheckpointInterval(globalCommittables);
-            DeltaLog deltaLog = DeltaLog.forTable(conf, basePath.getPath());
+            DeltaLog deltaLog = DeltaLog.forTable(conf, basePath.toString());
 
             for (long checkpointId : committablesPerCheckpoint.keySet()) {
                 OptimisticTransaction transaction = deltaLog.startTransaction();


flink/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
```diff
@@ -212,7 +212,7 @@ public List<DeltaGlobalCommittable> commit(List<DeltaGlobalCommittable> globalCo
         if (appId != null) { // means there are committables to process
             SortedMap<Long, List<DeltaCommittable>> committablesPerCheckpoint =
                 groupCommittablesByCheckpointInterval(globalCommittables);
-            DeltaLog deltaLog = DeltaLog.forTable(conf, basePath.getPath());
+            DeltaLog deltaLog = DeltaLog.forTable(conf, basePath.toString());
```
zsxwing, line 215: nit: we can pass `basePath` to `forTable` directly as `basePath` is a `Path`.

pabloFloresHdz: Unfortunately, `basePath` is an `org.apache.flink.core.fs.Path` and `forTable` expects an `org.apache.hadoop.fs.Path`. Is there a way to transform between `Paths`?

zsxwing: Ah, I see. Do you know what's the difference between `org.apache.flink.core.fs.Path` and `org.apache.hadoop.fs.Path`? Is `new org.apache.hadoop.fs.Path(org.apache.flink.core.fs.Path.toString)` always working?

zsxwing: Will `new Path(basePath.toUri)` be safer?

pabloFloresHdz: Flink and Hadoop `Path` seem very similar to me, but they implement different interfaces, which I imagine are used by their corresponding frameworks. I can't say that all Flink paths are valid hadoop paths, yet using uri may not be safer as it seems like both implementations allow wildcards in the paths which the `Uri` class escapes

pkubit-g: @pabloFloresHdz first and foremost thanks for reporting the bug and providing the PR !

as for the solution, i've checked and it looks like the wildcards are being preserved in `Uri` class. 

Besides, even though it's not part of this discussion, i'm not sure if in this case wildcards should be allowed as the path should directly point to the Delta's table. @scottsand-db does the delta-standalone allow for wildcards in the table's path ?

Nonetheless, to me it also seems like `toUri()` solution is safer than relying on `toString()` implementation. It's also what Flink is doing in their code, see [toHadoopPath()](https://github.com/apache/flink/blob/master/flink-filesystems/flink-hadoop-fs/src/main/java/org/apache/flink/runtime/fs/hdfs/HadoopFileSystem.java#L222) in `org.apache.flink.runtime.fs.hdfs.HadoopFileSystem`

Here's the sum up for all the methods:
Screenshot from 2022-05-13 14-55-40

pabloFloresHdz: @zsxwing @pkubit-g thank you for your comments! I've updated the code to build a hadoop path using `Path.toUri` as suggested

zsxwing (approved):

scottsand-db (approved):

allisonport-db:
Can we add a unit test? Something like...
- Set `fs.defaultFS` to a non-existent file system
- Create a sink with correct full path

And then verify that without the fix this test fails. Thanks!

tdas:
@pkubit-g can you suggest which test suite will be a good place to add a test. To extend @allisonport-db suggestion, I would just copy one of the simple unit test, and update it with a hadoop conf where `fs.defaultFS` is set to a non-existent file system. Without this fix, that test should fail.

pkubit-g:
@scottsand-db lgtm

@tdas probably `DeltaGlobalCommitterTest` as `DeltaGlobalCommitter` is the only place when we use this `Path` object pointing to the Delta table.

scottsand-db:
Hi @pabloFloresHdz - just following up on the status of this. Would love if you added a unit test, but if not we will merge this PR as is and add a unit test next.

pabloFloresHdz:
> Hi @pabloFloresHdz - just following up on the status of this. Would love if you added a unit test, but if not we will merge this PR as is and add a unit test next.

Sorry for the delay in response, I've not been able to follow up. I attempted to write a unit test for this, but it seem like there is a validation of `fs.defaultFS` if I set it to something invalid, the test won't run. I tried setting it to some local paths, but I can't make the test fail removing the fix. I tried with hdfs and s3a as well, but it tries to connect to those services before running the test. Currently, I'm trying using MiniDFS which seems promising. If anyone has suggestions on how to test with a dummy Filesystem  I would really appreciate the help. 

Thanks in advance

scottsand-db:
cc @pkubit-g

pkubit-g:
@pabloFloresHdz have you tried for instance using `https://github.com/findify/s3mock` ? Looks like it might do the trick. I've never used it personally though. For S3 mocking i have only experience with Python's super great lib called `moto`, but this `findify/s3mock` looks like it may provide similar functionality.  
So we would need to first create this S3 mock object and then pass some configuration to the Hadoop conf's object (like S3 endpoint from the mock etc.) and then pass this very conf to the DeltaGlobalCommitter's constructor along with a test Delta table's path starting with "s3://"

pabloFloresHdz:
@pkubit-g Thanks for your comment! I'll explore that option. I've seen some SO posts suggesting extending hadoop's Filesystem. I will explore this as well when I have a chance

scottsand-db:
Hi @pabloFloresHdz, just an FYI, we would like to release a Delta Connectors 0.4.1 next week - would be great if you could get this test case in by early next week. Just wanted to let you know our timelines. We are here to help!

Xoerk:
Hi, 
Any plans to merge it soon?  
@scottsand-db, where can we see the timelines of planned releases?

flink/src/test/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitterTest.java
```diff
@@ -390,6 +393,46 @@ public void testGlobalCommittableSerializerWithEmptyCommittables() throws IOExce
         assertTrue(deserialized.getDeltaCommittables().isEmpty());
     }
 
+    @Test
+    public void testUseFullPathForDeltaLog() throws Exception {
```
scottsand-db, line 397: Hi @pabloFloresHdz , can you add some inline comments in this test case please?

i'm a little unsure exactly what/how it is testing your fix. a few questions
- what is the scheme of the files in the table?
- is this different than the `mockfs:///` scheme you have set as the default?
- if it is different, or if it is the same, please add inline comments explaining how this tests the fix.

Appreciate the help! We are looking forward to release Delta Connectors 0.4.1 right after this is merged.

pabloFloresHdz: Added comments to the test, please let me know if there are still comments to address

Xoerk:
> [scottsand-db](/scottsand-db) @scottsand-db, thanks for the quick reply! looking forward to the release...

flink/src/test/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitterTest.java
```diff
@@ -390,6 +393,54 @@ public void testGlobalCommittableSerializerWithEmptyCommittables() throws IOExce
...
```
scottsand-db, line 397: running this is giving me error
```
objc[75549]: Class JavaLaunchHelper is implemented in both /Library/Java/JavaVirtualMachines/jdk1.8.0_131.jdk/Contents/Home/bin/java (0x1085e74c0) and /Library/Java/JavaVirtualMachines/jdk1.8.0_131.jdk/Contents/Home/jre/lib/libinstrument.dylib (0x1087c64e0). One of the two will be used. Which one is undefined.
SLF4J: Class path contains multiple SLF4J bindings.
SLF4J: Found binding in [jar:file:/Users/scott.sandre/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/apache/logging/log4j/log4j-slf4j-impl/2.12.1/log4j-slf4j-impl-2.12.1.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: Found binding in [jar:file:/Users/scott.sandre/Library/Caches/Coursier/v1/https/repo1.maven.org/maven2/org/slf4j/slf4j-log4j12/1.7.25/slf4j-log4j12-1.7.25.jar!/org/slf4j/impl/StaticLoggerBinder.class]
SLF4J: See http://www.slf4j.org/codes.html#multiple_bindings for an explanation.
SLF4J: Actual binding is of type [org.apache.logging.slf4j.Log4jLoggerFactory]

java.lang.NoClassDefFoundError: com/fasterxml/jackson/module/scala/ScalaObjectMapper

	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:467)
	at java.net.URLClassLoader.access$100(URLClassLoader.java:73)
```

do you think that
```
"org.apache.hadoop" % "hadoop-common" % hadoopVersion % "test" classifier "tests",
```

is bringing in some extra dependencies?

flink/src/test/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitterTest.java
```diff
@@ -390,6 +393,54 @@ public void testGlobalCommittableSerializerWithEmptyCommittables() throws IOExce
...
+        //GIVEN
+        int numAddedFiles = 3;
+
+        DeltaSinkTestUtils.initTestForPartitionedTable(tablePath.getPath());
```
scottsand-db, line 401: can you assert what the scheme is? and if it is/isn't raw?

flink/src/test/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitterTest.java
pabloFloresHdz: I don't think adding hadoop-common is the problem. I saw that error the whole time in my IDE (even before adding any tests). I just ran the tests using `sbt` from the command line without issues

scottsand-db (approved):

scottsand-db:
will merge once tests pass!

PR #346, opened 2022-05-25
[delta-standalone] Implement "get version at time X" APIs

scottsand-db:
See the design doc [here](https://docs.google.com/document/d/1j9tHRmqSGuTlbwxjnFexaDFGJ-2LUCH6DuWGnGZrMDE/edit#heading=h.4cz970y1mk93).

This PR adds two new APIs
- `long getVersionBeforeOrAtTimestamp(long timestamp)`
- `long getVersionAtOrAfterTimestamp(long timestamp)`

The latter is used to provide the "streaming" semantics needed for the Delta/Flink connector. Specifically, this means that when using the "startingTimestamp" option for a streaming/continuous query, we return the commit version "at or after" that timestamp.

The result of this method (`getVersionAtOrAfterTime`) can then be passed to `DeltaLog::getSnapshotForVersionAsOf` to get the metadata for a particular version, and to `DeltaLog::getChanges` to get the changes from a particular version, for the version corresponding to the input timestamp.

This is different from the "batch" semantics, where for options like "timestampAsOf" we use the snapshot version "at or before" that timestamp.

### DIFF
--- a/standalone/src/main/java/io/delta/standalone/DeltaLog.java
+++ b/standalone/src/main/java/io/delta/standalone/DeltaLog.java
@@ -108,7 +108,7 @@ public interface DeltaLog {
     Iterator<VersionLog> getChanges(long startVersion, boolean failOnDataLoss);
 
     /**
-     * Returns the latest commit that happened before or at {@code timestamp}.
+     * Returns the latest version that was committed before or at {@code timestamp}.
      *
      * Specifically:
      * <ul>
@@ -124,10 +124,10 @@ public interface DeltaLog {
      * @throws IllegalArgumentException if the timestamp is less than the timestamp of any committed
      *                                  version
      */
-    long getVersionBeforeOrAtTime(long timestamp);
+    long getVersionBeforeOrAtTimestamp(long timestamp);
 
     /**
-     * Returns the latest commit that happened at or after {@code timestamp}.
+     * Returns the latest version that was committed at or after {@code timestamp}.
      *
      * Specifically:
      * <ul>
@@ -143,7 +143,7 @@ public interface DeltaLog {
      * @throws IllegalArgumentException if the timestamp is more than the timestamp of any committed
      *                                  version
      */
-    long getVersionAtOrAfterTime(long timestamp);
+    long getVersionAtOrAfterTimestamp(long timestamp);
 
     /**
      * @return Whether a Delta table exists at this directory.
--- a/standalone/src/main/java/io/delta/standalone/DeltaLog.java
+++ b/standalone/src/main/java/io/delta/standalone/DeltaLog.java
@@ -108,6 +108,44 @@ public interface DeltaLog {
     Iterator<VersionLog> getChanges(long startVersion, boolean failOnDataLoss);
 
     /**
+     * Returns the latest commit that happened before or at {@code timestamp}.
+     *
+     * Specifically:
+     * <ul>
+     *     <li>if a commit version exactly matches the provided timestamp, we return it</li>
+     *     <li>else, we return the latest commit version with a timestamp less than the
+     *         provided one</li>
+     *     <li>If the provided timestamp is less than the timestamp of any committed version,
+     *         we throw an error.</li>
+     * </ul>.
+     *
+     * @param timestamp the number of milliseconds since midnight, January 1, 1970 UTC
+     * @return latest commit that happened before or at {@code timestamp}.
+     * @throws IllegalArgumentException if the timestamp is less than the timestamp of any committed
+     *                                  version
+     */
+    long getVersionBeforeOrAtTime(long timestamp);
+
+    /**
+     * Returns the latest commit that happened at or after {@code timestamp}.
+     *
+     * Specifically:
+     * <ul>
+     *     <li>if a commit version exactly matches the provided timestamp, we return it</li>
+     *     <li>else, we return the earliest commit version with a timestamp greater than the
+     *         provided one</li>
+     *     <li>If the provided timestamp is larger than the timestamp of any committed version,
+     *         we throw an error.</li>
+     * </ul>.
+     *
+     * @param timestamp the number of milliseconds since midnight, January 1, 1970 UTC
+     * @return latest commit that happened at or before {@code timestamp}.
+     * @throws IllegalArgumentException if the timestamp is more than the timestamp of any committed
+     *                                  version
+     */
+    long getVersionAtOrAfterTime(long timestamp);
+
+    /**
      * @return Whether a Delta table exists at this directory.
      */
     boolean tableExists();
--- a/standalone/src/main/java/io/delta/standalone/DeltaLog.java
+++ b/standalone/src/main/java/io/delta/standalone/DeltaLog.java
@@ -108,7 +108,8 @@ public interface DeltaLog {
     Iterator<VersionLog> getChanges(long startVersion, boolean failOnDataLoss);
 
     /**
-     * Returns the latest version that was committed before or at {@code timestamp}.
+     * Returns the latest version that was committed before or at {@code timestamp}. If no version
+     * exists, returns -1.
      *
      * Specifically:
      * <ul>
@@ -127,7 +128,8 @@ public interface DeltaLog {
     long getVersionBeforeOrAtTimestamp(long timestamp);
 
     /**
-     * Returns the latest version that was committed at or after {@code timestamp}.
+     * Returns the latest version that was committed at or after {@code timestamp}. If no version
+     * exists, returns -1.
      *
      * Specifically:
      * <ul>
--- a/standalone/src/main/scala/io/delta/standalone/internal/DeltaHistoryManager.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/DeltaHistoryManager.scala
@@ -20,9 +20,10 @@ import java.sql.Timestamp
 
 import scala.collection.JavaConverters._
 
-import io.delta.standalone.DeltaLog
 import org.apache.hadoop.fs.Path
+
 import io.delta.standalone.storage.LogStore
+
 import io.delta.standalone.internal.actions.{Action, CommitInfo, CommitMarker}
 import io.delta.standalone.internal.exception.DeltaErrors
 import io.delta.standalone.internal.logging.Logging
--- a/standalone/src/main/scala/io/delta/standalone/internal/DeltaHistoryManager.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/DeltaHistoryManager.scala
@@ -69,15 +69,22 @@ private[internal] case class DeltaHistoryManager(deltaLog: DeltaLogImpl) extends
   /**
    * Returns the latest commit that happened at or before `time`.
    *
+   * If the given timestamp is outside the range of [earliestCommit, latestCommit] then use params
+   * `canReturnLastCommit` and `canReturnEarliestCommit` to control whether an exception is thrown
+   * or the corresponding earliest/latest commit is returned. See param docs below.
+   *
    * @param timestamp the timestamp to search for
    * @param canReturnLastCommit Whether we can return the latest version of the table if the
    *                            provided timestamp is after the latest commit
    * @param mustBeRecreatable Whether the state at the given commit should be recreatable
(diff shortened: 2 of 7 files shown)


standalone/src/main/scala/io/delta/standalone/internal/DeltaHistoryManager.scala
```diff
@@ -70,13 +70,26 @@ private[internal] case class DeltaHistoryManager(deltaLog: DeltaLogImpl) extends
...
+   * @param canReturnEarliestCommit Whether we can return the earliest commit if no such commit
+   *                                exists.
    * @throws RuntimeException if the state at the given commit in not recreatable
    * @throws IllegalArgumentException if the provided timestamp is before the earliest commit or
    *                                  after the latest commit
    */
-  def getActiveCommitAtTime(timestamp: Timestamp): Commit = {
+  def getActiveCommitAtTime(
+      timestamp: Timestamp,
+      canReturnLastCommit: Boolean = false,
```
scottsand-db, line 84: this defaults introduce no new behaviour to existing usage in `SnapshotManagement::getSnapshotForTimestampAsOf`

standalone/src/main/scala/io/delta/standalone/internal/DeltaHistoryManager.scala
```diff
@@ -87,15 +100,32 @@ private[internal] case class DeltaHistoryManager(deltaLog: DeltaLogImpl) extends
...
 
     commit
   }
 
+  /**
+   * Get the earliest commit available for this table. Note that this version isn't guaranteed to
+   * exist when performing an action as a concurrent operation can delete the file during cleanup.
+   * This value must be used as a lower bound.
+   */
+  def getEarliestDeltaFile(deltaLog: DeltaLogImpl): Long = {
```
scottsand-db, line 117: copied from Runtime.

scottsand-db, line 117: copied pretty much verbatim from Delta OSS (just added a variable instead of doing it inline).

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
```diff
@@ -133,6 +134,45 @@ private[internal] class DeltaLogImpl private(
...
+  override def getVersionBeforeOrAtTime(timestamp: Long): Long = {
+    // Note: if the provided timestamp is earlier than any committed version, then
+    // `getActiveCommitAtTime` will throw IllegalArgumentException (specifically,
+    // `DeltaErrors.timestampEarlierThanTableFirstCommit`).
+    history.getActiveCommitAtTime(
+      new Timestamp(timestamp),
+      // e.g. if we give time T+2 and last commit has time T, then we DO want that last commit
+      canReturnLastCommit = true,
+      mustBeRecreatable = false,
+      // e.g. we give time T-1 and first commit has time T, then do not want that earliest commit
```
scottsand-db, line 147: nit: `NOT` instead of `not`.

standalone/src/main/java/io/delta/standalone/DeltaLog.java
```diff
@@ -107,6 +107,44 @@ public interface DeltaLog {
...
+     *     <li>If the provided timestamp is less than the timestamp of any committed version,
+     *         we throw an error.</li>
+     * </ul>.
+     *
+     * @param timestamp the number of milliseconds since midnight, January 1, 1970 UTC
+     * @return latest commit that happened before or at {@code timestamp}.
+     * @throws IllegalArgumentException if the timestamp is less than the timestamp of any committed
+     *                                  version
+     */
+    long getVersionBeforeOrAtTime(long timestamp);
```
zsxwing, line 127: It's better to call it either `getVersionBeforeOrAt` or `getVersionBeforeOrAtTimestamp`. I prefer `getVersionBeforeOrAtTimestamp`.

standalone/src/main/scala/io/delta/standalone/internal/DeltaHistoryManager.scala
```diff
@@ -70,13 +70,26 @@ private[internal] case class DeltaHistoryManager(deltaLog: DeltaLogImpl) extends
    * Returns the latest commit that happened at or before `time`.
```
allisonport-db, line 70: Can we just improve the docs to this function? I didn't understand how all the parameters affected the output without reading the code.

A) Improve this general doc, since isn't `Returns the latest commit that happened at or before time.` no longer true?
B) canReturnLastCommit - this was a little confusing since the original documentation doesn't really clearly mention that this function will not return the current (aka latest) commit version and would instead throw an error. I think if that is clarified above this specific param doc is okay.
C) canReturnEarliestCommit - this is confusing the way it's phrased. This is if the timestamp is before the first commit of the table correct? If so, we return that first commit. Is there any other scenario where this parameter affects the return value? If not, can we just specifically say this is what it is, ie. "if the provided timestamp is before the first commit of the table return that first commit" or something along those lines?
D) Can we improve the docs for `@throws IllegalArgumentException`? It's confusing... Especially with respect to the two new parameters. This would also be more clear with the added docs about not returning the latest commit version (see B)

standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
```diff
@@ -459,6 +464,75 @@ abstract class DeltaLogSuiteBase extends FunSuite {
       assert(log.tableExists())
     }
   }
+
+  test("getVersionBeforeOrAtTime and getVersionAtOrAfterTime") {
```
allisonport-db, line 468: What happens for a delta table that hasn't been committed yet? Should it be the `FileNotFound` exception I'm seeing or should we see the `IllegalArgumentException` mentioned in the docs?

allisonport-db: Can we test that these functions don't care if that version is reproducible or not?

allisonport-db (commented):
Minor comments. Mostly clarify documentation, and two questions about testing.

standalone/src/main/scala/io/delta/standalone/internal/DeltaHistoryManager.scala
tdas, line 84: you have now 3 booleans, which is 2^3 = 8 combination of input params. do you need all of them? if not why add so much complexity. why not just implement the exact 2-3 versions that you need. is that hard?

scottsand-db: So we need:
- `getVersionBeforeOrAtTimestamp` uses (true, false, false)
- `getVersionAtOrAfterTimestamp` uses (false, false, true)
- `SnapshotManagement::getSnapshotForTimestampAsOf` uses all defaults, which is (false, true, false)

As you can see, we do use true/false values for all 3 booleans.

Also, sure, there are 8 combinations of the booleans, but they are only used for exclusive scenarios
- `mustBeRecreatable` is a must-have, and has no impact on `canReturnEarliestCommit` or `canReturnLastCommit`
- `canReturnEarliestCommit` is only relevant if `commit.timestamp > time`
- `canReturnLastCommit` is only relevant if `commit.timestamp < time`

Thus, the combination of `canReturnEarliestCommit` and `canReturnLastCommit` have no impact on one another.

Also, this is just copying what Delta Lake does [here](https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/DeltaHistoryManager.scala#L113-L117). This means we've circled back around to the question of: do we copy what Delta Lake does, and make debugging errors more consistent or easier, or do something different? I prefer copying what Delta Lake does. Plus, in this code, I think this code makes sense.

Let me know what you think.

standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
```diff
@@ -459,6 +464,75 @@ abstract class DeltaLogSuiteBase extends FunSuite {
...
+      val logPath = new Path(dir.getCanonicalPath, "_delta_log")
+      val logDir = new File(dir.getCanonicalPath, "_delta_log")
+      // local file system truncates to seconds
+      val nowEpochMs = System.currentTimeMillis() / 1000 * 1000
+
+      val delta0 = FileNames.deltaFile(logPath, 0)
+      val delta1 = FileNames.deltaFile(logPath, 1)
+      val delta2 = FileNames.deltaFile(logPath, 2)
+
+      new File(logDir, delta0.getName).setLastModified(nowEpochMs)
```
tdas, line 495: i think its easier to read the test if you actually use constants. like `10000`, `20000`, `30000`

tdas (changes requested):
i am not sure of the complexity of the param list

standalone/src/main/scala/io/delta/standalone/internal/DeltaHistoryManager.scala
tdas: oh i see. the existing code is already messy, and we are copying from it. thanks for clarifying that. i dont quite like it, but in the interest of consistency, I am okay with this. however lets improve the docs further as @allisonport-db has also commented.

standalone/src/main/scala/io/delta/standalone/internal/DeltaHistoryManager.scala
```diff
@@ -70,13 +70,26 @@ private[internal] case class DeltaHistoryManager(deltaLog: DeltaLogImpl) extends
...
+   * @param canReturnLastCommit Whether we can return the latest version of the table if the
+   *                            provided timestamp is after the latest commit
+   * @param mustBeRecreatable Whether the state at the given commit should be recreatable
+   * @param canReturnEarliestCommit Whether we can return the earliest commit if no such commit
+   *                                exists.
    * @throws RuntimeException if the state at the given commit in not recreatable
    * @throws IllegalArgumentException if the provided timestamp is before the earliest commit or
    *                                  after the latest commit
    */
-  def getActiveCommitAtTime(timestamp: Timestamp): Commit = {
```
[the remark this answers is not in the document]
scottsand-db: good question. we use this because Delta Lake uses `Timestamp`.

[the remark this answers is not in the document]
scottsand-db: we can refactor this later if we chose

standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
scottsand-db: done

standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
```diff
@@ -492,47 +496,114 @@ abstract class DeltaLogSuiteBase extends FunSuite {
...
 
-      new File(logDir, delta0.getName).setLastModified(nowEpochMs)
-      new File(logDir, delta1.getName).setLastModified(nowEpochMs + 1.minutes)
-      new File(logDir, delta2.getName).setLastModified(nowEpochMs + 2.minutes)
+      val TIME_BEFORE_START = nowEpochMs - 1.minutes
+      val TIME_START = nowEpochMs
+      val TIME_PLUS_30_SECS = nowEpochMs + 30.seconds
+      val TIME_PLUS_ONE_MINUTE = nowEpochMs + 1.minutes
+      val TIME_END = nowEpochMs + 2.minutes
+      val TIME_PAST_END = nowEpochMs + 3.minutes
```
allisonport-db, line 504: Btw, I think @tdas comment here was to just use actual number constants and not have these variables.

allisonport-db (approved):
LGTM

standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
scottsand-db: darn, i see. thanks!

scottsand-db: updated.

PR #353, opened 2022-05-31
[delta-standalone] Add check that schema contains partition columns

scottsand-db:
This PR adds a check to OptimisticTransaction to ensure that metadata schemas actually contain the metadata partition columns during writes. We also check that there is at least one non-partition column. This check is typically performed in TransactionalWrite (in Delta Lake) [here](https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/files/TransactionalWrite.scala#L138-L151) however delta-standalone doesn't perform data writes, only metadata (i.e. transaction log) writes. Thus, this check was never added before now.

Before this fix, it was perfectly possible to write `schema={col1, col2}` and `partitionCols={part1, part2}` (partitions not a subset of schema).
It was also perfectly possible to write `schema={part1, part2}` and `partitionCols={part1, part2}` (no data columns).

The first commit adds the actual relevant code: 49c052dd045da281739bff3c4b7f67093b276eef.

The next commits iteratively fix all of the tests that were broken.


scottsand-db:
cc @kristoffSC

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
```diff
@@ -206,6 +207,7 @@ private[internal] class OptimisticTransactionImpl(
     }
 
     verifyNewMetadata(latestMetadata)
+    checkPartitionColumns(latestMetadata.partitionColumns, latestMetadata.schema)
```
tdas, line 210: why is this check not already part of the function `verifyNewMetadata` which seem to verify all of the metadata, including schema?

standalone/src/main/scala/io/delta/standalone/internal/exception/DeltaErrors.scala
```diff
@@ -332,6 +332,16 @@ private[internal] object DeltaErrors {
         " cannot be set at the same time. Please set only one group of them.")
   }
 
+  def partitionColumnNotFoundException(partCol: String, schema: StructType): Throwable = {
+    new DeltaStandaloneException(s"Partition column $partCol not found in schema:\n" +
+      s"${schema.getTreeString}")
+  }
+
+  def nonPartitionColumnAbsentException(): Throwable = {
```
tdas, line 340: why absent for one error and not found for the other error?

tdas (changes requested):
- few minor comments
- in addition, are there checks that all the fileactions have specified all the partition column values?

standalone/src/main/scala/io/delta/standalone/internal/exception/DeltaErrors.scala
scottsand-db: just following what Delta Lake is currently using.
I advocate for keeping method names the same when we can. It makes searching them up in the two code bases easier.

scottsand-db:
@tdas 

> are there checks that all the fileactions have specified all the partition column values?

Yes. We check that here
https://github.com/delta-io/connectors/blob/master/standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala#L281

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
scottsand-db: it can be. as I responded on your other comment, I do try and keep the method names between delta-standalone and delta-lake as similar, or identical, as possible. this helps in debugging the code (e.g. if a bug arises in delta-lake, similar code and method names in delta-standalone will make it easier to fix here, too).

what do you think?

flink/src/test/java/io/delta/flink/sink/DeltaSinkBatchExecutionITCase.java
```diff
@@ -75,7 +75,7 @@ public void runDeltaSinkTest() throws Exception {
         List<AddFile> initialDeltaFiles = deltaLog.snapshot().getAllFiles();
         int initialTableRecordsCount = TestParquetReader.readAndValidateAllTableRecords(deltaLog);
         long initialVersion = deltaLog.snapshot().getVersion();
-        assertEquals(initialDeltaFiles.size(), 2);
+        assertEquals(2, initialDeltaFiles.size());
```
tdas, line 78: why this change?

kristoffSC: Hi,
I can take this one, since this is my change. 
`assertEqual` API is ` assertEqual(expected, actual).
 
Originally it was the other way around, so our "actual" result was used as "expected" value in assertion. That resulted with confusing assert message if test failed.
 
 There is some inconsistency between Junit API and other libraries that provides assert methods :)
`

tdas (approved):
i am good  with the prod code changes as long as @kristoffSC has reviewed all test code

kristoffSC:
@tdas 
All changes in Flink connector tests are actually mine, merged into Scott's PR from this:
https://github.com/scottsand-db/connectors/pull/46

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
```diff
@@ -461,6 +467,24 @@ private[internal] class OptimisticTransactionImpl(
...
+   * Check that the schema contains all partition columns and at least one non-partition column
+   */
+  private def checkPartitionColumns(partitionCols: Seq[String], schema: StructType): Unit = {
+    // schema contains all partition column
+    val schemaCols = schema.getFieldNames.toSet
+    partitionCols.foreach { partCol =>
+      if (!schemaCols.contains(partCol)) {
+        throw DeltaErrors.partitionColumnNotFoundException(partCol, schema)
+      }
+    }
```
allisonport-db, line 480: Minor thought, this could instead check that `partitionCols` is a subset of `schemaCols`, then our error can list all missing partition columns instead of just one by one, if there were more than 1 omitted from the schema.

This is a minor nit, no biggie either way.

standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionLegacySuite.scala
```diff
@@ -560,7 +565,7 @@ class OptimisticTransactionLegacySuite extends FunSuite {
       val add = AddFile("test", Map.empty, 1, 1, dataChange = false)
 
       val log0 = DeltaLog.forTable(new Configuration(), dir.getCanonicalPath)
-      log0.startTransaction().commit(Metadata() :: add :: Nil, manualUpdate, engineInfo)
+      log0.startTransaction().commit(metadata_2 :: add :: Nil, manualUpdate, engineInfo)
```
allisonport-db, line 568: why is this one `metadata_2`?

standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionLegacySuite.scala
```diff
@@ -612,11 +617,11 @@ class OptimisticTransactionLegacySuite extends FunSuite {
       val tx1 = log.startTransaction()
 
       val tx2 = log.startTransaction()
-      tx2.updateMetadata(ConversionUtils.convertMetadata(Metadata()))
+      tx2.updateMetadata(metadataJ)
       tx2.commit(Iterable().asJava, manualUpdate, engineInfo)
 
       assertThrows[MetadataChangedException] {
-        tx1.updateMetadata(ConversionUtils.convertMetadata(Metadata()))
+        tx1.updateMetadata(metadataJ_2)
```
allisonport-db, line 624: Isn't this conceptually changing the test? Previously it wasn't changing the table schema.

scottsand-db: I could keep `metadataJ`, and the test still passes, but I think it's better to test two conflicting schemas instead.

scottsand-db: but I'm easy.

standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionLegacySuite.scala
scottsand-db: I can change it to `metadataJ`.

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
scottsand-db: sgtm

standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionLegacySuite.scala
allisonport-db: I mean I think the failure with updating the same metadata is the more extreme version of the test. Like I'd assume if that throws an error, two different metadatas would. Not sure the same can be said for vice-versa

scottsand-db: Hm. You could easily argue though that two concurrent txns that update with the exact same metadata shouldn't throw an error.

scottsand-db: I'll just do both :)

allisonport-db: i agree.. but then were testing incorrect behavior and should change the impl :/

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
```diff
@@ -473,10 +473,11 @@ private[internal] class OptimisticTransactionImpl(
   private def checkPartitionColumns(partitionCols: Seq[String], schema: StructType): Unit = {
     // schema contains all partition column
     val schemaCols = schema.getFieldNames.toSet
-    partitionCols.foreach { partCol =>
-      if (!schemaCols.contains(partCol)) {
-        throw DeltaErrors.partitionColumnNotFoundException(partCol, schema)
-      }
+
+    val partitionsColsNotInSchema = partitionCols.toSet.diff(schemaCols).toSeq
```
allisonport-db, line 477: Should we be concerned about performance for tables with many many columns? I.e. we could do `schemaCols.toSet.subset(partitionCols)`

I'm thinking it's not at a scale where it's a big deal.

standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
```diff
@@ -472,33 +472,40 @@ abstract class DeltaLogSuiteBase extends FunSuite {
...
+
+        if (shouldThrow) {
+          val e = intercept[DeltaStandaloneException] {
+            log.startTransaction().updateMetadata(metadata)
+          }.getMessage
+
+          assert(
+            e.contains(s"Partition column(s) ${missingPartCols.mkString(",")} not found in schema"))
+        } else {
+          log.startTransaction().updateMetadata(metadata)
```
allisonport-db, line 506: nit: `.commit()` (previous test does this, I'm neutral on whether it's needed)

standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
```diff
@@ -472,33 +472,40 @@ abstract class DeltaLogSuiteBase extends FunSuite {
...
+    val schema = new StructType()
+      .add("part_1", new StringType())
+      .add("part_2", new LongType())
+      .add("foo", new IntegerType())
+      .add("bar", new BooleanType())
+
+    val correctPartitionsCols = Seq("part_1", "part_2")
+
+    Seq(
+      (correctPartitionsCols, Nil), // equal to
```
allisonport-db, line 484: what is this equal to?... isn't the schema columns [part_1, part_2, foo, bar]?

I'm not sure how this is different than the following test.

scottsand-db: My intention was that `part_1` and `part_2` are the partition columns, hence `correctPartitionsCols`.

So we have cases like
- correctPartitionsCols ==> which is "equal to" correctPartitionsCols
- `Seq("part_1")` ==> which is a subset of correctPartitionsCols

I'll make this more explicit.

standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
scottsand-db: it's not needed. we verify the metadata in `updateMetadata`

standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
allisonport-db: But isn't those being the correctPartitionsCols just arbitrary? Like for `(Seq("part_1"), Nil)` why isn't `correctPartitionCols=Seq("part_1")`

standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
allisonport-db: yeah I was just thinking we don't externally "know" this / what if it was changed in the future

scottsand-db: if it was changed in the future, then the test will fail, which is what we want. then we know we need to update the code or fix the test.

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
scottsand-db: do you mean, "calculate the `diff` only if `.subset` is false"?

sorry, little confused because a) schemaCols is already a set, b) `.subset` method doesnt exist, and c) if you mean `.subsetOf`, we want "is partitionsCols a subset of schemaCol", not the other way around.

scottsand-db: but ya, all these set operations are happening in-memory with a few thousand columns at most. I'm not concerned about performance here.

allisonport-db: > do you mean, "calculate the `diff` only if `.subset` is false"?

yep

Yeah I meant other way around with `subsetOf`. I agree it doesn't matter :)

PR #354, opened 2022-05-31
[delta-standalone] Implement actions iterator from `VersionLog`

6a0juu:
Currently VersionLog.getActions returns a list of actions. We would like Iterator to avoid memory pressure.


standalone/src/main/java/io/delta/standalone/VersionLog.java
```diff
@@ -51,4 +56,29 @@ public long getVersion() {
...
     }
+
+    /**
+     * @return an {@code CloseableIterator} of the actions for this table version
+     */
+    @Nonnull
+    public CloseableIterator<Action> getActionsIterator() {
+
+        // reset the iterator
+        actionIterator = actions.iterator();
```
scottsand-db, line 67: why not move this `actionIterator` variable into the `new CloseableIterator<Action>() {` instance below?

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
```diff
@@ -104,7 +104,7 @@ private[internal] class DeltaLogImpl private(
 
   override def getChanges(
       startVersion: Long,
-      failOnDataLoss: Boolean): java.util.Iterator[VersionLog] = {
+      failOnDataLoss: Boolean): java.util.Iterator[MemoryOptimizedVersionLog] = {
```
scottsand-db, line 107: nit: you can keep this as VersionLog.

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,46 @@
+package io.delta.standalone.internal
```
scottsand-db, line 1: nit: copyright

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,46 @@
...
+import scala.jdk.CollectionConverters.seqAsJavaListConverter
+
+import io.delta.standalone.actions.{Action => ActionJ}
+import io.delta.standalone.data.CloseableIterator
+
+import io.delta.standalone.internal.actions.Action
+import io.delta.standalone.internal.util.ConversionUtils
+import io.delta.standalone.internal.util.Implicits.CloseableIteratorOps
+
+private[internal] class MemoryOptimizedVersionLog(
```
scottsand-db, line 12: nit: classdoc. what does this class do? what makes it special?

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,46 @@
...
+    version: Long,
+    supplier: () => CloseableIterator[String])
+  extends io.delta.standalone.VersionLog(version, java.util.List[ActionJ]) {
+
+  private def getNewActionIterator(stringIterator: CloseableIterator[String]) =
+    new CloseableIterator[ActionJ]() {
+
+      override def next(): ActionJ = {
+        ConversionUtils.convertAction(
+          io.delta.standalone.internal.actions.Action.fromJson(stringIterator.next))
```
scottsand-db, line 22: below you use `Action.fromJson` but here you are using `io.delta.standalone.internal.actions.Action.fromJson`. why the difference?

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,46 @@
...
+        stringIterator.close()
+      }
+
+      override def hasNext: Boolean = {
+        stringIterator.hasNext
+      }
+    }
+
+  override def getActionsIterator: CloseableIterator[ActionJ] = {
+    getNewActionIterator(supplier())
```
scottsand-db, line 36: nit: you can just move the contents of `getNewActionIterator` here. no need to define an extra private method for it. why make me look here `getActionsIterator` only to have to then go and read another function `getNewActionIterator`?

scottsand-db (changes requested):
looks great! looking forward to seeing some tests. thanks for doing this!

scottsand-db (changes requested):
Thanks for adding tests! Looking good. Left some comments.

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,67 @@
...
+import io.delta.standalone.internal.util.Implicits.CloseableIteratorOps
+
+/**
+ * Scala implementation of Java class [[VersionLog]].
+ * To save memory, full action list is loaded only when calling [[getActions]]
+ */
+private[internal] class MemoryOptimizedVersionLog(
+    version: Long,
+    supplier: () => CloseableIterator[String])
+  extends io.delta.standalone.VersionLog(version, new java.util.ArrayList[ActionJ]()) {
```
scottsand-db, line 36: you can just use `VersionLog` not `io.delta.standalone.VersionLog`

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
```diff
@@ -123,13 +122,9 @@ private[internal] class DeltaLogImpl private(
...
 
-      new VersionLog(
+      new MemoryOptimizedVersionLog(
         version,
-        store.read(p, hadoopConf)
-          .toArray
-          .map(x => ConversionUtils.convertAction(Action.fromJson(x)))
-          .toList
-          .asJava)
+        () => store.read(p, hadoopConf)).asInstanceOf[VersionLog]
```
scottsand-db, line 127: you don't need to do `.asInstanceOf[VersionLog]`.
`MemoryOptimizedVersionLog` _is a VersionLog_. It is a child of the parent.

scottsand-db: but I told you earlier to keep the same interface. hm. let me think.

standalone/src/main/java/io/delta/standalone/VersionLog.java
6a0juu: `actionIterator` is moved into [the instance](https://github.com/delta-io/connectors/pull/354/files#diff-d6ec31e5e8e28e2b824601cc142ee00166ba3a37bec5b2520eacbb1e850daef7R69)

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
6a0juu: This change reverted and added a type transformation at [here](https://github.com/delta-io/connectors/pull/354/files#diff-0ebc696377a1d97f1e3b082548ba108df4107da1429f0436ae2a84c4aba7c0f7R128)

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
6a0juu: Added [here](https://github.com/delta-io/connectors/pull/354/files#diff-c96de1028503f1efa5245b83b0c25240d4eb866edc7778a9ee70e929a4ff06a2R1)

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
6a0juu: Class doc and method doc added, start from [here](https://github.com/delta-io/connectors/pull/354/files#diff-c96de1028503f1efa5245b83b0c25240d4eb866edc7778a9ee70e929a4ff06a2R28)

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
6a0juu: No difference, I wrote the full name because of some legacy dependency naming problems. Fixed [here](https://github.com/delta-io/connectors/pull/354/files#diff-c96de1028503f1efa5245b83b0c25240d4eb866edc7778a9ee70e929a4ff06a2R48)

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
6a0juu: Made `getNewActionIterator` anonymous and move to [here](https://github.com/delta-io/connectors/pull/354/files#diff-c96de1028503f1efa5245b83b0c25240d4eb866edc7778a9ee70e929a4ff06a2R39)

standalone/src/main/java/io/delta/standalone/VersionLog.java
```diff
@@ -16,20 +16,23 @@
...
 
 /**
  * {@link VersionLog} is the representation of all actions (changes) to the Delta Table
  * at a specific table version.
  */
-public final class VersionLog {
+public class VersionLog {
     private final long version;
 
-    @Nonnull
```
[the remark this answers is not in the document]
6a0juu: I wonder if a null list in a `MemoryOptimizedVersionLog` instance will violate this annotation?

standalone/src/main/java/io/delta/standalone/VersionLog.java
```diff
@@ -16,20 +16,23 @@
 
 package io.delta.standalone;
 
+import java.io.IOException;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
 import javax.annotation.Nonnull;
 
+import com.fasterxml.jackson.databind.JsonNode;
```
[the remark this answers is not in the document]
6a0juu: Removed. This is an auto-filling typo

standalone/src/main/java/io/delta/standalone/VersionLog.java
```diff
@@ -51,4 +54,33 @@ public long getVersion() {
...
+
+    /**
+     * @return an {@code CloseableIterator} of the actions for this table version
+     */
+    @Nonnull
+    public CloseableIterator<Action> getActionsIterator() {
+
+        return new CloseableIterator<Action>() {
+            /**
+             * A wrapper class transforming {@code CloseableIterator<String>}
```
scottsand-db, line 66: where is the `CloseableIterator<String>`? `actionIterator` is an `Iterator<Action>`.
I think you can delete this comment.
Also, this inline class is pretty trivial. You don't need this comment anyways

standalone/src/main/java/io/delta/standalone/VersionLog.java
```diff
@@ -51,4 +54,33 @@ public long getVersion() {
     public List<Action> getActions() {
         return Collections.unmodifiableList(actions);
     }
+
+    /**
+     * @return an {@code CloseableIterator} of the actions for this table version
+     */
+    @Nonnull
+    public CloseableIterator<Action> getActionsIterator() {
+
```
scottsand-db, line 63: delete blank line

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
```diff
@@ -123,13 +122,10 @@ private[internal] class DeltaLogImpl private(
...
-      new VersionLog(
+      new MemoryOptimizedVersionLog(
         version,
-        store.read(p, hadoopConf)
-          .toArray
-          .map(x => ConversionUtils.convertAction(Action.fromJson(x)))
-          .toList
-          .asJava)
+        () => store.read(p, hadoopConf))
+        .asInstanceOf[VersionLog]
```
scottsand-db, line 128: another option is that we change DeltaLog.java API for `getChanges` from `Iterator<VersionLog>` to `Iterator<? extends VersionLog>`.

cc @vkorukanti .

in this case, we wouldn't need to use `.asInstanceOf`

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
scottsand-db: btw @6a0juu please resolve comments once you have fixed the code.

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,73 @@
...
+
+  /** @return an [[CloseableIterator]] of [[Action]] for this table version */
+  override def getActionsIterator: CloseableIterator[ActionJ] = new CloseableIterator[ActionJ]() {
+    /**
+     * A wrapper class transforming [[CloseableIterator]] of [[String]]
```
scottsand-db, line 40: nit: this comment should use `//` not `/**`

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
```diff
@@ -0,0 +1,143 @@
...
+      defaultVersionNumber,
+      supplierWithCounter
+    )
+
+    assert(versionLogWithIterator.getVersion == defaultVersionNumber,
+      s"versionLog.getVersion() should be $defaultVersionNumber other than " +
+        s"${versionLogWithIterator.getVersion}")
+
+    /**
+     * Calling counter increased only when a new [[CloseableIterator]] is instantiated. i.e.
```
scottsand-db, line 133: nit: this comment should use // not /**

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
scottsand-db: also describe the parameters. what is `version`, what is `supplier`.

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
```diff
@@ -0,0 +1,143 @@
...
+
+  test("basic operation for MemoryOptimizedVersionLog.scala") {
+
+    checkVersionLog(new MemoryOptimizedVersionLog(
+      defaultVersionNumber,
+      () => stringCloseableIterator
+    ))
+  }
+
+  test("CloseableIterator should not be instantiated when supplier is not used ") {
```
scottsand-db, line 121: nit: extra space at end of test name

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
```diff
@@ -0,0 +1,143 @@
...
+
+    val versionLogWithIterator = new MemoryOptimizedVersionLog(
+      defaultVersionNumber,
+      supplierWithCounter
```
scottsand-db, line 125: `supplierWithCounter` is only ever used here. so why is it a class-level variable? why increase the scope of it?

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
```diff
@@ -0,0 +1,143 @@
...
+      s"versionLog.getVersion() should be $defaultVersionNumber other than " +
+        s"${newVersionLog.getVersion}")
+    assert(newActionList.size() == actionList.size())
+    assert(newActionList
+      .toArray()
+      .zip(actionList.toArray())
+      .count(x => x._1 == x._2) == newActionList.size())
+
+    stringIterator = stringList.iterator
+    anotherStringIterator = stringList.iterator
```
scottsand-db, line 96: this is still a bad code smell IMO. you are updating a variable here .. during a test helper ... which mean the state of this class-level variable is dependent on when tests are being run.

I would try and find a way to re-write these tests without these dependencies / coupling between different tests and class-level variables.

zsxwing:
A high level suggestion: instead of adding a new subclass of VersionLog, can we make VersionLog smart enough like this?
```java
public final class VersionLog {
    private final long version;

    private List<Action> actions;

    private Supplier<CloseableIterator<Action>> actionsSupplier;

    VersionLog(long version, @Nonnull Supplier<CloseableIterator<Action>> actionsSupplier) {
        this.version = version;
        this.actionsSupplier = actionsSupplier;
        this.actions = null;
    }

    public VersionLog(long version, @Nonnull List<Action> actions) {
        this.version = version;
        this.actionsSupplier = null;
        this.actions = actions;
    }

    /**
     * @return the table version at which these actions occurred
     */
    public long getVersion() {
        return version;
    }

    /**
     * @return an unmodifiable {@code List} of the actions for this table version
     */
    @Nonnull
    public List<Action> getActions() {
        synchronized(this) {
            if (actions == null) {
                List<Action> localActions = new ArrayList<>();
                try(CloseableIterator<Action> iter = getActionsIterator()) {
                    while (iter.hasNext()) {
                        localActions.add(iter.next());
                    }
                    actions = localActions;
                } catch (IOException e) {
                    throw new UncheckedIOException(e);
                }
            }
        }
        return Collections.unmodifiableList(actions);
    }

    @Nonnull
    public CloseableIterator<Action> getActionsIterator() {
        synchronized (this) {
            if (actionsSupplier != null) {
                return actionsSupplier.get();
            } else {
                // create a `CloseableIterator` from `actions`
                return null;
            }
        }
    }
}
```
Benefits of this approach:
- If an existing user calls `getActions` multiple times today, they will not see any obvious difference.
- Not need to add a subclass.

standalone/src/main/java/io/delta/standalone/VersionLog.java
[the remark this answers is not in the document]
6a0juu: Oh it is an empty list. Added it back.

standalone/src/main/java/io/delta/standalone/VersionLog.java
6a0juu: Deleted

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
6a0juu: Fixed

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
6a0juu: Fixed

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
```diff
@@ -0,0 +1,117 @@
...
+
+    @throws[java.io.IOException]
+    override def close(): Unit = {}
+
+    override def hasNext: Boolean = {
+      stringIterator.hasNext
+    }
+  }
+
+  private def checkVersionLog(newVersionLog: VersionLog): Unit = {
```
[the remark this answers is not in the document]
6a0juu: Fixed [here](https://github.com/delta-io/connectors/pull/354/files#diff-31c34f77bca032d4c2f1ac05d83ac401f06ef496723b34ffc31543c2b53b87eaR99)

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
6a0juu: Fixed

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
6a0juu: Fixed

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
6a0juu: Moved `applyCounter` and `supplierWithCounter` into the test

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,67 @@
...
+    override def hasNext: Boolean = {
+      stringIterator.hasNext
+    }
+  }
+
+  /** @return an [[List]] of [[Action]] for this table version */
+  override def getActions: java.util.List[ActionJ] = supplier()
+    .toArray
+    .map(x => ConversionUtils.convertAction(Action.fromJson(x)))
+    .toList
```
scottsand-db: you could save this result to a private variable, and then return Collections.unmodifiableList(_private_list). this is more performant.

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,73 @@
...
+      stringIterator.close()
+    }
+
+    override def hasNext: Boolean = {
+      stringIterator.hasNext
+    }
+  }
+
+  /** @return an [[List]] of [[Action]] for this table version */
+  override def getActions: java.util.List[ActionJ] = {
```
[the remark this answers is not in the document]
6a0juu: Fixed

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
```diff
@@ -0,0 +1,117 @@
...
+
+    val newActionList = newVersionLog.getActions
+
+    assert(newVersionLog.getVersion == defaultVersionNumber,
+      s"versionLog.getVersion() should be ${defaultVersionNumber} other than " +
```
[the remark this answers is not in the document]
6a0juu: Clue deleted

standalone/src/main/java/io/delta/standalone/VersionLog.java
```diff
@@ -51,4 +54,33 @@ public long getVersion() {
...
+        return new CloseableIterator<Action>() {
+            /**
+             * A wrapper class transforming {@code CloseableIterator<String>}
+             * to {@code CloseableIterator<Action>}
+             */
+
+            final Iterator<Action> actionIterator = actions.iterator();
```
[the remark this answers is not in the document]
6a0juu: Fixed

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
```diff
@@ -0,0 +1,117 @@
...
+      defaultVersionNumber,
+      () => actionCloseableIterator
+    )
+    checkVersionLog(versionLogWithIterator)
+  }
+
+  test("CloseableIterator should not be instantiated when supplier is not used ") {
+    val newFunction1: () => CloseableIterator[String] =
+      new (() => CloseableIterator[String]) {
+        var applyCounter: Int = 0
```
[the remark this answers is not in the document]
6a0juu: Fixed

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
6a0juu: [New doc] added

6a0juu:
Closed this PR. Made new PR based on @zsxwing implementation [here](https://github.com/delta-io/connectors/pull/363)

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
6a0juu: [Here](https://github.com/6a0juu/connectors/runs/6732843754?check_suite_focus=true#step:8:218) I met an error with generic type casting while using `Iterator<? extends VersionLog>`. Found [some reason](https://users.scala-lang.org/t/return-generic-type-and-upcast/4356/4) for this. Here are 3 solutions:
- Convert all `VersionLog` to `MemoryOptimizedVersionLog` in Flink test.
- Still use `.asInstanceOf`.
- Make an interface over 2 implementations.

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,93 @@
...
+ * @param version the table version at which these actions occurred
+ * @param supplier provide [[CloseableIterator]] of actions for fetching information inside all
+ *                 [[Action]] stored in this table version
+ */
+private[internal] class MemoryOptimizedVersionLog(
+    version: Long,
+    supplier: () => CloseableIterator[String])
+  extends VersionLog(version, new java.util.ArrayList[ActionJ]()) {
+
+  private var _private_action_list: Option[java.util.List[ActionJ]] = None
```
zsxwing, line 50: We can use Scala lazy to simplify the code
```suggestion
  private lazy val cachedActions: java.util.List[ActionJ] = {
    // put the converting code here.
  }
```

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
6a0juu: Made some exception re-throwing efforts [here](https://github.com/delta-io/connectors/pull/354/files#diff-c96de1028503f1efa5245b83b0c25240d4eb866edc7778a9ee70e929a4ff06a2R61)

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
zsxwing: We can change `deltaPaths.map { status =>` to `deltaPaths.map[VersionLog]` to tell Scala to pick up the parent class.

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
6a0juu: [Updated](https://github.com/delta-io/connectors/pull/354/files#diff-c96de1028503f1efa5245b83b0c25240d4eb866edc7778a9ee70e929a4ff06a2R51)

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,90 @@
...
+ * @param supplier provide [[CloseableIterator]] of actions for fetching information inside all
+ *                 [[Action]] stored in this table version
+ */
+private[internal] class MemoryOptimizedVersionLog(
+    version: Long,
+    supplier: () => CloseableIterator[String])
+  extends VersionLog(version, new java.util.ArrayList[ActionJ]()) {
+  import io.delta.standalone.internal.util.Implicits._
+
+  private lazy val _private_action_list: java.util.List[ActionJ] = {
```
zsxwing, line 51: nit: We use `camel case`. This can be called `cachedActions`.

6a0juu: [Updated](https://github.com/delta-io/connectors/pull/354/files#diff-c96de1028503f1efa5245b83b0c25240d4eb866edc7778a9ee70e929a4ff06a2R51)

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
6a0juu: Change the variables to functions [here](https://github.com/delta-io/connectors/pull/354/commits/650396808340bf3dedb295dfd29f6b91fc4f364e#diff-31c34f77bca032d4c2f1ac05d83ac401f06ef496723b34ffc31543c2b53b87eaR41) so we can make a new object for every test. No need to reset iterators during the test now.

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
6a0juu: > We can change `deltaPaths.map { status =>` to `deltaPaths.map[VersionLog]` to tell Scala to pick up the parent class.

[Updated](https://github.com/delta-io/connectors/pull/354/commits/46dda69d7fdc92b53d1a149b92834fb1daa6259a#diff-0ebc696377a1d97f1e3b082548ba108df4107da1429f0436ae2a84c4aba7c0f7R117)

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,90 @@
...
+  /** @return an [[CloseableIterator]] of [[Action]] for this table version */
+  override def getActionsIterator: CloseableIterator[ActionJ] = this.synchronized {
+    new CloseableIterator[ActionJ]() {
+      // A wrapper class transforming CloseableIterator[String] to CloseableIterator[Action]
+
+      private val stringIterator = supplier.apply()
+
+      @throws[java.io.IOException]
+      override def next(): ActionJ = {
+        Try(ConversionUtils.convertAction(Action.fromJson(stringIterator.next))).recover {
```
scottsand-db, line 68: why was this needed?

6a0juu: This change is based on [Ryan's comment](https://github.com/delta-io/connectors/pull/354#issuecomment-1145050152) ensuring thread-safe. I am not sure the call stack behind `.next()` method but it probably will raise error when multiple threads are consuming one `stringIterator`?

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,90 @@
...
+      }
+
+      override def hasNext: Boolean = {
+        stringIterator.hasNext
+      }
+    }
+  }
+
+  /** @return an unmodifiable [[List]] of [[Action]] for this table version */
+  override def getActions: java.util.List[ActionJ] = this.synchronized {
```
zsxwing, line 85: Remove `this.synchronized`. `lazy val` is thread safe.

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,90 @@
...
+    // CloseableIterator is automatically closed by
+    // io.delta.standalone.internal.util.Implicits.CloseableIteratorOps.toArray
```
zsxwing, line 87: nit: remove unused comments

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,90 @@
...
+        stringIterator.close()
+      }
+
+      override def hasNext: Boolean = {
+        stringIterator.hasNext
+      }
+    }
+  }
+
+  /** @return an unmodifiable [[List]] of [[Action]] for this table version */
```
zsxwing, line 84: nit: remove this. We don't need to copy the comment from the parent class

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,90 @@
...
+      private val stringIterator = supplier.apply()
+
+      @throws[java.io.IOException]
+      override def next(): ActionJ = {
+        Try(ConversionUtils.convertAction(Action.fromJson(stringIterator.next))).recover {
+          case e: java.io.IOException => throw new UncheckedIOException(e)
+        }.get
+      }
+
+      @throws[java.io.IOException]
```
zsxwing, line 73: nit: this is not needed as we are in Scala

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,90 @@
...
+      .toList
+      .asJava
+  }
+
+  /** @return an [[CloseableIterator]] of [[Action]] for this table version */
+  override def getActionsIterator: CloseableIterator[ActionJ] = this.synchronized {
+    new CloseableIterator[ActionJ]() {
+      // A wrapper class transforming CloseableIterator[String] to CloseableIterator[Action]
+
+      private val stringIterator = supplier.apply()
```
zsxwing, line 64: nit: 
```suggestion
      private val stringIterator = supplier()
```

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,90 @@
...
+
+      @throws[java.io.IOException]
```
zsxwing, line 66: nit: remove this as it's not needed in Scala.

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
zsxwing: The returned Iterator is okay to not be thread-safe. My comment is mostly for `getActions` which we are going to cache `actions`. Scala `lazy val` would provide the guarantee and we don't need to do extra work.

This method can be as simple as you did before
```
ConversionUtils.convertAction(Action.fromJson(stringIterator.next()))
```

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,90 @@
...
+  private lazy val cachedActions: java.util.List[ActionJ] = {
+    supplier()
+      .toArray
+      .map(x => ConversionUtils.convertAction(Action.fromJson(x)))
+      .toList
+      .asJava
+  }
+
+  /** @return an [[CloseableIterator]] of [[Action]] for this table version */
+  override def getActionsIterator: CloseableIterator[ActionJ] = this.synchronized {
```
zsxwing, line 60: Remove `this.synchronized`.

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,90 @@
...
+
+  private lazy val cachedActions: java.util.List[ActionJ] = {
+    supplier()
+      .toArray
+      .map(x => ConversionUtils.convertAction(Action.fromJson(x)))
+      .toList
+      .asJava
+  }
+
+  /** @return an [[CloseableIterator]] of [[Action]] for this table version */
```
zsxwing, line 59: nit: remove this. We don't need to copy the comment from the parent class

zsxwing (commented):
Overall looks good. Left a few nits. Mostly style issues.

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
6a0juu: Just curious, what would happen if the called methods throws a java exception to Scala?

zsxwing (approved):

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
```diff
@@ -0,0 +1,135 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.delta.standalone.internal
+
+import io.delta.storage.CloseableIterator
+import org.scalatest.FunSuite
+import scala.collection.JavaConverters._
```
scottsand-db, line 21: wrong import order

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,80 @@
...
+    version: Long,
+    supplier: () => CloseableIterator[String])
+  extends VersionLog(version, new java.util.ArrayList[ActionJ]()) {
+  import io.delta.standalone.internal.util.Implicits._
+
+  private lazy val cachedActions: java.util.List[ActionJ] = {
+    supplier()
+      .toArray
+      .map(x => ConversionUtils.convertAction(Action.fromJson(x)))
+      .toList
```
scottsand-db, line 53: you've lost the comment that says how this is automatically closed.

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
```diff
@@ -0,0 +1,135 @@
...
+  private def checkVersionLog(newVersionLog: VersionLog): Unit = {
+
+    val newActionList = newVersionLog.getActions
+
+    assert(newVersionLog.getVersion == defaultVersionNumber)
+    assert(newActionList.size() == actionList.size())
+    assert(newActionList
+      .toArray()
+      .zip(actionList.toArray())
+      .count(x => x._1 == x._2) == newActionList.size())
```
scottsand-db, line 85: what is this testing?

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
```diff
@@ -0,0 +1,135 @@
...
+    val newActionIterator = newVersionLog.getActionsIterator
+
+    (1 to listLength).foreach( _ => {
+      assert(newActionIterator.hasNext && actionCloseableIterator.hasNext)
+      assert(newActionIterator.next() == actionCloseableIterator.next())
+    })
+  }
+
+  test("basic operation for VersionLog.java") {
+
```
scottsand-db, line 96: nit: delete empty link

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
```diff
@@ -0,0 +1,135 @@
...
+  test("basic operation for MemoryOptimizedVersionLog.scala") {
+
+    checkVersionLog(new MemoryOptimizedVersionLog(
+      defaultVersionNumber,
+      () => stringCloseableIterator
+    ))
+  }
+
+  test("CloseableIterator should not be instantiated when supplier is not used") {
+
```
scottsand-db, line 112: nit: delete empty link

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
scottsand-db: isn't that what `assert(newActionIterator.next() == actionCloseableIterator.next())` is checking below?

furthermore, instead of having each AddFile being the same (i.e. same path) `AddFile(1.toString, Map.empty, 1, 1, dataChange = true).json)`, each AddFile could have a unique path.

so during the `(1 to listLength)` call, you can also compare the paths

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
6a0juu: Deleted

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,80 @@
...
+import io.delta.standalone.VersionLog
+import io.delta.standalone.actions.{Action => ActionJ}
+
+import io.delta.standalone.internal.actions.Action
+import io.delta.standalone.internal.util.ConversionUtils
+
+
+/**
+ * Scala implementation of Java class [[VersionLog]].
+ * To save memory, full action list is loaded only when calling [[getActions]].
```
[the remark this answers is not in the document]
6a0juu: Updated

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
6a0juu: Fixed

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
```diff
@@ -0,0 +1,135 @@
...
+import io.delta.standalone.internal.actions.{Action, AddFile}
+import io.delta.standalone.internal.util.ConversionUtils
+
+class VersionLogSuite extends FunSuite {
+
+  private val defaultVersionNumber = 33
+  private val listLength = 300
+  private val stringList: List[String] = List.fill(listLength)(
+    AddFile(1.toString, Map.empty, 1, 1, dataChange = true).json)
+  private val actionList: java.util.List[ActionJ] = stringList
```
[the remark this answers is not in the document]
scottsand-db: sorry, how is this "resolved"? It is still a modifiable list.

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
6a0juu: `AddFile` with different path added [here](https://github.com/delta-io/connectors/pull/354/files#diff-31c34f77bca032d4c2f1ac05d83ac401f06ef496723b34ffc31543c2b53b87eaR36). And the comparison between action iterators is now [decided by path](https://github.com/delta-io/connectors/pull/354/files#diff-31c34f77bca032d4c2f1ac05d83ac401f06ef496723b34ffc31543c2b53b87eaR101).

And code is reformatted by `scalafmt`.

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
```diff
@@ -0,0 +1,150 @@
...
+
+import io.delta.standalone.internal.actions.{Action, AddFile}
+import io.delta.standalone.internal.util.ConversionUtils
+
+class VersionLogSuite extends FunSuite {
+  private val defaultVersionNumber = 33
+  private val listLength = 10
+  private val stringList: java.util.List[String] = Collections.unmodifiableList(
+    List
+      .tabulate(listLength)(x =>
```
scottsand-db, line 37: using `tabulate` is fine, but most often in this code base we do `(0 until X).map { x => ... }` just an FYI.

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,82 @@
...
+import io.delta.standalone.internal.actions.Action
+import io.delta.standalone.internal.util.ConversionUtils
+
+/**
+ * Scala implementation of Java class [[VersionLog]], provides a way to iterate through actions
+ * without loading the entire actions into memory when [[getActionsIterator]] is used.
+ *
+ * To save memory, full action list is loaded only when calling [[getActions]].
+ * [[CloseableIterator]] of actions, instead of [[List]] of actions, is passed into the class and
+ * the full action list is only instantiated when calling [[getActions]]. The memory occupied by
```
scottsand-db, line 37: duplicate information in the javadoc

above you say `full action list is loaded only when calling [[getActions]].`. Here you repeat that info again, `the full action list is only instantiated when calling [[getActions]]`

6a0juu: Removed duplicated description.

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,81 @@
...
+
+import io.delta.standalone.internal.actions.Action
+import io.delta.standalone.internal.util.ConversionUtils
+
+/**
+ * Scala implementation of Java class [[VersionLog]], provides a way to iterate through actions
+ * without loading the entire actions into memory when [[getActionsIterator]] is used.
+ *
+ * [[CloseableIterator]] of actions, instead of [[List]] of actions, is passed into the class and
+ * the full action list is only instantiated when calling [[getActions]]. The memory occupied by
```
scottsand-db, line 36: `The memory occupied by action list is saved here when action list is long.`

We won't do any sort of check on the length of the action list, do we? Is this sentence correct?

standalone/src/test/scala/io/delta/standalone/internal/VersionLogSuite.scala
```diff
@@ -33,8 +33,8 @@ class VersionLogSuite extends FunSuite {
   private val defaultVersionNumber = 33
   private val listLength = 10
   private val stringList: java.util.List[String] = Collections.unmodifiableList(
-    (0 until listLength).map{
-      x => AddFile(x.toString, Map.empty, 1, 1, dataChange = true).json
+    (0 until listLength).map{ x =>
```
scottsand-db, line 36: nit: `.map { x =>`

6a0juu: Fixed it in the [next commit](https://github.com/delta-io/connectors/pull/354/commits/297c7cf14bcdebe7f03586cf2e3ca54f3d11c26f)

scottsand-db: Are you use? there needs to be a space after `.map`
correct: `.map { x =>`
incorrect: `.map{x =>`

6a0juu: Extra space added

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
6a0juu: We don't check the length. Class doc changed to `The memory is saved when [[getActions]] not been called`

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,81 @@
...
+ * action list is saved when [[getActions]] not been called.
```
scottsand-db, line 37: is this sentence correct, too?

`The memory occupied by action list is saved when [[getActions]] not been called.`

I think "saved" is a confusing word here. Are we "saving/caching?" something? Or are we _not_ caching something and thus "saving/consuming" less memory?

6a0juu: Rewritten the detailed class description

standalone/src/main/scala/io/delta/standalone/internal/MemoryOptimizedVersionLog.scala
```diff
@@ -0,0 +1,81 @@
...
+import io.delta.standalone.internal.actions.Action
+import io.delta.standalone.internal.util.ConversionUtils
+
+/**
+ * Scala implementation of Java class [[VersionLog]], provides a way to iterate through actions
+ * without loading the entire actions into memory when [[getActionsIterator]] is used.
+ *
+ * [[MemoryOptimizedVersionLog]] is constructed by [[CloseableIterator]] of [[Action]]. The full
+ * [[List]] of [[Action]] is not instantiated until [[getActions]] is called, which uses less
+ * memory than [[VersionLog]] with disabled [[getActions]].
```
scottsand-db, line 37: I don't know what `disabled [[getActions]]` means. how is a method disabled?
feel free to run your next comment by me via Slack

scottsand-db (approved):
LGTM

PR #366, opened 2022-06-06
[delta-standalone] Add `DataType::fromJson` API

scottsand-db:
This PR adds a `DataType fromJson(String json)` API to the DataType class.

This is just a helpful, quality-of-life public API to have.

As well, the Flink/Delta Source Connector may want to, in the future, serialize and then deserialize the snapshot metadata schema, as a performance optimization.


allisonport-db (approved):
LGTM

PR #370, opened 2022-06-08
Update accepted hadoop configuration keys for setting LogStore class

allisonport-db:
- Update accepted LogStore _scheme_ conf keys to allow the "spark." prefix
- Update accepted LogStore _class_ conf keys
    - Keep legacy support for `io.delta.standalone.LOG_STORE_CLASS_KEY`
    - Main support is now `delta.logStore.class` and we also allow the "spark." prefix for this key

### Motivation

Support the same conf keys as we do in delta-spark so that settings can be copied between the two.


standalone/src/main/scala/io/delta/standalone/internal/sources/StandaloneHadoopConf.scala
```diff
@@ -24,6 +24,9 @@ private[internal] object StandaloneHadoopConf {
   /** Time zone as which time-based parquet values will be encoded and decoded. */
   val PARQUET_DATA_TIME_ZONE_ID = "io.delta.standalone.PARQUET_DATA_TIME_ZONE_ID"
 
-  /** Class name for the desired [[LogStore]] implementation to be used. */
-  val LOG_STORE_CLASS_KEY = "io.delta.standalone.LOG_STORE_CLASS_KEY"
+  /** Legacy key for the class name of the desired [[LogStore]] implementation to be used. */
+  val LEGACY_LOG_STORE_CLASS_KEY = "io.delta.standalone.LOG_STORE_CLASS_KEY"
+
+  /** Key for the class name of the desired [[LogStore]] implementation to be used. */
+  val LOG_STORE_CLASS_KEY = "delta.logStore.class"
```
allisonport-db, line 31: Should we move `def logStoreSchemeConfKey(scheme: String): String = s"delta.logStore.${scheme}.impl"` to here as well? Seems inconsistent.

scottsand-db (approved):
LGTM. Added a comment, a question, and asked for 1 minor test change.

PR #377, opened 2022-06-16
[Doc/Flink] Add build configuration

kination:
Hello,
I've just add configuration for gradle build.

Please check 🙏

### DIFF
--- a/flink/README.md
+++ b/flink/README.md
@@ -109,6 +109,30 @@ Scala 2.12:
 </project>
 ```
 
+#### Gradle
+
+```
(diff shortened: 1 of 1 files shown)


scottsand-db:
Hi @kination, as of now we don't plan on supporting Gradle, nor including any documentation for it. The more build tools we have to support, the harder releases and dependency management will likely become.

kination:
@scottsand-db oh, got it.
I've just added because I was working on my side with gradle. If it blocks the progress, you could ignore it.

PR #384, opened 2022-06-30
Update flink `README.md`

scottsand-db:
- add table of contents
- remove java boilerplate from flink examples
- add two flink source examples, one for bounded + time travel, one for continuous + time travel
- have distinct flink sink and flink source sections. each section has its own examples
- misc fixes (e.g. make the known limitation only for 0.4.x, indentation, etc).

### DIFF
--- a/flink/README.md
+++ b/flink/README.md
@@ -171,8 +171,6 @@ In this example we show how to create a `DeltaSink` and plug it to an
 existing `org.apache.flink.streaming.api.datastream.DataStream`.
 
 ```java
-package com.example;
-
 import io.delta.flink.sink.DeltaSink;
 import org.apache.flink.core.fs.Path;
 import org.apache.flink.streaming.api.datastream.DataStream;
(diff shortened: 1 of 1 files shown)


flink/README.md
```diff
@@ -21,53 +38,236 @@ Depending on the version of the connector you can use it with following Apache F
 |  0.4.x (Sink Only)  |    >= 1.12.0    |
 |        0.5.x        |    >= 1.13.0    |
```
kristoffSC, line 39: 0.5.0 would be
`1.13.0 <= X <=1.14.5`

scottsand-db: are there any restrictions on the sink? does it work with flink 1.15 ?

kristoffSC: I would have to check this, because I was using a job with source and sink.
However for sure there are some issues with the connector with this version.

The 1.14.x branch seems ok for both source and sink though.

kristoffSC: @scottsand-db Ive checked a Sink only job with Flink 1.15 and it fails with 
`Caused by: java.lang.NoSuchMethodError: 'org.apache.flink.metrics.MetricGroup org.apache.flink.api.connector.sink.Sink$InitContext.metricGroup()'`

I would say its an issue with incompatible API in Flink 1.15

PR #388, opened 2022-07-08
Checkpoint write should use DeltaLog.hadoopConf

zsxwing:
We should pass DeltaLog.hadoopConf to parquet writer so that it will use the correct hadoop conf.


scottsand-db (approved):
LGTM.

Issue #371, opened 2022-06-09
Flink DeltaSink: Add stats to AddFile actions.


gopik:
The implementation passes null for the stats field for AddFile action. Adding stats to the log would enable optimization of queries on the delta table as well as help upsert implementation.

cc: @dennyglee

gopik:
The stats can either be computed while writeElement call (like the way record count is maintained) or it can be computed after parquet file is written (for example in toAddFile in DeltaPendingFile). 

The file is visible only after writer is committed. If the file can be accessed (at its temp path) before committed, it can be computed as soon as the in progress writer is closed.

Otherwise, It can either be computed in DeltaCommitter soon after the file is committed (and visible at the targetPath) or in DeltaGlobalCommitter as part of toAddFile callback. In case of DeltaCommitter, this can be executed in parallel if there are multiple sink tasks. In case of global committer, all these would be executed in single thread.

I found that bucketPath visibility is not available everywhere (filename is available everywhere). Is there any specific reason to isolate the path to a few components?

scottsand-db:
> Is there any specific reason to isolate the path to a few components?

In general, we only want to expose internal fields to the components that need them. So, if your new features requires exposing this field to other classes, that's fine.

@gopik thanks for making this issue. Would you be interested in implementing your proposed solution? If so, want to start off with a short design doc, detailing the problem, requirements, and design decisions?

gopik:
@scottsand-db Sure, I'll work on the doc and update.

gopik:
https://docs.google.com/document/d/1GNujU7XpV2eG7OevFBSZeYKNRM-ZrUx7egb7ti5hAF4/edit#

gopik:
Reference impl: https://github.com/delta-io/delta-rs/blob/main/rust/src/writer/stats.rs

scottsand-db:
Hi @gopik - the decision doc still seems incomplete. Can you let me know when it is ready for review?

gopik:
@scottsand-db Is there a reusable impl for serializing/deserializing add file stats?

scottsand-db:
@gopik - add file stats are just a string, to be interpreted as JSON. isn't that sufficient?

scottsand-db:
> I've updated it now. Could you take a look again and add a comment on specifically what needs to be added?

Can you please include, for each design decision, which option is preferred? and the rationale behind it? and the pros/cons of each option.

[...]

PR #393, opened 2022-07-12
FlinkDeltaSink: Add column stats from the parquet file to delta transaction log

gopik:
Extracts the column level stats from parquet file and add it to the delta log in json format, fixes https://github.com/delta-io/connectors/issues/371

- The implementation is not complete as it may not work with all data types yet. Current implementation supports float/double/long/int/binary. String stats are captured as binary.

- Stats extraction needs to do a file io. Current implementation does this in global committer which is run in single thread. This implementation was done for simplicity, we may need to find an alternative to compute stats in parallel.


gopik:
@scottsand-db Anything else I need to do before this can be reviewed?

tdas:
@scottsand-db is out on vacation, hence the delay. I will try to take a look until he is back.

gopik:
@tdas, thanks for the update.

scottsand-db:
Hi @gopik - I'm back from vacation and will take a look this week. Thanks for your patience!

scottsand-db:
@kristoffSC - take a look?

scottsand-db:
Linking the design doc for future visibility: https://docs.google.com/document/d/1GNujU7XpV2eG7OevFBSZeYKNRM-ZrUx7egb7ti5hAF4/edit

flink/src/main/java/io/delta/flink/sink/internal/writer/stats/DeltaFileStats.java
```diff
@@ -0,0 +1,246 @@
...
+        return objectMapper.getNodeFactory().numberNode(stat.getMin());
+    }
+
+    @Override
+    public JsonNode getMax() {
+        return objectMapper.getNodeFactory().numberNode(stat.getMax());
+    }
+}
+
+class JsonStatInteger extends JsonStat {
```
scottsand-db, line 134: these couldn't be written using a template? perhaps my java is out of date

flink/src/main/java/io/delta/flink/sink/internal/writer/stats/DeltaFileStats.java
```diff
@@ -0,0 +1,246 @@
...
+    }
+}
+
+/**
+ * Converts Parquet stats to delta json format (and vice versa).
+ *
+ * The referenced columns are not delta logical columns but the physical columns within the parquet
+ * file.
+ */
+public class DeltaFileStats {
```
scottsand-db, line 192: can you include an example of what this looks like? e.g. for a table with schema
```
|-- a: struct
|    |-- b: struct
|    |    |-- c: long
|-- d: int
```

flink/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/DeltaPendingFile.java
```diff
@@ -112,20 +119,31 @@ public LinkedHashMap<String, String> getPartitionSpec() {
...
      *
+     * @param basePath Base path for the file.
      * @return {@link AddFile} object generated from input
      */
-    public AddFile toAddFile() {
+    public AddFile toAddFile(Path basePath, StructType schema) {
         LinkedHashMap<String, String> partitionSpec = this.getPartitionSpec();
+        String stats = null;
         long modificationTime = this.getLastUpdateTime();
         String filePath = PartitionPathUtils.generatePartitionPath(partitionSpec) +
```
scottsand-db, line 129: isn't `basePath` essentially the same as `PartitionPathUtils.generatePartitionPath(partitionSpec)`, considering that `filePath` = `PartitionPathUtils.generatePartitionPath(partitionSpec)` + `fileName`

flink/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/DeltaPendingFile.java
```diff
@@ -112,20 +119,31 @@ public LinkedHashMap<String, String> getPartitionSpec() {
...
             this.getFileName();
+        try {
+            ParquetFileStats parquetStats = ParquetFileStats.readStats(
```
scottsand-db, line 132: i'm curious what @kristoffSC thinks about perform a read on every write.

scottsand-db (changes requested):

flink/src/main/java/io/delta/flink/sink/internal/writer/stats/DeltaFileStats.java
gopik: It would be great if we can code this using generics. I tried the following but didn't work:

class JsonStatNumber<T> extends JsonStat {
    Statistics<T> stat
   JsonNode getMin() {
      JsonNode.numberNode(stat.getMin());
  }
}

numberNode impl in JsonNode has overload for each numeric type. But T is not constrained to the types that is allowed by numberNode overloads. There might be some other tricks though to make it compile.

Another approach I'm exploring is, instead of manual serialization to json, if we converted it to first Avro GenericRecord and then serialize it to json. The advantage of GenericRecord is, avro library has bunch of utilities on top of record. For example you get comparable implementation for free which useful for merging the files by primary key order.

flink/src/main/java/io/delta/flink/sink/internal/writer/stats/DeltaFileStats.java
gopik: Added this to the javadoc:

+ * For a table with following schema:
+ *
+ * |-- a: struct
+ * |    |-- b: struct
+ * |    |    |-- c: long
+ * |-- d: int
+ *
+ * parquet stats would be captured as:
+ * recordCount: 1000
+ * a.b.c, LongStatistics{min=10, max=2000, nullCount=0}
+ * d,     IntStatistics{min=5, max=5, nullCount=20}
+ *
+ * This implementation recreates the nested structure required for delta stats which then become:
+ * {
+ *   "numRecords": 1000,
+ *   "minValues": {
+ *     "a": {
+ *       "b": {
+ *         "c": 10
+ *       }
+ *     },
+ *     "d": 5
+ *   },
+ *   "maxValues": {
+ *     "a": {
+ *       "b": {
+ *         "c": 2000
+ *       }
+ *     },
+ *     "d": 5
+ *   },
+ *   "nullCounts": {
+ *     "a": {
+ *       "b": {
+ *         "c": 0
+ *       }
+ *     },
+ *     "d": 20
+ *   }
+ * }

flink/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/DeltaPendingFile.java
gopik: It is, but in the current implementation this is made in one place and passed to all the components using arguments.
Also, it doesn't have to be this if my understanding is correct. It could be any directory prefix if there are no partitions.

kristoffSC:
Hi @gopik 
Just a small question, 
do you have check style configured accordingly to instruction?
https://github.com/delta-io/connectors/tree/master/dev

I think couple things like missing empty lines ect. might be off.

flink/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/DeltaPendingFile.java
kristoffSC: yeah, 
I'm a little bit concern that we will do this from one thread -> Global Committer. 
I'm wondering how painful would be to open 100, 200 etc parquet files and read footer from every one.

Maybe we could have some benchmark that would compare before and after for setup like:
500 or 1k events/s,
10 or 100 writers,
checkpoint every 10 seconds
Test duration 2 minutes.
+ some requirements for columns (number, value range etc).

Also @scottsand-db I'm wondering how Delta gets the statistics? It compute them on a fly or from Parquet file?

flink/src/main/java/io/delta/flink/sink/internal/writer/stats/ParquetFileStats.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private final Map<ColumnPath, Statistics<?>> columnStats;
+    private long rowCount;
+
+    private ParquetFileStats(ParquetMetadata parquetMetadata) {
+        this.columnStats = load(parquetMetadata);
+    }
+
+    /**
+     * Factory method to read stats given path to parquet file.
```
scottsand-db, line 36: can you add some more high-level information about how you are reading the stats? full scan? just the footer? etc.

scottsand-db:
@kristoffSC - delta on spark computes the stats while writing out the parquet file.

flink/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/DeltaPendingFile.java
gopik: @kristoffSC @scottsand-db I've added details to "Read stats in global committer vs file committer" section in the [doc](https://docs.google.com/document/d/1GNujU7XpV2eG7OevFBSZeYKNRM-ZrUx7egb7ti5hAF4/edit#heading=h.lgucd4d560q9). Please take a look.

We can increase the parallelism by pre computing stats in DeltaFileCommitter so that global committer's toAddFile doesn't wait for a file read.

flink/src/main/java/io/delta/flink/sink/internal/writer/stats/ParquetFileStats.java
gopik: Added comment.

gopik:
> Hi @gopik Just a small question, do you have check style configured accordingly to instruction? https://github.com/delta-io/connectors/tree/master/dev
> 
> I think couple things like missing empty lines ect. might be off.

Yes, I've configured it accordingly. I also ran build/sbt flink/checkstyle and it finished successfully.

kristoffSC:
> > Hi @gopik Just a small question, do you have check style configured accordingly to instruction? https://github.com/delta-io/connectors/tree/master/dev
> > I think couple things like missing empty lines ect. might be off.
> 
> Yes, I've configured it accordingly. I also ran build/sbt flink/checkstyle and it finished successfully.

Yeah,
we have noticed that "checkstyle" does not catch all the things and formatter can mess things even more.

The thing I was referring to was missing empty line here:
image

I have a habit to run code formatter manually before every commit (ctr + alt + L in IntelliJ on windows),
but I had my battles with it, since it can mess the import order sometimes even though I have proper configuration ;)

flink/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/DeltaPendingFile.java
scottsand-db: I see. Good point. `basePath` is correct.

flink/src/main/java/io/delta/flink/sink/internal/writer/stats/DeltaFileStats.java
```diff
@@ -0,0 +1,286 @@
...
+ * |    |-- b: struct
+ * |    |    |-- c: long
+ * |-- d: int
+ *
+ * parquet stats would be captured as:
+ * recordCount: 1000
+ * a.b.c, LongStatistics{min=10, max=2000, nullCount=0}
+ * d,     IntStatistics{min=5, max=5, nullCount=20}
+ *
+ * This implementation recreates the nested structure required for delta stats which then become:
```
scottsand-db, line 64: this is great javadoc. but why is it for `JsonStatNoop`? shouldn't this be for `DeltaFileStats`?

flink/src/main/java/io/delta/flink/sink/internal/writer/stats/DeltaFileStats.java
```diff
@@ -0,0 +1,286 @@
...
+        return objectMapper.getNodeFactory().nullNode();
+    }
+    @Override
+    public JsonNode getNullCount() { return objectMapper.getNodeFactory().nullNode(); }
+}
+
+/**
+ * Factory class that instantiates appropriate {@link JsonStat} based on the delta {@link DataType}.
+ */
+class JsonStatFactory {
```
scottsand-db, line 114: can you move all of these `JsonStat...` to `JsonFileStats.java`

flink/src/main/java/io/delta/flink/sink/internal/writer/stats/DeltaFileStats.java
```diff
@@ -0,0 +1,286 @@
...
+                stat -> statFactory.newJsonStat(stat).getNullCount()));
+        try {
+            // Delta log requires stat json to be escaped as a string, hence we need 2
+            // level serialization.
+            return objectMapper.writeValueAsString(root.toString());
+        } catch (JsonProcessingException e) {
+            throw new RuntimeException(e);
+        }
+    }
+    private JsonNode getStat(Type parquetType, ArrayList<String> path,
```
scottsand-db, line 266: nit: can you format like:
```java
private JsonNode getStat(
        Type parquetType,
        ArrayList<String> path,
        Function<Statistics<?>,
        JsonNode> toJsonNode) {
```

flink/src/main/java/io/delta/flink/sink/internal/writer/stats/ParquetFileStats.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    public long getRowCount() {
+        return rowCount;
+    }
+
+    private Map<ColumnPath, Statistics<?>> load(ParquetMetadata parquetMetadata) {
+        Map<ColumnPath, Statistics<?>> columnStats = new HashMap<>();
+        MessageType parquetSchema = parquetMetadata.getFileMetaData().getSchema();
+        // assert that parquetSchema and deltaSchema match.
```
scottsand-db, line 66: where is this assert happening?

flink/src/main/java/io/delta/flink/sink/internal/writer/stats/ParquetFileStats.java
```diff
@@ -0,0 +1,81 @@
...
+
+    public Map<ColumnPath, Statistics<?>> getColumnStats() {
+        return columnStats;
+    }
+
+    public long getRowCount() {
+        return rowCount;
+    }
+
+    private Map<ColumnPath, Statistics<?>> load(ParquetMetadata parquetMetadata) {
```
scottsand-db, line 63: nit: javadoc explaining what you are doing (i.e merging the stats)

flink/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
```diff
@@ -270,9 +270,11 @@ private void doCommit(OptimisticTransaction transaction,
             " checkpointId=" + checkpointId +
             " files:" + logFiles
         );
+        StructType streamSchema = SchemaConverter.toDeltaDataType(rowType);
+        LOG.info("Schema: " + streamSchema);
         for (DeltaCommittable deltaCommittable : committables) {
             DeltaPendingFile deltaPendingFile = deltaCommittable.getDeltaPendingFile();
-            AddFile action = deltaPendingFile.toAddFile();
+            AddFile action = deltaPendingFile.toAddFile(basePath, streamSchema);
```
scottsand-db, line 277: note that here, `basePath` is `Root path of the DeltaTable`

flink/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/DeltaPendingFile.java
```diff
@@ -112,20 +119,31 @@ public LinkedHashMap<String, String> getPartitionSpec() {
     /**
      * Converts {@link DeltaPendingFile} object to a {@link AddFile} object
      *
+     * @param basePath Base path for the file.
```
scottsand-db, line 122: this is wrong. this is `Root path of the DeltaTable`

flink/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/DeltaPendingFile.java
```diff
@@ -112,20 +119,31 @@ public LinkedHashMap<String, String> getPartitionSpec() {
...
-    public AddFile toAddFile() {
+    public AddFile toAddFile(Path basePath, StructType schema) {
         LinkedHashMap<String, String> partitionSpec = this.getPartitionSpec();
+        String stats = null;
         long modificationTime = this.getLastUpdateTime();
         String filePath = PartitionPathUtils.generatePartitionPath(partitionSpec) +
             this.getFileName();
+        try {
+            ParquetFileStats parquetStats = ParquetFileStats.readStats(
+                new Path(basePath, filePath).toString());
```
scottsand-db, line 133: please inline comment that this is the absolute file path.

flink/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/DeltaPendingFile.java
```diff
@@ -112,20 +119,31 @@ public LinkedHashMap<String, String> getPartitionSpec() {
...
+            DeltaFileStats deltaStats = new DeltaFileStats(schema, parquetStats);
+            stats = deltaStats.toJson();
+        } catch (IOException ioe) {
+            // TODO: If we fail computing stats, we should throw and prevent commit.
+            LOG.error("failed computing stats", ioe);
```
scottsand-db, line 138: can you wrap this whole stats processing inside of a feature flag? `flink.stats.collect` default true. this value will be provided inside of the `conf: Configuration` when the Flink Sink Builder is constructed.

flink/src/test/java/io/delta/flink/sink/internal/writer/stats/ParquetFileStatsTest.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private static final String PATH = "/test-data/test-parquet-stats" +
+        "/part-d0e3e782-2a5d-49ca-a4d0-3a9df5a8f37c-0.snappy.parquet";
+
+    @Test
+    public void testParquetStatsFromFile_fieldTypes() throws Exception {
+        File resourcesDirectory = new File("src/test/resources");
+        String initialTablePath =
+            resourcesDirectory.getAbsolutePath() + PATH;
+        ParquetFileStats stats = ParquetFileStats.readStats(initialTablePath);
```
scottsand-db, line 28: this will read all the parquet files, right?

scottsand-db (changes requested):
Looking good. Left some comments.

Major requests:
- feature flag
- benchmark this feature to determine the overhead of read-after-write for the stats parsing

flink/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/DeltaPendingFile.java
kristoffSC: I will strongly suggest use `DeltaFileCommitter` over `GlobalCommitter` approach for solution that is based on reading stats back from Parquet file.

flink/src/main/java/io/delta/flink/sink/internal/writer/stats/DeltaFileStats.java
gopik: Thanks, it should be for `DeltaFileStats`. moved.

flink/src/main/java/io/delta/flink/sink/internal/writer/stats/DeltaFileStats.java
gopik: Moved to separate file/class.

flink/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/DeltaPendingFile.java
gopik: @kristoffSC @scottsand-db I've closed on all the comments except the following:
1. Hide stats impl behind feature flag - Could you point me to an example of any existing flag.
2. Implement stats computation in `DeltaFileCommitter` instead of `GlobalCommitter` - I'm working on this.
3. Benchmarking - do we have an existing benchmarking setup? If so, could you point. me to the docs.

flink/src/test/java/io/delta/flink/sink/internal/writer/stats/ParquetFileStatsTest.java
gopik: No, this reads only the file pointed by `PATH` variable. Let me delete other files from the repo.

flink/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/DeltaPendingFile.java
scottsand-db: @kristoffSC can you suggest the best way to add a feature flag for the Sink? It doesn't seem like there's any easy setup for this.

do you suggest we
a) have an `.option` API? on the sink?
b) extract the feature flag from the hadoop conf?

kristoffSC: Hi,
I would recommend to do it via `.option(...)`. And the implementation for `.option(...`) for Delta Sink Builder in my opinion should be implemented in separated branch.

The hadoop conf is needed/used only by delta standalone. There is no interaction between Flink specific classes from Delta connector with hadoop conf and I think we should need to have a good reason to change this.

@gopik
 you can check  `DeltaSourceBuilderBase.java` to see how `.option(...)` was implemented there and do the same thing for `DeltaSinkBuilder`. Heads up, both builder implementations differ in low level details. 

Also please note that .option(...) is overloaded, it accepts boolean, int, long and String as option values. Yours needs to have the same.

As for the feature flag, you need to have a `DeltaSourceOptions.java` like class for Sink and add the flag there as a `DeltaConfigOption` instance. 

A good example to follow would be IGNORE_DELETES option from `DeltaSourceOptions.java`. Having the API for .option(...) you can use the same definition as IGNORE_DELETES has, use same BooleanOptionTypeConverter and it should all be working.

flink/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
```diff
@@ -270,9 +270,11 @@ private void doCommit(OptimisticTransaction transaction,
             " checkpointId=" + checkpointId +
             " files:" + logFiles
         );
+        StructType streamSchema = SchemaConverter.toDeltaDataType(rowType);
+        LOG.info("Schema: " + streamSchema);
```
SteNicholas, line 274: Could the schema log move into above `Files to be committed to the Delta table:` log?

flink/src/main/java/io/delta/flink/sink/internal/writer/stats/JsonFileStats.java
```diff
@@ -0,0 +1,208 @@
...
+        public static final Logger LOG = LoggerFactory.getLogger(JsonStat.class);
+        protected ObjectMapper objectMapper;
+        private final Statistics<?> stat;
+
+        JsonStat(ObjectMapper objectMapper, Statistics<?> stat) {
+            this.objectMapper = objectMapper;
+            this.stat = stat;
+        }
+
+        public abstract JsonNode getMin();
```
SteNicholas, line 65: Does this need to introduce the `getRowCount` interface?

flink/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/DeltaPendingFile.java
```diff
@@ -112,20 +119,33 @@ public LinkedHashMap<String, String> getPartitionSpec() {
...
         String filePath = PartitionPathUtils.generatePartitionPath(partitionSpec) +
             this.getFileName();
+        try {
+            // Pass absolute path of the parquet file (rootPath + filePath within the root)
+            ParquetFileStats parquetStats = ParquetFileStats.readStats(
+                new Path(basePath, filePath).toString());
+            DeltaFileStats deltaStats = new DeltaFileStats(schema, parquetStats);
+            stats = deltaStats.toJson();
+        } catch (IOException ioe) {
+            // TODO: If we fail computing stats, we should throw and prevent commit.
```
SteNicholas, line 139: Which situation will fail to compute stats?

flink/src/main/java/io/delta/flink/sink/internal/writer/stats/JsonFileStats.java
gopik: Did you mean "numRecords"? This is added directly in DeltaFileStats.toJson impl.

flink/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/DeltaPendingFile.java
gopik: This implementation reads stats from the parquet files, so IOException during file reads may prevent stats computation.

flink/src/main/java/org/apache/flink/streaming/api/functions/sink/filesystem/DeltaPendingFile.java
```diff
@@ -112,20 +119,33 @@ public LinkedHashMap<String, String> getPartitionSpec() {
     /**
      * Converts {@link DeltaPendingFile} object to a {@link AddFile} object
      *
+     * @param basePath Root path of the DeltaTable.
```
SteNicholas, line 122:
```suggestion
     * @param basePath Root path of the delta table.
```

flink/src/main/java/io/delta/flink/sink/internal/writer/stats/JsonFileStats.java
SteNicholas: @gopik, right. I mean the `numRecords` for stats. I'm not sure whether it's better to add the `numRecords` in `JsonFileStats`.

gopik: The top level json is assembled in DeltaFileStats.toJson() and numRecords is a top level field (as are minValues, maxValues, nullCounts etc). JsonFileStats is a helper class that maps column types from parquet types to json. numRecords doesn't belong to any column. Could you explain via some code how do add numRecords in JsonFileStats?

scottsand-db:
Hi @gopik - just following up on this. You still need to refactor it so that the parquet footer reading is done inside `DeltaFileCommitter` instead of `DeltaGlobalCommitter`, correct?

gopik:
@scottsand-db @kristoffSC I've moved the call to compute stats from Global committer to delta committer. Stats needs 2 inputs, full file path and the schema.
There are 2 approaches -

1. DeltaCommitter has path and schema and passes it to DeltaPendingFile just after commit. DeltaPendingFile.onCommit(filePath, Schema). Would have been simpler, but looks like a lot of coupling between committer and pending file.
2. DeltaPendingFile has filePath and schema, only needs a onCommit invocation to compute stats. Looks cleaner but complex because of additional serialization/deserialization logic.

Please take a look and suggest if this approach seems ok or if you have other ideas.

scottsand-db:
Posting for visibility: this PR is currently blocked by https://github.com/delta-io/connectors/pull/425, as we need a way to provide options to the sink API to be able to wrap this feature behind a feature flag.

scottsand-db:
Hi @gopik - now that https://github.com/delta-io/connectors/pull/425 has been merged, are you interested in continuing this PR?

gopik:
@scottsand-db yes, I'll start working on this from next week.
@kristoffSC : could you take a look at the changes related to serialization due to this change and see if they are ok? Specifically changes in flink/src/main/java/io/delta/flink/sink/internal/committables/DeltaCommittableSerializer.java

scottsand-db:
@gopik just following up - any update?

gopik:
@scottsand-db I'm waiting for an ack from @kristoffSC on the approach. This change requires changes to serialization for checkpointing.

gopik:
@scottsand-db @kristoffSC This change is now ready for review. The implementation is now simplified to avoid any changes to serialization during checkpointing by avoiding any state in DeltaPendingFile. I've also added the config option to compute stats optionally (disabled by default).

vkorukanti:

PR #394, opened 2022-07-12
Create `delta-standalone-parquet` project to refactor public API parquet dependencies out from `delta-standalone`

### LINKED ISSUE #362: Error when running benchmark tests
Hi,

We would like to try using this connector in our production environment and for that, we started to run some benchmarking tests, our Flink  app consumes events from Kinesis where each event is ~2k, then uses delta sink to convert to parquet and into delta.
Our parquet schema contains ~40 columns in different datatype.

We have tried running performance tests and noticed failures in:

```
2022-06-02 15:11:54 | at org.apache.parquet.hadoop.codec.NonBlockedCompressorStream.write(NonBlockedCompressorStream.java:48)
2022-06-02 15:11:54 | at org.apache.parquet.hadoop.codec.SnappyCompressor.setInput(SnappyCompressor.java:99)
2022-06-02 15:11:54 | at java.nio.ByteBuffer.allocateDirect(ByteBuffer.java:311)
2022-06-02 15:11:54 | at java.nio.DirectByteBuffer.<init>(DirectByteBuffer.java:123)
2022-06-02 15:11:54 | at java.nio.Bits.reserveMemory(Bits.java:695)
2022-06-02 15:11:54 | 2022-06-02 12:11:54,918 WARN org.apache.flink.runtime.taskmanager.Task [] - Sink DeltaAssetPacketSink (1/4)#8 (c1a0c51d8c2274b48092b9c0db8009ee) switched from RUNNING to FAILED with failure cause: java.lang.OutOfMemoryError: Direct buffer memory
```

And also failures such as:
```
2022-06-02 12:41:39 | at com.github.mjakubowski84.parquet4s.RowParquetRecord.$anonfun$write$1$adapted(ParquetRecord.scala:167)
2022-06-02 12:41:39 | at com.github.mjakubowski84.parquet4s.RowParquetRecord.$anonfun$write$1(ParquetRecord.scala:173)
2022-06-02 12:41:39 | at com.github.mjakubowski84.parquet4s.MapParquetRecord.write(ParquetRecord.scala:460)
2022-06-02 12:41:39 | at org.apache.parquet.schema.GroupType.getType(GroupType.java:207)
2022-06-02 12:41:39 | at org.apache.parquet.schema.GroupType.getFieldIndex(GroupType.java:175)
2022-06-02 12:41:39 | }
2022-06-02 12:41:39 | }
2022-06-02 12:41:39 | optional binary value (STRING);
 2022-06-02 12:41:39 | required binary key (STRING);
 2022-06-02 12:41:39 | repeated group key_value {
```

Our testing Includes a load of sending ~2M events separate across 30min.
Our setup includes:
- Flink 1.14.4
- Scala 2.12
- 2 TaskManagers with 4gig memory
- Parallelism 2
- Checkpoint every 15min

AWS Kinesis event source, 
Sink - S3 as Parquet and Delta log storage

It seems like the Flush of the delta happens only at checkpoint which might cause the out-of-memory problem.
Is there any configuration that could be done on the Delta sink to modify it? Is there any benchmarking test done on this connector?

scottsand-db:
## The Problem
We recently decided to shade our hadoop-parquet dependencies in delta-standalone to resolve various dependency conflict issues in downstream connectors, like what was experienced with the delta-flink connector. This was done in #382 to resolve #362.

However, delta-standalone exposes a public API in `ParquetSchemaConverter` that exposed parquet types. Since we shaded these dependencies, users of this API would experience errors like

```
[error] /home/runner/work/connectors/connectors/flink/src/main/java/io/delta/flink/sink/internal/writer/stats/DeltaFileStats.java:195:1: incompatible types: shadedelta.org.apache.parquet.schema.MessageType cannot be converted to org.apache.parquet.schema.MessageType
[error] ParquetSchemaConverter.deltaToParquet(schema)
```

This happened in #393 in this [build run](https://github.com/delta-io/connectors/runs/7294813067?check_suite_focus=true).

## The Solution
To solve this, we move the `ParquetSchemaConverter` class, and its test, into a separate `standalone-parquet` project. We keep the package name the same.`standalone-parquet` then depends on `delta-standalone-original`, and then `delta-standalone` (the project we actually publish) depends on `standalone-parquet`.

We test this refactor and new project by adding a trivial test to `delta-flink` which uses a public API from `ParquetSchemaConverter`.

### How was this solution tested?

I compared the jars of the master jar and the jar produced from this change for `delta-standalone`, using `jar -tf`, and they were the same.


utils/src/main/java/io/delta/utils/ParquetSchemaConverter.java
```diff
@@ -14,12 +14,12 @@
  * limitations under the License.
```
scottsand-db, line 14: refactored from `delta-standalone` to `delta-utils`.

utils/src/test/scala/io/delta/utils/ParquetSchemaConverterSuite.scala
```diff
@@ -14,13 +14,13 @@
  * limitations under the License.
```
scottsand-db, line 14: refactored from `delta-standalone` to `delta-utils`.

build.sbt
```diff
@@ -651,6 +671,7 @@ def flinkScalaVersion(scalaBinaryVersion: String): String = {
 val flinkVersion = "1.13.0"
 lazy val flink = (project in file("flink"))
   .dependsOn(standaloneCosmetic % "provided")
+  .dependsOn(utils)
```
zsxwing, line 674: We should remove this. `flink` should just depend on the single jar generated from `standaloneCosmetic`

build.sbt
```diff
@@ -404,6 +404,26 @@ def scalaCollectionPar(version: String) = version match {
     case _ => Seq()
 }
 
+lazy val utils = (project in file("utils"))
+  .dependsOn(standaloneCosmetic % "provided")
+  .enablePlugins(GenJavadocPlugin, JavaUnidocPlugin)
+  .settings(
+    name := "delta-utils",
+    commonSettings,
+    publishArtifact := scalaBinaryVersion.value == "2.12", // only publish once
```
zsxwing, line 413: we should not publish this.

build.sbt
```diff
@@ -404,6 +404,26 @@ def scalaCollectionPar(version: String) = version match {
...
+    Test / publishArtifact := false,
+    crossPaths := false,
+    libraryDependencies ++= Seq(
+      "org.apache.parquet" % "parquet-hadoop" % "1.12.0" % "provided",
+      "org.scalatest" %% "scalatest" % scalaTestVersion % "test",
+
+      // Compiler plugins
+      // -- Bump up the genjavadoc version explicitly to 0.18 to work with Scala 2.12
+      compilerPlugin("com.typesafe.genjavadoc" %% "genjavadoc-plugin" % "0.18" cross CrossVersion.full)
+    )
```
zsxwing, line 424: The javadoc may be a problem since we now remove `ParquetSchemaConverter` from the standalone project.

utils/src/main/java/io/delta/utils/ParquetSchemaConverter.java
```diff
@@ -14,12 +14,12 @@
...
  */
 
-package io.delta.standalone.util;
```
zsxwing, line 17: we cannot change this package. This should be the same to maintain the backward compatibility

zsxwing (commented):
How about calling the new project `standalone-parquet`? `utils` is not a good name since most of utils still stay in `standalone`.

build.sbt
```diff
@@ -404,13 +405,25 @@ def scalaCollectionPar(version: String) = version match {
     case _ => Seq()
 }
 
+lazy val standaloneParquet = (project in file("standalone-parquet"))
```
zsxwing, line 408: nit: add a comment to explain why we need a separate project, such as
```
The public API ParquetSchemaConverter exposes Parquet classes in its methods so we cannot apply shading rules on it. However, sbt-assembly doesn't allow to exclude a single file. Hence, we create a separate project to skip the shading.
```

flink/src/test/java/io/delta/flink/misc/TestParquetDependency.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.apache.parquet.schema.MessageType;
+import org.junit.Test;
+
+import io.delta.standalone.types.LongType;
+import io.delta.standalone.types.StringType;
+import io.delta.standalone.types.StructType;
+import io.delta.standalone.util.ParquetSchemaConverter;
+
+public class TestParquetDependency {
```
zsxwing, line 27: How about moving this test to `testStandaloneCosmetic` since this test has nothing to do with flink?

zsxwing (approved):
LGTM. Only minor style comments

zsxwing (approved):
LGTM

flink/src/test/java/io/delta/flink/misc/TestParquetDependency.java
scottsand-db: spoke offline. we will leave it as is. this tests that delta-flink pulls in the right parquet dependencies, given that delta-standalone-parquet required parquet-hadoop 1.12.0.

PR #397, opened 2022-07-13
Column stats based file pruning: 1. parse and store the column stats

6a0juu:
Implement the column stats storage structs and parsing methods.

Test the parse and storage of column stats and its parsing methods
* The column stats parsing with no nested columns
* Parsing with nested columns
* Parsing with missing stats
* Parsing with stats cannot be found on table schema

### DIFF
--- a/standalone/src/main/java/io/delta/standalone/expressions/Column.java
+++ b/standalone/src/main/java/io/delta/standalone/expressions/Column.java
@@ -1,11 +1,13 @@
 package io.delta.standalone.expressions;
 
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.Objects;
 import java.util.Set;
 
 import io.delta.standalone.data.RowRecord;
 import io.delta.standalone.types.*;
+import io.delta.standalone.internal.exception.DeltaErrors;
 
 /**
  * A column whose row-value will be computed based on the data in a {@link RowRecord}.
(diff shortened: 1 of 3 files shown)


standalone/src/main/java/io/delta/standalone/expressions/Column.java
```diff
@@ -17,51 +19,62 @@
  * <a href="https://github.com/delta-io/delta/blob/master/PROTOCOL.md#primitive-types">Delta Transaction Log Protocol: Primitive Types</a>.
  */
 public final class Column extends LeafExpression {
-    private final String name;
+    private final String[] name;
```
scottsand-db, line 22: `name` is no longer a good variable name. I have no idea what `String[] name` is now, why is it an array?

standalone/src/main/scala/io/delta/standalone/internal/scan/FilteredDeltaScanImpl.scala
```diff
@@ -57,4 +80,79 @@ final private[internal] class FilteredDeltaScanImpl(
...
+      if (curSchema.isEmpty || !curSchema.get.dataType.isInstanceOf[StructType]) {
+        return None
+      }
+      curSchema = curSchema.get.nestedColumns.get(x)
+    }
+    curSchema
+  }
+
+  /** Parse the entire [[AddFile.stats]] string based on table schema */
+  def parseColumnStats(
```
scottsand-db, line 96: where will this be used? currently I only see this being used by `def testStatsParser` inside of `DataSkippingSuite`.

also, so this is not a pure function? this will change the class variable `statsStore`?

standalone/src/main/scala/io/delta/standalone/internal/scan/FilteredDeltaScanImpl.scala
```diff
@@ -57,4 +80,79 @@ final private[internal] class FilteredDeltaScanImpl(
...
+      tableSchema: DataType,
+      addFile: AddFile): Unit = {
+    // Reset the store for current AddFile
+    statsStore = ColumnNode()
```
scottsand-db, line 100: overall i'm a little confused why parsing the stats for an AddFile is updating some class-wide state.

standalone/src/main/scala/io/delta/standalone/internal/scan/FilteredDeltaScanImpl.scala
```diff
@@ -41,6 +61,9 @@ final private[internal] class FilteredDeltaScanImpl(
   private val (metadataConjunction, dataConjunction) =
     PartitionUtils.splitMetadataAndDataPredicates(expr, partitionColumns)
 
+  /** The object saving column stats based on the structure of table schema */
+  private var statsStore = ColumnNode()
```
scottsand-db, line 65: how will this be used? how should it be updated? once per query? once per AddFile? i'm concerned why it is a class-level variable, (`var`, not `val`). I see it best being used as
- a class-level `val` that is calculated once on construction
- a private `val` inside of a helper method that is calculated per each AddFile.

Overall I'm not really sure what this variable is going to be used for.

standalone/src/main/java/io/delta/standalone/expressions/Column.java
```diff
@@ -17,51 +19,62 @@
...
+        if (name.length < 1) {
+            throw DeltaErrors.invalidColumnName(Arrays.toString(name));
+        }
+
         this.name = name;
         this.dataType = dataType;
 
         if (dataType instanceof IntegerType) {
-            evaluator = (record -> record.getInt(name));
+            evaluator = (record -> record.getInt(name[0]));
```
scottsand-db, line 37: you seem to only be using `name[0]`. what is special about the 0th index? what does it represent? if it is so special, why not assign it to a local variable, instead of calling `name[0]` repeatedly.

scottsand-db (changes requested):

scottsand-db (changes requested):

6a0juu:
Project plan changed, will make an end-to-end first, then extend support for nested columns and other expression types.

PR #406, opened 2022-07-25
Make delta-standalone jar include ParquetSchemaConverter

zsxwing:
This PR fixes the issue that the delta-standalone jar is missing `ParquetSchemaConverter` by:

- Create a dummy project `standaloneWithoutParquetUtils` to allow `standaloneParquet` depending on the shaded delta-standalone jar.
- Shade `standaloneParquet`.
- `standaloneCosmetic` depends on the shaded `standaloneParquet` which include `ParquetSchemaConverter`.
-  Add `testParquetUtilsWithStandaloneCosmetic` and move tests for `ParquetSchemaConverter` into it to verify we shade the code correctly.


.github/workflows/test.yaml
```diff
@@ -22,7 +22,7 @@ jobs:
             ~/.ivy2
             ~/.cache/coursier
             ~/.m2
-          key: build-cache-2-with-scala_${{ matrix.scala }}
+          key: build-cache-3-with-scala_${{ matrix.scala }}
```
zsxwing, line 25: change the key to re-cache since we made a few dependency changes recently.

build.sbt
```diff
@@ -372,15 +372,15 @@ lazy val hive2Tez = (project in file("hive2-tez")) settings (
...
   .settings(
     name := "delta-standalone",
     commonSettings,
     releaseSettings,
     exportJars := true,
-    Compile / packageBin := (standalone / assembly).value,
+    Compile / packageBin := (standaloneParquet / assembly).value,
     libraryDependencies ++= scalaCollectionPar(scalaVersion.value) ++ Seq(
       "org.apache.hadoop" % "hadoop-client" % hadoopVersion % "provided",
+      "org.apache.parquet" % "parquet-hadoop" % "1.12.0" % "provided",
```
zsxwing, line 383: Adding this back as 0.4.1 has this in pom.xml as a provided dependency.

scottsand-db (approved):
LGTM

scottsand-db:
Verified Hive jars. Awesome.

PR #408, opened 2022-07-25
Upgrade delta-storage dependency version to 2.0.0

scottsand-db:


allisonport-db (approved):
lgtm, should we do anything to make sure this stays up to date?

scottsand-db:
> lgtm, should we do anything to make sure this stays up to date?

Any suggestions? They exist in different repos. Perhaps we could write a script that runs in every github ci job?

PR #413, opened 2022-07-27
Column stats based file pruning: expression support p0

6a0juu:
Support following expression type in [column stats based file pruning](https://docs.google.com/document/d/1D2ukbGInriKTOZJl3r3pCZoQRji40rSuFgnOyzhHz7k/edit).
* equalTo
* greaterThan
* greaterThanOrEqual
* lessThan
* lessThanOrEqual
* OR 
* AND

### DIFF
--- a/standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
@@ -275,8 +275,8 @@ private[internal] object DataSkippingUtils {
    *
    * - When left child is literal value and right child is column, we use `litColRule`. This will
    * rewrite the expression to `colLitRule` format (the left child is column and right child is
-   * literal value) in [[buildBinaryComparatorFilter]] and then call [[constructDataFilters]]
-   * again to build the target expression by `colLitRule`.
+   * literal value) in [[buildBinaryComparatorFilter]] and then use the rewritten expression call
+   * [[constructDataFilters]] again to build the target expression by `colLitRule`.
    *
    * - When both children are columns, we use `colColRule` to build the stats filter.
    *
--- a/standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
@@ -310,8 +310,7 @@ private[internal] object DataSkippingUtils {
         val litColRule = (r: LitColRule) => new GreaterThan(r.right, r.left)
 
         // (col1 < col2) -> (MIN.col1 < MAX.col2)
-        val colColRule = (r: ColColRule) =>
-          new LessThan(r.leftMin, r.rightMax)
+        val colColRule = (r: ColColRule) => new LessThan(r.leftMin, r.rightMax)
         buildBinaryComparatorFilter(dataSchema, lt, colColRule, colLitRule, litColRule)
 
       case gt: GreaterThan =>
@@ -322,8 +321,7 @@ private[internal] object DataSkippingUtils {
         val litColRule = (r: LitColRule) => new LessThan(r.right, r.left)
 
         // (col1 > col2) -> (MAX.col1 > MIN.col2)
-        val colColRule = (r: ColColRule) =>
-          new GreaterThan(r.leftMax, r.rightMin)
+        val colColRule = (r: ColColRule) => new GreaterThan(r.leftMax, r.rightMin)
         buildBinaryComparatorFilter(dataSchema, gt, colColRule, colLitRule, litColRule)
 
       case leq: LessThanOrEqual =>
@@ -334,8 +332,7 @@ private[internal] object DataSkippingUtils {
         val litColRule = (r: LitColRule) => new GreaterThanOrEqual(r.right, r.left)
 
         // (col1 <= col2) -> (MIN.col1 <= MAX.col2)
-        val colColRule = (r: ColColRule) =>
-          new LessThanOrEqual(r.leftMin, r.rightMax)
+        val colColRule = (r: ColColRule) => new LessThanOrEqual(r.leftMin, r.rightMax)
         buildBinaryComparatorFilter(dataSchema, leq, colColRule, colLitRule, litColRule)
 
       case geq: GreaterThanOrEqual =>
@@ -346,8 +343,7 @@ private[internal] object DataSkippingUtils {
         val litColRule = (r: LitColRule) => new LessThanOrEqual(r.right, r.left)
 
         // (col1 >= col2) -> (MAX.col1 >= MIN.col2)
-        val colColRule = (r: ColColRule) =>
-          new GreaterThanOrEqual(r.leftMax, r.rightMin)
+        val colColRule = (r: ColColRule) => new GreaterThanOrEqual(r.leftMax, r.rightMin)
         buildBinaryComparatorFilter(dataSchema, geq, colColRule, colLitRule, litColRule)
 
       case and: And =>
--- a/standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
@@ -46,24 +46,26 @@ private[internal] object DataSkippingUtils {
     new FloatType, new IntegerType, new LongType, new ShortType)
 
   /**
-   * The building rule when left and right children are both columns. The 4 parameters from left to
-   * right are: MIN of left child, MAX of left child, MIN of right child, MAX of right child.
+   * The building rule when left and right children are both columns.
+   * The 4 parameters from left to right are: MIN of left child, MAX of left child,
+   * MIN of right child, MAX of right child.
    */
-  private type ColColRule = (Column, Column, Column, Column) => Expression
+  private case class ColColRule(
+      leftMin: Column, leftMax: Column, rightMin: Column, rightMax: Column)
 
   /**
-   * The building rule when left child is a column and right child is a literal value. The 3
-   * parameters from left to right are: MIN of left child, MAX of left child, the literal value
-   * at the right child.
+   * The building rule when left child is a column and right child is a literal value.
+   * The 3 parameters from left to right are: MIN of left child, MAX of left child,
+   * the literal value at the right child.
    */
-  private type ColLitRule = (Column, Column, Literal) => Expression
+  private case class ColLitRule(leftMin: Column, leftMax: Column, right: Literal)
 
   /**
-   * The building rule when left child is a literal value and right child is a column. The 2
-   * parameters from left to right are: the literal value at the left child. the column at the
-   * right child.
+   * The building rule when left child is a literal value and right child is a column.
+   * The 2 parameters from left to right are: the literal value at the left child,
+   * the column at the right child.
    */
-  private type LitColRule = (Literal, Column) => Expression
+  private case class LitColRule(left: Literal, right: Column)
 
   /**
    * Build stats schema based on the schema of data columns, the first layer
@@ -238,24 +240,24 @@ private[internal] object DataSkippingUtils {
   def buildBinaryComparatorFilter(
       dataSchema: StructType,
       expr: BinaryComparison,
-      colColRule: ColColRule,
-      colLitRule: ColLitRule,
-      litColRule: LitColRule): Option[Expression] = {
+      colColRule: ColColRule => Expression,
+      colLitRule: ColLitRule => Expression,
+      litColRule: LitColRule => Expression): Option[Expression] = {
     (expr.getLeft, expr.getRight) match {
       case (e1: Column, e2: Column) =>
         // Apply `colColRule` with MIN and MAX from both children.
-        val leftMinMaxCol = getMinMaxColumn(dataSchema, e1.name).getOrElse { return None }
-        val rightMinMaxCol = getMinMaxColumn(dataSchema, e2.name).getOrElse { return None }
-        Some(colColRule(leftMinMaxCol._1, leftMinMaxCol._2, rightMinMaxCol._1, rightMinMaxCol._2))
+        val (leftMin, leftMax) = getMinMaxColumn(dataSchema, e1.name).getOrElse { return None }
+        val (rightMin, rightMax) = getMinMaxColumn(dataSchema, e2.name).getOrElse { return None }
+        Some(colColRule(ColColRule(leftMin, leftMax, rightMin, rightMax)))
       case (e1: Column, e2: Literal) =>
         // Apply `colLitRule` with MIN and MAX of the column at left child and the literal value
         // at the right child.
-        val leftMinMaxCol = getMinMaxColumn(dataSchema, e1.name).getOrElse { return None }
-        Some(colLitRule(leftMinMaxCol._1, leftMinMaxCol._2, e2))
+        val (leftMin, leftMax) = getMinMaxColumn(dataSchema, e1.name).getOrElse { return None }
+        Some(colLitRule(ColLitRule(leftMin, leftMax, e2)))
       case (e1: Literal, e2: Column) =>
         // Apply `litColRule` by swap the left and right child. Then `constructDataFilters` will use
         // `colLitRule` to solve this problem.
-        constructDataFilters(dataSchema, Some(litColRule(e1, e2)))
+        constructDataFilters(dataSchema, Some(litColRule(LitColRule(e1, e2))))
 
       // If left and right children are both literal value, we return the original expression.
       case (_: Literal, _: Literal) => Some(expr)
@@ -285,71 +287,67 @@ private[internal] object DataSkippingUtils {
     expr match {
       case eq: EqualTo =>
         // (col1 == lit1) -> (MIN.col1 <= lit1 AND MAX.col1 >= lit1)
-        val colLitRule = (minCol: Column, maxCol: Column, lit1: Literal) =>
+        val colLitRule = (r: ColLitRule) =>
           new And(
-            new LessThanOrEqual(minCol, lit1),
-            new GreaterThanOrEqual(maxCol, lit1))
+            new LessThanOrEqual(r.leftMin, r.right),
+            new GreaterThanOrEqual(r.leftMax, r.right))
 
         // (lit1 == col1) -> (col1 == lit1)
-        val litColRule = (lit1: Literal, col1: Column) => new EqualTo(col1, lit1)
+        val litColRule = (r: LitColRule) => new EqualTo(r.right, r.left)
 
         // (col1 == col2) -> (MIN.col1 <= MAX.col2 AND MAX.col1 >= MIN.col2)
-        val colColRule = (e1Min: Column, e1Max: Column, e2Min: Column, e2Max: Column) =>
+        val colColRule = (r: ColColRule) =>
           new And(
-            new LessThanOrEqual(e1Min, e2Max),
-            new GreaterThanOrEqual(e1Max, e2Min))
+            new LessThanOrEqual(r.leftMin, r.rightMax),
+            new GreaterThanOrEqual(r.leftMax, r.rightMin))
         buildBinaryComparatorFilter(dataSchema, eq, colColRule, colLitRule, litColRule)
 
       case lt: LessThan =>
         // (col1 < lit1) -> (MIN.col1 < lit1)
-        val colLitRule = (minCol: Column, _: Column, lit1: Literal) =>
-          new LessThan(minCol, lit1)
+        val colLitRule = (r: ColLitRule) => new LessThan(r.leftMin, r.right)
 
         // (lit1 < col1) -> (col1 > lit1)
-        val litColRule = (lit1: Literal, col1: Column) => new GreaterThan(col1, lit1)
+        val litColRule = (r: LitColRule) => new GreaterThan(r.right, r.left)
 
         // (col1 < col2) -> (MIN.col1 < MAX.col2)
-        val colColRule = (e1Min: Column, _: Column, _: Column, e2Max: Column) =>
-          new LessThan(e1Min, e2Max)
+        val colColRule = (r: ColColRule) =>
+          new LessThan(r.leftMin, r.rightMax)
         buildBinaryComparatorFilter(dataSchema, lt, colColRule, colLitRule, litColRule)
 
       case gt: GreaterThan =>
         // (col1 > lit1) -> (MAX.col1 > lit1)
-        val colLitRule = (_: Column, maxCol: Column, lit1: Literal) =>
-          new GreaterThan(maxCol, lit1)
+        val colLitRule = (r: ColLitRule) => new GreaterThan(r.leftMax, r.right)
 
         // (lit1 > col1) -> (col1 < lit1)
-        val litColRule = (lit1: Literal, col1: Column) => new LessThan(col1, lit1)
+        val litColRule = (r: LitColRule) => new LessThan(r.right, r.left)
 
         // (col1 > col2) -> (MAX.col1 > MIN.col2)
-        val colColRule = (_: Column, e1Max: Column, e2Min: Column, _: Column) =>
-          new GreaterThan(e1Max, e2Min)
+        val colColRule = (r: ColColRule) =>
+          new GreaterThan(r.leftMax, r.rightMin)
         buildBinaryComparatorFilter(dataSchema, gt, colColRule, colLitRule, litColRule)
 
       case leq: LessThanOrEqual =>
         // (col1 <= lit1) -> (MIN.col1 <= lit1)
-        val colLitRule = (minCol: Column, _: Column, lit1: Literal) =>
-          new LessThanOrEqual(minCol, lit1)
+        val colLitRule = (r: ColLitRule) => new LessThanOrEqual(r.leftMin, r.right)
 
         // (lit1 <= col1) -> (col1 >= lit1)
-        val litColRule = (lit1: Literal, col1: Column) => new GreaterThanOrEqual(col1, lit1)
+        val litColRule = (r: LitColRule) => new GreaterThanOrEqual(r.right, r.left)
 
         // (col1 <= col2) -> (MIN.col1 <= MAX.col2)
-        val colColRule = (e1Min: Column, _: Column, _: Column, e2Max: Column) =>
-          new LessThanOrEqual(e1Min, e2Max)
+        val colColRule = (r: ColColRule) =>
+          new LessThanOrEqual(r.leftMin, r.rightMax)
         buildBinaryComparatorFilter(dataSchema, leq, colColRule, colLitRule, litColRule)
 
       case geq: GreaterThanOrEqual =>
         // (col1 >= lit1) -> (MAX.col1 >= lit1)
-        val colLitRule = (_: Column, maxCol: Column, lit1: Literal) =>
-          new GreaterThanOrEqual(maxCol, lit1)
+        val colLitRule = (r: ColLitRule) => new GreaterThanOrEqual(r.leftMax, r.right)
 
         // (lit1 >= col1) -> (col1 <= lit1)
-        val litColRule = (lit1: Literal, col1: Column) => new LessThanOrEqual(col1, lit1)
+        val litColRule = (r: LitColRule) => new LessThanOrEqual(r.right, r.left)
 
         // (col1 >= col2) -> (MAX.col1 >= MIN.col2)
-        val colColRule = (_: Column, e1Max: Column, e2Min: Column, _: Column) =>
-          new GreaterThanOrEqual(e1Max, e2Min)
+        val colColRule = (r: ColColRule) =>
+          new GreaterThanOrEqual(r.leftMax, r.rightMin)
         buildBinaryComparatorFilter(dataSchema, geq, colColRule, colLitRule, litColRule)
 
       case and: And =>
--- a/standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
@@ -52,13 +52,8 @@ private[internal] object DataSkippingUtils {
   /** The building rule when left child is a column and right child is a literal value. */
   private case class ColLitWrapper(colMin: Column, colMax: Column, lit: Literal)
 
-  /**
-   * The building rule when left child is a literal value and right child is a column. The
-   * `litColRule` will transform a binary expression from `expr(lit, col)` to `new_expr(col, lit)`.
-   * Afterwards `constructDataFilters` will use `colLitRule` to build the result from
-   * `new_expr(col, lit)` with `colLitRule`.
-   */
-  private case class LitColWrapper(col: Column, lit: Literal)
+  /** The building rule when left child is a literal value and right child is a column. */
+  private case class LitColWrapper(lit: Literal, col: Column)
 
   /**
    * Build stats schema based on the schema of data columns, the first layer
@@ -211,8 +206,8 @@ private[internal] object DataSkippingUtils {
    *
    * Supported expressions are: LessThan, LessThanOrEqual, GreaterThan, GreaterThanOrEqual, EqualTo
    *
-   * We only handle the case when `Column` or `Literal` value is the child of binary comparator.
-   * Because they are the only two leaf expression types.
+   * We only handle the case when both children of the binary comparator are either `Column` or
+   * `Literal` value. Because these are the only two supported leaf expression types.
    *
    * For example, `col1` is a column and `lit1` is a literal value. When building column stats
    * predicate with query predicate `col1 < lit1`. This method will follow the `colLitRule` because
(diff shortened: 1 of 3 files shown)


standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,18 +168,85 @@ private[internal] object DataSkippingUtils {
...
+  }
+
+  /**
+   * Build column stats predicate based on [[BinaryComparison]] expression in query predicate.
+   * Specifically, this method will apply one of the rules in parameter based on the expression
+   * type of left and right children.
+   *
+   * @param dataSchema The schema of data columns in table.
+   * @param expr       The expression from query predicate.
+   * @param ccRule     The building clRule when left and right children are both columns.
```
scottsand-db, line 199: Your definition of `ccRule` starts by saying it is `The building clRule`. I'm confused.

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,18 +168,85 @@ private[internal] object DataSkippingUtils {
...
+   * @param clRule     The building clRule when left child is a column and right child is a literal
```
scottsand-db, line 200: `clRule` isn't a proper known. So saying that the param `clRule` is "The building clRule" isn't helpful.

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,18 +168,85 @@ private[internal] object DataSkippingUtils {
...
+        Some(ccRule(leftMinMaxCol._1, leftMinMaxCol._2, rightMinMaxCol._1, rightMinMaxCol._2))
+      case (e1: Column, e2: Literal) =>
+        val leftMinMaxCol = getMinMaxColumn(dataSchema, e1.name).getOrElse { return None }
+        Some(clRule(leftMinMaxCol._1, leftMinMaxCol._2, e2))
+      case (e1: Literal, e2: Column) =>
+        constructDataFilters(dataSchema, Some(lcRule(e1, e2)))
+
+      // If left and right children are both literal value, we return the original expression.
+      case (_: Literal, _: Literal) => Some(expr)
+      case _ => None
```
scottsand-db, line 226: it is worth mentioning here, or in the method doc, why we only care about Columns and Literals. Why not any other expressions?

IMO it is because these two are terminal / leaf expressions.

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,18 +168,85 @@ private[internal] object DataSkippingUtils {
...
-   * Assume `col1` and `col2` are columns in query predicate, `l1` and `l2` are two
-   * literal values in query predicate. Let `f` be this method `constructDataFilters`.
-   * Now two rules are applied:
-   * - (col1 == l1) -> (MIN.col1 <= l1 AND MAX.col1 >= l1)
-   * - f(expr1 AND expr2) -> f(expr1) AND f(expr2)
+   * Assume `col1`, `col2` are columns in query predicate, `l1` is a literal values in query
+   * predicate, and `expr1`, `expr2` are two predicates in the query predicate. Let `f` be this
+   * method `constructDataFilters`.
+   *
+   * Now applied rules:
```
scottsand-db, line 237: if you copied this logic from delta lake, which I think you have, then you should give a URL link the that original code/comment and say this is where we have derived this logic from.

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,18 +168,85 @@ private[internal] object DataSkippingUtils {
...
+   * (col1 == l1) -> (MIN.col1 <= l1 AND MAX.col1 >= l1)
```
scottsand-db, line 238: IMO you should copy these individual comments to be beside the actual `case` statements below which implement them. That way, next to each case statement, is a simple human-readable comment that explains what's going on.

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -192,30 +257,65 @@ private[internal] object DataSkippingUtils {
       dataSchema: StructType,
       dataConjunction: Option[Expression]): Option[Expression] =
     dataConjunction match {
```
scottsand-db, line 259: why not just check if `dataConjunction` is empty early on?

that way you can match on `dataConjunction.get` instead of on `dataConjunction`.
This would mean that we don't have to wrap every case statement in `Some(...)`

standalone/src/test/scala/io/delta/standalone/internal/DataSkippingSuite.scala
```diff
@@ -216,7 +216,7 @@ class DataSkippingSuite extends FunSuite {
   }
 
   /**
-   * Filter: (col2 == 1 AND col2 == 2) (1 <= i <= 20)
+   * Filter: (col2 == 1 && col2 == 2) (1 <= i <= 20)
```
scottsand-db, line 219: you don't need the `(1 <= i <= 20)` here. does this significantly impact my understanding of this test case?

standalone/src/test/scala/io/delta/standalone/internal/DataSkippingUtilsSuite.scala
```diff
@@ -179,9 +182,34 @@ class DataSkippingUtilsSuite extends FunSuite {
         eqCast("col1", new LongType, long1),
         eqCast("col2", new LongType, long2))))
 
-    // col1 >= 1, `>=` is not supported
+    // partial `And`: `IsNotNull` is not supported, only return one child as output.
```
scottsand-db, line 185: what is "partial `And`"?

scottsand-db: have we implemented any sort of `partial and` implementation?

scottsand-db (changes requested):
Looking good. mainly just comments on your documentation.

standalone/src/test/scala/io/delta/standalone/internal/DataSkippingUtilsSuite.scala
6a0juu: Yes. [here](https://github.com/delta-io/connectors/pull/413/files#diff-ac21cc1184f49d8ec1b628b5106c84db06443f287ae0dfa16cf8aac5302be5f9R331) the `And` will return the non-null child when the other child get None value (the None returning is because of the unsupported expression type). This happens when building the column stats filter. Not when evaluating the filter.

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,58 +168,187 @@ private[internal] object DataSkippingUtils {
...
-   * - f(expr1 AND expr2) -> f(expr1) AND f(expr2)
+   * We only handle the case when `Column` or `Literal` value is the child of binary comparator.
+   * Because they are the only two leaf expression types.
+   *
+   * @param dataSchema The schema of data columns in table.
+   * @param expr       The expression from query predicate.
+   * @param ccRule     The building rule when left and right children are both columns.
+   * @param clRule     The building rule when left child is a column and right child is a literal
+   *                   value.
+   * @param lcRule     The building rule when left child is a literal value and left child is a
```
scottsand-db, line 207: `when left child` ... `and left child`. you say `left` twice

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,58 +168,187 @@ private[internal] object DataSkippingUtils {
...
-   * literal values in query predicate. Let `f` be this method `constructDataFilters`.
-   * Now two rules are applied:
-   * - (col1 == l1) -> (MIN.col1 <= l1 AND MAX.col1 >= l1)
-   * - f(expr1 AND expr2) -> f(expr1) AND f(expr2)
+   * We only handle the case when `Column` or `Literal` value is the child of binary comparator.
+   * Because they are the only two leaf expression types.
+   *
+   * @param dataSchema The schema of data columns in table.
+   * @param expr       The expression from query predicate.
+   * @param ccRule     The building rule when left and right children are both columns.
```
scottsand-db, line 204: if left and right children are both columns, why is the type `(Column, Column, Column, Column)`? 
basically, this is a complex parameter, and it isn't clear what all of its values are from the parameter description.

consider using case classes, with good documentation, instead of tuples.

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,58 +168,187 @@ private[internal] object DataSkippingUtils {
...
+    (expr.getLeft, expr.getRight) match {
+      case (e1: Column, e2: Column) =>
+        val leftMinMaxCol = getMinMaxColumn(dataSchema, e1.name).getOrElse { return None }
+        val rightMinMaxCol = getMinMaxColumn(dataSchema, e2.name).getOrElse { return None }
+        Some(ccRule(leftMinMaxCol._1, leftMinMaxCol._2, rightMinMaxCol._1, rightMinMaxCol._2))
+      case (e1: Column, e2: Literal) =>
+        val leftMinMaxCol = getMinMaxColumn(dataSchema, e1.name).getOrElse { return None }
+        Some(clRule(leftMinMaxCol._1, leftMinMaxCol._2, e2))
+      case (e1: Literal, e2: Column) =>
+        constructDataFilters(dataSchema, Some(lcRule(e1, e2)))
```
scottsand-db, line 227: why does (Literal, Column) call `constructDataFilters` and yet `(Column, Literal)` does not?

scottsand-db: you should add inline comments explaining what each `case` statement is doing, and why they are different

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,58 +168,187 @@ private[internal] object DataSkippingUtils {
...
+   *
+   * Supported expressions are: LessThan, LessThanOrEqual, GreaterThan, GreaterThanOrEqual, EqualTo
    *
-   * Assume `col1` and `col2` are columns in query predicate, `l1` and `l2` are two
-   * literal values in query predicate. Let `f` be this method `constructDataFilters`.
-   * Now two rules are applied:
-   * - (col1 == l1) -> (MIN.col1 <= l1 AND MAX.col1 >= l1)
-   * - f(expr1 AND expr2) -> f(expr1) AND f(expr2)
+   * We only handle the case when `Column` or `Literal` value is the child of binary comparator.
+   * Because they are the only two leaf expression types.
```
scottsand-db, line 200: With all of these complex params, it isn't clear what precisely is being built here. You should add an example.

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,58 +168,187 @@ private[internal] object DataSkippingUtils {
...
+            new GreaterThanOrEqual(maxCol, e2))
 
-          Some(new And(
-              new LessThanOrEqual(minColumn, e2),
-              new GreaterThanOrEqual(maxColumn, e2)))
-        case _ => None
-      }
-      case Some(and: And) =>
+        // (l1 == col1) -> (col1 == l1)
+        val lcRule = (e1: Literal, e2: Column) => new EqualTo(e2, e1)
```
scottsand-db, line 263: `e1` and `e2` aren't the most readable variable names. why not use `lit` and `col` and make the code super readable?

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,58 +168,187 @@ private[internal] object DataSkippingUtils {
...
+        // (col1 == col2) -> (MIN.col1 <= MAX.col2 AND MAX.col1 >= MIN.col2)
+        val ccRule = (e1Min: Column, e1Max: Column, e2Min: Column, e2Max: Column) =>
+          new And(
+            new LessThanOrEqual(e1Min, e2Max),
+            new GreaterThanOrEqual(e1Max, e2Min))
+        buildBinaryComparatorFilter(dataSchema, eq, ccRule, clRule, lcRule)
+
+      case lt: LessThan =>
+        // (col1 < l1) -> (MIN.col1 < l1)
+        val clRule = (minCol: Column, _: Column, e2: Literal) =>
```
scottsand-db, line 274: yeah, why not make `clRule` and `lcRule` actual types?
```scala
private type ColLitRule = ....
```

scottsand-db: then you can just do `val clRule: ColLitRule = ....` and everything is typed nicely.

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,58 +168,187 @@ private[internal] object DataSkippingUtils {
...
+        val ccRule = (_: Column, e1Max: Column, e2Min: Column, _: Column) =>
+          new GreaterThanOrEqual(e1Max, e2Min)
+        buildBinaryComparatorFilter(dataSchema, geq, ccRule, clRule, lcRule)
+
+      case and: And =>
         val e1 = constructDataFilters(dataSchema, Some(and.getLeft))
         val e2 = constructDataFilters(dataSchema, Some(and.getRight))
-
         (e1, e2) match {
+          // f(expr1 AND expr2) -> f(expr1) AND f(expr2)
```
scottsand-db, line 328: if you reference "partial and" in another test, and you had to link me to this implementation here, then you should add a comment here explaining how you are implementing "partial and"

standalone/src/test/scala/io/delta/standalone/internal/DataSkippingSuite.scala
```diff
@@ -332,15 +332,15 @@ class DataSkippingSuite extends FunSuite {
   }
 
   /**
-   * Filter: (col1 <= 1)
+   * Filter: IsNull(col1)
    * Column stats filter: None
    * Output: All files.
-   * Reason: Because LessThanOrEqual is currently unsupported in building column stats predicate,
+   * Reason: Because `IsNull` is currently unsupported in building column stats predicate,
```
scottsand-db, line 338: will we ever support `IsNull`? What is the reason for not supporting it now?

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
6a0juu: Added parameter description for anonymous function.

standalone/src/test/scala/io/delta/standalone/internal/DataSkippingSuite.scala
6a0juu: Changed this case to `IN`, we don't support `IN` because `IN`-list is not ordered so we cannot get the MIN and MAX value from it and cannot build the column stats filter. Please see [this](https://docs.google.com/document/d/1D2ukbGInriKTOZJl3r3pCZoQRji40rSuFgnOyzhHz7k/edit#heading=h.7wihkzdo7ims).

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -45,6 +45,26 @@ private[internal] object DataSkippingUtils {
   final val supportedDataType = Seq(new BooleanType, new ByteType, new DoubleType,
     new FloatType, new IntegerType, new LongType, new ShortType)
 
+  /**
+   * The building rule when left and right children are both columns. The 4 parameters from left to
+   * right are: MIN of left child, MAX of left child, MIN of right child, MAX of right child.
+   */
+  private type ColColRule = (Column, Column, Column, Column) => Expression
```
scottsand-db, line 52: Recall my comment here: https://github.com/delta-io/connectors/pull/413#discussion_r932562721

Personally, I think that using a `case class` is better than having to document the left min, right min, left max, etc etc.

private case class `ColColWrapper(leftMin: Column, leftMax: Column, ...`

Now this doesn't need any comments, because the code is self describing and clear

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,58 +188,198 @@ private[internal] object DataSkippingUtils {
...
-              new LessThanOrEqual(minColumn, e2),
-              new GreaterThanOrEqual(maxColumn, e2)))
-        case _ => None
-      }
-      case Some(and: And) =>
+        // (lit1 == col1) -> (col1 == lit1)
+        val litColRule = (lit1: Literal, col1: Column) => new EqualTo(col1, lit1)
+
+        // (col1 == col2) -> (MIN.col1 <= MAX.col2 AND MAX.col1 >= MIN.col2)
+        val colColRule = (e1Min: Column, e1Max: Column, e2Min: Column, e2Max: Column) =>
```
scottsand-db, line 297: Do you notice a lot of code duplication with all of this long tuple declarations?
Using case classes will solve that

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,58 +188,198 @@ private[internal] object DataSkippingUtils {
...
+  def buildBinaryComparatorFilter(
+      dataSchema: StructType,
+      expr: BinaryComparison,
+      colColRule: ColColRule,
+      colLitRule: ColLitRule,
+      litColRule: LitColRule): Option[Expression] = {
+    (expr.getLeft, expr.getRight) match {
+      case (e1: Column, e2: Column) =>
+        // Apply `colColRule` with MIN and MAX from both children.
+        val leftMinMaxCol = getMinMaxColumn(dataSchema, e1.name).getOrElse { return None }
```
scottsand-db, line 247:
```scala
val (leftMin, leftMax) = getMinMaxColumn(dataSchema, e1.name).getOrElse { return None }
```

would be much clearer.

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,58 +188,198 @@ private[internal] object DataSkippingUtils {
...
+        val rightMinMaxCol = getMinMaxColumn(dataSchema, e2.name).getOrElse { return None }
+        Some(colColRule(leftMinMaxCol._1, leftMinMaxCol._2, rightMinMaxCol._1, rightMinMaxCol._2))
```
scottsand-db, line 249: that way instead of `._1` and `._2` you can just use `leftMin` etc

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
scottsand-db: btw if you use this case class, you probably don't need to define the type ColColRule.
Since `ColColWrapper => Expression` is easy enough.

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
scottsand-db: so this would be
```scala
val colColRule = (w: ColColWrapper) => ... ... `w.leftMin, w.rightMax`
```
etc etc

scottsand-db: and this has much less code duplication

scottsand-db (changes requested):

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -45,6 +45,28 @@ private[internal] object DataSkippingUtils {
   final val supportedDataType = Seq(new BooleanType, new ByteType, new DoubleType,
     new FloatType, new IntegerType, new LongType, new ShortType)
 
+  /**
+   * The building rule when left and right children are both columns.
+   * The 4 parameters from left to right are: MIN of left child, MAX of left child,
+   * MIN of right child, MAX of right child.
```
scottsand-db, line 51: you don't need these 2 lines. the code is self-describing

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -45,6 +45,28 @@ private[internal] object DataSkippingUtils {
...
+   */
+  private case class ColColRule(
+      leftMin: Column, leftMax: Column, rightMin: Column, rightMax: Column)
+
+  /**
+   * The building rule when left child is a column and right child is a literal value.
+   * The 3 parameters from left to right are: MIN of left child, MAX of left child,
+   * the literal value at the right child.
```
scottsand-db, line 59: ditto

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -45,6 +45,28 @@ private[internal] object DataSkippingUtils {
...
+   */
+  private case class ColLitRule(leftMin: Column, leftMax: Column, right: Literal)
+
+  /**
+   * The building rule when left child is a literal value and right child is a column.
+   * The 2 parameters from left to right are: the literal value at the left child,
+   * the column at the right child.
```
scottsand-db, line 66: ditto

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,58 +190,190 @@ private[internal] object DataSkippingUtils {
...
-   * - f(expr1 AND expr2) -> f(expr1) AND f(expr2)
+   * For example, `col1` is a column and `lit1` is a literal value. When building column stats
+   * predicate with query predicate `col1 < lit1`. This method will follow the `colLitRule` because
+   * left child is `col1` and right child is `lit1`. It will make the `MIN.col1` and `MAX.col1`,
+   * then pass these columns and the literal value `lit1` into the `clRule` (For operator `<`, the
+   * rule is `${MIN.column} < ${literal}`). So the output expression is `MIN.col1 < lit1`.
+   *
+   * @param dataSchema The schema of data columns in table.
+   * @param expr       The expression from query predicate.
+   * @param colColRule The building rule when left and right children are both columns.
```
scottsand-db, line 232: `colColRule` and `ColColRule` is a little weird. One is a case class, the other is a function from that case class to `Expression`. This is inconsistent.

In [this](https://github.com/delta-io/connectors/pull/413#discussion_r932729748) comment I suggested `ColColWrapper`.

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,58 +190,190 @@ private[internal] object DataSkippingUtils {
...
+        // (col1 == col2) -> (MIN.col1 <= MAX.col2 AND MAX.col1 >= MIN.col2)
+        val colColRule = (r: ColColRule) =>
+          new And(
+            new LessThanOrEqual(r.leftMin, r.rightMax),
+            new GreaterThanOrEqual(r.leftMax, r.rightMin))
+        buildBinaryComparatorFilter(dataSchema, eq, colColRule, colLitRule, litColRule)
+
+      case lt: LessThan =>
+        // (col1 < lit1) -> (MIN.col1 < lit1)
+        val colLitRule = (r: ColLitRule) => new LessThan(r.leftMin, r.right)
```
scottsand-db, line 307: much cleaner, eh :)

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -45,6 +45,28 @@ private[internal] object DataSkippingUtils {
...
+   */
+  private case class ColColRule(
+      leftMin: Column, leftMax: Column, rightMin: Column, rightMax: Column)
+
+  /**
+   * The building rule when left child is a column and right child is a literal value.
+   * The 3 parameters from left to right are: MIN of left child, MAX of left child,
+   * the literal value at the right child.
+   */
+  private case class ColLitRule(leftMin: Column, leftMax: Column, right: Literal)
```
scottsand-db, line 61: why not use `lit` instead of `right`?

scottsand-db (changes requested):

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,58 +178,190 @@ private[internal] object DataSkippingUtils {
...
+        Some(colColRule(ColColWrapper(leftMin, leftMax, rightMin, rightMax)))
+      case (e1: Column, e2: Literal) =>
+        // Apply `colLitRule` with MIN and MAX of the column at left child and the literal value
+        // at the right child.
+        val (leftMin, leftMax) = getMinMaxColumn(dataSchema, e1.name).getOrElse { return None }
+        Some(colLitRule(ColLitWrapper(leftMin, leftMax, e2)))
+      case (e1: Literal, e2: Column) =>
+        // Apply `litColRule` by swap the left and right child. Then `constructDataFilters` will use
+        // `colLitRule` to solve this problem.
+        constructDataFilters(dataSchema, Some(litColRule(LitColWrapper(e1, e2))))
```
scottsand-db, line 248: you still haven't answered my question here: https://github.com/delta-io/connectors/pull/413#discussion_r932563235.

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,58 +178,190 @@ private[internal] object DataSkippingUtils {
...
+        // (lit1 <= col1) -> (col1 >= lit1)
+        val litColRule = (r: LitColWrapper) => new GreaterThanOrEqual(r.col, r.lit)
+
+        // (col1 <= col2) -> (MIN.col1 <= MAX.col2)
+        val colColRule = (r: ColColWrapper) => new LessThanOrEqual(r.leftMin, r.rightMax)
+        buildBinaryComparatorFilter(dataSchema, leq, colColRule, colLitRule, litColRule)
+
+      case geq: GreaterThanOrEqual =>
+        // (col1 >= lit1) -> (MAX.col1 >= lit1)
+        val colLitRule = (r: ColLitWrapper) => new GreaterThanOrEqual(r.leftMax, r.lit)
```
scottsand-db, line 328: It's not clear to me why the order of `(col, lit)` or `(lit, col)` means that one of them uses the `max` (e.g. here you use r.leftMax), but the other does not (e.g. below you just use (r.col, r.lit)

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -170,58 +178,190 @@ private[internal] object DataSkippingUtils {
...
+      colColRule: ColColWrapper => Expression,
+      colLitRule: ColLitWrapper => Expression,
+      litColRule: LitColWrapper => Expression): Option[Expression] = {
+    (expr.getLeft, expr.getRight) match {
+      case (e1: Column, e2: Column) =>
+        // Apply `colColRule` with MIN and MAX from both children.
+        val (leftMin, leftMax) = getMinMaxColumn(dataSchema, e1.name).getOrElse { return None }
+        val (rightMin, rightMax) = getMinMaxColumn(dataSchema, e2.name).getOrElse { return None }
+        Some(colColRule(ColColWrapper(leftMin, leftMax, rightMin, rightMax)))
+      case (e1: Column, e2: Literal) =>
```
scottsand-db, line 240: nit: using `col` and `e2` as variable names will make this more readable than `e1` and `e2`. little things like this make the code easier to read and maintain for future developers.

scottsand-db: This is very similar to my comment here: https://github.com/delta-io/connectors/pull/413#discussion_r932567764

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
6a0juu: In `LitColRule`, we cast `(Literal, Column)` to `(Column, Literal)` format, with some expression type change. Also I have added the comment [here](https://github.com/delta-io/connectors/pull/413/files#diff-ac21cc1184f49d8ec1b628b5106c84db06443f287ae0dfa16cf8aac5302be5f9R246)

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
6a0juu: In `LitColRule`, we cast `(Literal, Column)` to `(Column, Literal)` format, with some expression type change. Also I have added the comment [here](https://github.com/delta-io/connectors/pull/413/files#diff-ac21cc1184f49d8ec1b628b5106c84db06443f287ae0dfa16cf8aac5302be5f9R246)

scottsand-db: you state `by swap the left and right child` - where is this swap occurring?

scottsand-db: In other words, can you explain this to me please:
In my opinion, if we have `case (c: Column, l: Literal) => doSomething(c, l)` then when we have `case (l: Literal, c: Column)` it should perform `doSomething(c, l)`. Right now, your two different cases, which only change by the order of params, seem to be doing very very different things, and it's not clear to me why.

scottsand-db: You also stated, `In LitColRule, we cast (Literal, Column) to (Column, Literal) format` ... but `litColRule` is defined as a `The building rule when left child is a literal value and right child is a`. It doesn't mention anything about swapping.

Let me know if my position isn't clear enough and you'd like me to clarify something. Also, feel free to msg me and we can book a time to call and discuss this design. Cheers.

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
6a0juu: Please see [here](https://github.com/delta-io/connectors/pull/413/files#diff-ac21cc1184f49d8ec1b628b5106c84db06443f287ae0dfa16cf8aac5302be5f9R247)

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
6a0juu: Swapped the parameter order of `litColRule`. And I have updated the comments to 
```
// The `litColRule` will transform a binary expression from `expr(lit, col)` to
// `new_expr(col, lit)`. Afterwards `constructDataFilters` will use `colLitRule` to build
// the result from `new_expr(col, lit)` with `colLitRule`.
```

scottsand-db: @6a0juu - I'd suggest responding/answering the questions/comments above, instead of just pushing more code :)

For example, the question I asked you to explain [here](https://github.com/delta-io/connectors/pull/413#discussion_r933597732) is still not clear to me.

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
```diff
@@ -52,8 +52,13 @@ private[internal] object DataSkippingUtils {
...
 
-  /** The building rule when left child is a literal value and right child is a column. */
-  private case class LitColWrapper(lit: Literal, col: Column)
+  /**
+   * The building rule when left child is a literal value and right child is a column. The
+   * `litColRule` will transform a binary expression from `expr(lit, col)` to `new_expr(col, lit)`.
+   * Afterwards `constructDataFilters` will use `colLitRule` to build the result from
+   * `new_expr(col, lit)` with `colLitRule`.
+   */
+  private case class LitColWrapper(col: Column, lit: Literal)
```
scottsand-db, line 61: why this change? the order of the parameters before made more sense, given that the class is called `LitColWrapper`, and not `ColLitWrapper`. If there was a past comment of mine that caused this or confused you, please feel free to call it out and ask for clarification.

standalone/src/main/scala/io/delta/standalone/internal/util/DataSkippingUtils.scala
scottsand-db: For example, I don't see how swapping the parameter order of `litColRule` resolves the above discussion.

6a0juu: Thanks for the advice @scottsand-db ! I have updated the [class doc](https://github.com/delta-io/connectors/pull/413/files#diff-ac21cc1184f49d8ec1b628b5106c84db06443f287ae0dfa16cf8aac5302be5f9R259) with the relation ship of `buildBinaryComparatorFilter` and `constructDataFilters`, how the different rules handling different expression types, and how `litColRule` rewritten the expression and call `constructDataFilters` again to build the column stats filter.

6a0juu: For your question about how `litColRule` works, my answer is:
When left child is literal value and right child is column, we use `litColRule`. This will rewrite the expression to `colLitRule` format (the left child is column and right child is literal value) in `buildBinaryComparatorFilter` and then use the rewritten expression to call `constructDataFilters` again to build the target expression by `colLitRule`.

vkorukanti:

PR #414, opened 2022-07-27
Minor fixes integration tests

scottsand-db:
- update the `examples/build.sbt` to no longer depend on `parquet-hadoop` (that dependency is shaded in delta-standalone. we only need to provide it when we use the `ParquetSchemaConverter` class).
- we also add logging support inside of `examples/build.sbt`
- minor fix to flink example util, which now correctly lets us run the SBT command given in the flink example [readme](https://github.com/delta-io/connectors/blob/master/examples/flink-example/README.md). Without this change, it was looking side of `examples/examples/flink-example` instead of inside `examples/flink-example`.

### DIFF
--- a/examples/build.sbt
+++ b/examples/build.sbt
@@ -26,8 +26,7 @@ lazy val commonSettings = Seq(
   resolvers += Resolver.mavenLocal,
   libraryDependencies ++= Seq(
     "io.delta" %% "delta-standalone" % getStandaloneVersion(),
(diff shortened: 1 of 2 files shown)


allisonport-db (approved):
lgtm

allisonport-db:
I'm assuming you ran the flink example with the staged artifacts? I can't do it without rebasing since it includes DeltaSource in the example

scottsand-db:
@allisonport-db 

> I'm assuming you ran the flink example with the staged artifacts?

Yup.

PR #415, opened 2022-07-27
Update delta-flink javadocs and README.md

scottsand-db:
- Clarify in the README that the continuous delta source can either load the full table state and then start streaming changes, or just stream changes.
- Add delta options to the README
- Various javadoc improvements.

### DIFF
--- a/flink/README.md
+++ b/flink/README.md
@@ -173,7 +173,7 @@ The options relevant to this mode are
 - `startingTimestamp` - Starts reading changes from the table version written at or after the given timestamp.
 - `updateCheckIntervalMillis` - The interval, in milliseconds, at which we will check the underlying Delta table for any changes.
 - `ignoreDeletes` - When set to `true`, the Delta Source will be able to process table versions where data is deleted, and skip those deleted records.
-- `ignoreChanges` - When set to `true`, the Delta Source will be able to process table versions where data is changed (i.e. updated), and return those changed records. Note that this can lead to duplicate processing, as some Delta operations, like `UPDATE`, cause existing rows to be rewritten in new files. Those new files will be treated as new data and be reprocessed. This options subsumes `ignoreDeletes`. Therefore, if you set `ignoreChanges` to `true`, your stream will not be disrupted by either deletions or updates to the source table.
+- `ignoreChanges` - When set to `true`, the Delta Source will be able to process table versions where data is changed (i.e. updated), and return those changed records. Note that this can lead to duplicate processing, as some Delta operations, like `UPDATE`, may cause existing rows to be rewritten in new files. Those new files will be treated as new data and be reprocessed. This options subsumes `ignoreDeletes`. Therefore, if you set `ignoreChanges` to `true`, your stream will not be disrupted by either deletions or updates to the source table.
 - `columnNames` - Which columns to read. If not provided, the Delta Source source will read all columns.
 
 #### Table schema discovery
--- a/flink/README.md
+++ b/flink/README.md
@@ -14,9 +14,7 @@ Official Delta Lake connector for [Apache Flink](https://flink.apache.org/).
 - [Delta Source](#delta-source)
   - [Modes](#modes)
     - [Bounded Mode](#bounded-mode)
-      - [Bounded Mode Options](#bounded-mode-options)
     - [Continuous Mode](#continuous-mode)
-      - [Continuous Mode Options](#continuous-mode-options)
   - [Examples](#delta-source-examples)
 - [Usage](#usage)
   - [Maven](#maven)
@@ -160,8 +158,7 @@ The `DeltaSource` class provides factory methods to create sources for both mode
 ### Bounded Mode
 Suitable for batch jobs, where we want to read content of Delta table for specific table version only. Create a source of this mode using the `DeltaSource.forBoundedRowData` API.
 
-#### Bounded Mode Options
-The relevant options for this mode are
+The options relevant to this mode are
 - `versionAsOf` - Loads the state of the Delta table at that version.
 - `timestampAsOf` - Loads the state of the Delta table at the table version written at or before the given timestamp.
 - `columnNames` - Which columns to read. If not provided, the Delta Source source will read all columns.
@@ -171,8 +168,7 @@ Suitable for streaming jobs, where we want to continuously check the Delta table
 
 Note that by default, the Delta Source will load the full state of the latest Delta table, and then start streaming changes. When you use the `startingTimestamp` or `startingVersion` APIs on the `ContinuousDeltaSourceBuilder`, then the Delta Source will process changes only from that corresponding historical version.
 
-#### Continuous Mode Options
-The relevant options for this mode are
+The options relevant to this mode are
 - `startingVersion` - Starts reading changes from this table version.
 - `startingTimestamp` - Starts reading changes from the table version written at or after the given timestamp.
 - `updateCheckIntervalMillis` - The interval, in milliseconds, at which we will check the underlying Delta table for any changes.
--- a/flink/README.md
+++ b/flink/README.md
@@ -13,6 +13,10 @@ Official Delta Lake connector for [Apache Flink](https://flink.apache.org/).
   - [Examples](#delta-sink-examples)
 - [Delta Source](#delta-source)
   - [Modes](#modes)
+    - [Bounded Mode](#bounded-mode)
+      - [Bounded Mode Options](#bounded-mode-options)
+    - [Continuous Mode](#continuous-mode)
+      - [Continuous Mode Options](#continuous-mode-options)
   - [Examples](#delta-source-examples)
 - [Usage](#usage)
   - [Maven](#maven)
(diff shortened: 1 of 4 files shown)


flink/src/main/java/io/delta/flink/source/RowDataContinuousDeltaSourceBuilder.java
```diff
@@ -102,10 +105,8 @@ public RowDataContinuousDeltaSourceBuilder startingVersion(long startingVersion)
     }
 
     /**
-     * Sets value of "startingTimestamp" option. Applicable for {@link
-     * org.apache.flink.api.connector.source.Boundedness#CONTINUOUS_UNBOUNDED} mode only. This
-     * option is used to read only changes from {@link io.delta.standalone.Snapshot} that was
-     * generated at or before given timestamp.
+     * Sets value of "startingTimestamp" option. This option is used to read only changes from the
+     * {@link io.delta.standalone.Snapshot} that was generated at or before given timestamp.
```
allisonport-db, line 109: Is this not going to read changes made to the table at or after such timestamp?
"This option is used to read only changes from the Snapshot ..." makes me think it's reading changes in that snapshot which is confusing

allisonport-db (approved):
lgtm besides the 1 comment. thank you this is much better and clearer to me

flink/src/main/java/io/delta/flink/source/RowDataContinuousDeltaSourceBuilder.java
scottsand-db: Good point.

scottsand-db: Fixed it

flink/src/main/java/io/delta/flink/source/DeltaSource.java
```diff
@@ -12,61 +12,59 @@
...
- * To create a new instance of Delta source for a non-partitioned Delta table that will produce
- * {@link RowData} records that will contain all columns from Delta table:
+ * To create a new instance of {@link DeltaSource} for a Delta table that will produce
+ * {@link RowData} records that contain all table columns:
  * <pre>
  *     StreamExecutionEnvironment env = StreamExecutionEnvironment.getExecutionEnvironment();
  *     ...
  *     // Bounded mode.
- *     DeltaSource&lt;RowData&gt; deltaSink = DeltaSource.forBoundedRowData(
+ *     DeltaSource&lt;RowData&gt; deltaSource = DeltaSource.forBoundedRowData(
```
tdas, line 27: lol. funny that we had missed this in the earlier review.

flink/src/main/java/io/delta/flink/source/DeltaSource.java
```diff
@@ -75,18 +73,21 @@
...
  * <p>This source supports both bounded/batch and continuous/streaming modes. For the
- * bounded/batch case, the Delta Source processes all {@link AddFile} from Delta table Snapshot. In
- * the continuous/streaming case, the source periodically checks the Delta Table for any appending
- * changes and reads them.
+ * bounded/batch case, the Delta Source processes the full state of the Delta table. In
+ * the continuous/streaming case, the default Delta Source will also process the full state of the
+ * table, and then begin to periodically check the Delta table for any appending changes and read
+ * them. Using either of the {@link RowDataContinuousDeltaSourceBuilder#startingVersion} or
+ * {@link RowDataContinuousDeltaSourceBuilder#startingTimestamp} APIs will cause the Delta Source,
+ * in continuous mode, to stream only the changes from that historical version.
```
tdas, line 90: love it. this was super necessary.  i think this must be clarified clearly in the markdown docs as well.. not just in the java docs.

flink/src/main/java/io/delta/flink/source/RowDataBoundedDeltaSourceBuilder.java
```diff
@@ -150,8 +146,8 @@ public RowDataBoundedDeltaSourceBuilder option(String optionName, long optionVal
 
     /**
      * Creates an instance of {@link DeltaSource} for a stream of {@link RowData}. Created source
-     * will work in Bounded mode, meaning it will read content of configured Delta's snapshot
-     * version ignoring all changes done to this table after starting this source.
+     * will work in Bounded mode, meaning it will read the content of the configured Delta Snapshot
```
tdas, line 149:
```suggestion
     * will work in Bounded mode, meaning it will read the content of the configured Delta snapshot
```
snapshot is a general concept. It happens to be an internal class name but the user does not need to know that. so the public doc should refer to the concept with lower "s"

flink/src/main/java/io/delta/flink/source/RowDataContinuousDeltaSourceBuilder.java
```diff
@@ -145,34 +144,24 @@ public RowDataContinuousDeltaSourceBuilder updateCheckIntervalMillis(
...
-     * {@link org.apache.flink.api.connector.source.Boundedness#CONTINUOUS_UNBOUNDED}
-     * mode only. This option allows processing Delta table versions containing only {@link
-     * io.delta.standalone.actions.RemoveFile} actions.
-     *
-     * <p> If this option is set to true, Source connector will not throw an exception when
-     * processing version containing only {@link io.delta.standalone.actions.RemoveFile} actions
-     * regardless of {@link io.delta.standalone.actions.RemoveFile#isDataChange()} flag.
-     *
+     * Sets the "ignoreDeletes" option. This option allows processing Delta table versions where
+     * data is deleted.
```
tdas, line 148: why are you remove the explanation of what will happen when this is set to false.
also .. when its set to true... this does not explain what happens. ... what is the user impact. what does allow processing means. for example, it can lead to duplicate process of data as the existing rows are rewritten in new files and the new files will be treated as fully new data and reprocessed.
(i did not know write this well, so rephrase, but this is important for both java docs and markdown.

flink/src/main/java/io/delta/flink/source/RowDataContinuousDeltaSourceBuilder.java
```diff
@@ -145,34 +144,24 @@ public RowDataContinuousDeltaSourceBuilder updateCheckIntervalMillis(
...
 
     /**
-     * Sets the "ignoreChanges" option. Applicable for
-     * {@link org.apache.flink.api.connector.source.Boundedness#CONTINUOUS_UNBOUNDED}
-     * mode only. This option allows processing Delta table versions containing both {@link
-     * io.delta.standalone.actions.RemoveFile} and {@link io.delta.standalone.actions.AddFile}
-     * actions. This option subsumes {@link #ignoreDeletes} option.
+     * Sets the "ignoreChanges" option. This option allows processing Delta table versions where
+     * data is changed (i.e. updated), or deleted. This option subsumes {@link #ignoreDeletes}
+     * option. Therefore if you use ignoreChanges, your stream will not be disrupted by either
```
tdas, line 160: "use ignoreChanges" -> "set ignoreChanges to true"

same comment as above... explain the impact.

tdas (approved):

flink/src/main/java/io/delta/flink/source/RowDataContinuousDeltaSourceBuilder.java
scottsand-db: Existing rows are rewritten in new files --> that means there are AddFiles, which `ignoreChanges` is for. This is `ignoreDeletes`.

scottsand-db: I will update `ignoreChanges` docs, though.

flink/README.md
```diff
@@ -149,12 +153,32 @@ public DataStream<RowData> createDeltaSink(
...
+Delta Source can work in one of two modes, described below.
 
-The `DeltaSource` class provides factory methods to create sources for both modes.
-Please see [documentation](https://delta-io.github.io/connectors/latest/delta-flink/api/java/index.html) and examples for details.
+The `DeltaSource` class provides factory methods to create sources for both modes. Please see [documentation](https://delta-io.github.io/connectors/latest/delta-flink/api/java/index.html) and examples for details.
+
+### Bounded Mode
+Suitable for batch jobs, where we want to read content of Delta table for specific table version only. Create a source of this mode using the `DeltaSource.forBoundedRowData` API.
+
+#### Bounded Mode Options
```
tdas, line 163: this does not need to be subsection of its own. for literary perspective, its weird to have just 1 subsection

flink/README.md
```diff
@@ -149,12 +153,32 @@ public DataStream<RowData> createDeltaSink(
...
+The relevant options for this mode are
```
tdas, line 164:
```suggestion
The options relevant to this mode are
```

flink/README.md
scottsand-db: sgtm

flink/src/main/java/io/delta/flink/source/RowDataContinuousDeltaSourceBuilder.java
```diff
@@ -145,34 +142,28 @@ public RowDataContinuousDeltaSourceBuilder updateCheckIntervalMillis(
...
-     *
-     * <p> If this option is set to true, Source connector will not
-     * throw an exception when processing version containing combination of {@link
-     * io.delta.standalone.actions.RemoveFile} and {@link io.delta.standalone.actions.AddFile}
-     * actions regardless of {@link io.delta.standalone.actions.RemoveFile#isDataChange()} flag.
+     * Sets the "ignoreChanges" option. When set to true, this option allows processing Delta table
+     * versions where data is changed (i.e. updated) or deleted.
+     * <p>
+     * Note that setting this option to true can lead to duplicate processing of data, as, in the
+     * case of updates, existing rows are rewritten in new files, and those new files will be
```
tdas, line 160:
```suggestion
     * case of updates, existing rows may be rewritten in new files, and those new files will be
```

tdas (approved):

mattfysh:
@scottsand-db it looks like the first FAQ can now be updated

PR #417, opened 2022-07-28
Move javadocs from 0.5.0-SNAPSHOT into 0.5.0, and then 0.5.0 into latest

scottsand-db:


allisonport-db:
does this include the changes in #415? why do I only see renamed for 0.5.0-snapshot --> 0.5.0

scottsand-db:
> does this include the changes in https://github.com/delta-io/connectors/pull/415?

yes it does. that's why we are only renaming 0.5.0-SNAPSHOT to 0.5.0

allisonport-db (approved):

Issue #399, opened 2022-07-13
Performance Improvement: cache per-partition results during partition pruning


scottsand-db:
We perform partition pruning inside of [FilteredDeltaScanImpl](https://github.com/delta-io/connectors/blob/master/standalone/src/main/scala/io/delta/standalone/internal/scan/FilteredDeltaScanImpl.scala) by
- having some input query / expression `expr`
- creating a [PartitionRowRecord](https://github.com/delta-io/connectors/blob/master/standalone/src/main/scala/io/delta/standalone/internal/data/PartitionRowRecord.scala) for a given AddFile `partitionRowRecord`
- evaluating `partitionRowRecord` against `expr`

The result of this evaluation is a function of the input `expr`(fixed), the `partitionSchema` (fixed), and the AddFile partition values (variable per AddFile). We perform this evaluation [here](https://github.com/delta-io/connectors/blob/master/standalone/src/main/scala/io/delta/standalone/internal/scan/FilteredDeltaScanImpl.scala#L48).

Well, many AddFiles will have the same partition values. Thus, we can easily cache and save the evaluation result per unique partition values, and save on some computation.

sonhmai:
@scottsand-db I would like to contribute. can you assign this issue to me?

scottsand-db:
@sonhmai - great to hear! I will assign it to you. Cheers.

sonhmai:
@scottsand-db 
I assumed that the number of unique combinations of partition values can be contained in memory.

Hence, I'm thinking about just adding a `mutable.Map[Map[String, String], Boolean]` in [FilteredDeltaScanImpl](https://github.com/delta-io/connectors/blob/master/standalone/src/main/scala/io/delta/standalone/internal/scan/FilteredDeltaScanImpl.scala) to cache the unique `result` of unique partitionValues combination (the immutable map from `addFile.partitionValues`. What do you think?

scottsand-db:
Hi @sonhmai - that SGTM. @tdas - I don't expect any memory issues here, unless the user has > hundreds of thousands of partitions.

sonhmai:
@scottsand-db should this issue be closed as the MR was merged?

PR #419, opened 2022-07-30
Cache the partition pruning result in FilteredDeltaScanImpl to improve performance

sonhmai:
Cache the partition pruning result in FilteredDeltaScanImpl to improve performance (#399)
Signed-off-by: sonhmai <14060682+sonhmai@users.noreply.github.com>

Benchmark result: elapsed times, total 200 scans, 10k AddFile ops
- With Caching: total 6.448519193 secs, avg per scan 32.242595965 ms
- Without Caching: 7.452229266 secs, avg per run 37.26114633 ms
- Caching saves on avg 5.018550365 ms per scan

### DIFF
--- a/standalone/src/main/scala/io/delta/standalone/internal/scan/FilteredDeltaScanImpl.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/scan/FilteredDeltaScanImpl.scala
@@ -43,7 +43,7 @@ final private[internal] class FilteredDeltaScanImpl(
     hadoopConf: Configuration) extends DeltaScanImpl(replay) {
 
   private val partitionColumns = partitionSchema.getFieldNames.toSeq
-  private lazy val evaluationResults = mutable.Map.empty[Map[String, String], Boolean]
+  private val evaluationResults = mutable.Map.empty[Map[String, String], Boolean]
 
   private val (metadataConjunction, dataConjunction) =
     PartitionUtils.splitMetadataAndDataPredicates(expr, partitionColumns)
--- a/standalone/src/main/scala/io/delta/standalone/internal/scan/FilteredDeltaScanImpl.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/scan/FilteredDeltaScanImpl.scala
@@ -18,6 +18,8 @@ package io.delta.standalone.internal.scan
 
 import java.util.Optional
 
+import scala.collection.mutable
+
 import io.delta.standalone.expressions.Expression
 import io.delta.standalone.types.StructType
 
@@ -37,6 +39,7 @@ final private[internal] class FilteredDeltaScanImpl(
     partitionSchema: StructType) extends DeltaScanImpl(replay) {
 
   private val partitionColumns = partitionSchema.getFieldNames.toSeq
+  private val evaluationResults = mutable.Map.empty[Map[String, String], Boolean]
 
   private val (metadataConjunction, dataConjunction) =
     PartitionUtils.splitMetadataAndDataPredicates(expr, partitionColumns)
@@ -44,9 +47,13 @@ final private[internal] class FilteredDeltaScanImpl(
   override protected def accept(addFile: AddFile): Boolean = {
     if (metadataConjunction.isEmpty) return true
 
+    val cachedResult = evaluationResults.get(addFile.partitionValues)
+    if (cachedResult.isDefined) return cachedResult.get
+
     val partitionRowRecord = new PartitionRowRecord(partitionSchema, addFile.partitionValues)
-    val result = metadataConjunction.get.eval(partitionRowRecord)
-    result.asInstanceOf[Boolean]
+    val result = metadataConjunction.get.eval(partitionRowRecord).asInstanceOf[Boolean]
+    evaluationResults(addFile.partitionValues) = result
+    result
   }
 
   override def getInputPredicate: Optional[Expression] = Optional.of(expr)
--- a/standalone/src/main/scala/io/delta/standalone/internal/scan/FilteredDeltaScanImpl.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/scan/FilteredDeltaScanImpl.scala
@@ -54,7 +54,9 @@ final private[internal] class FilteredDeltaScanImpl(
   override protected def accept(addFile: AddFile): Boolean = {
     if (metadataConjunction.isEmpty) return true
 
-    val partitionRowRecord = new PartitionRowRecord(partitionSchema, addFile.partitionValues)
+    // found in micro-benchmarking that eagerly creating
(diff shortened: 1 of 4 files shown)


standalone/src/main/scala/io/delta/standalone/internal/scan/FilteredDeltaScanImpl.scala
```diff
@@ -37,16 +39,21 @@ final private[internal] class FilteredDeltaScanImpl(
     partitionSchema: StructType) extends DeltaScanImpl(replay) {
 
   private val partitionColumns = partitionSchema.getFieldNames.toSeq
+  private val evaluationResults = mutable.Map.empty[Map[String, String], Boolean]
```
sonhmai, line 42: not using concurrent map here to leave control to client if the `Scan` needs to be accessed concurrently. I think the default case which happens most often is that `FilteredDeltaScanImpl` is used by only a single thread?

sonhmai: safely(?) assumed here that the unique partition values Map can fit in memory.

scottsand-db: > not using concurrent map here to leave control to client if the Scan needs to be accessed concurrently. I think the default case which happens most often is that FilteredDeltaScanImpl is used by only a single thread?

this SGTM.

standalone/src/main/scala/io/delta/standalone/internal/scan/FilteredDeltaScanImpl.scala
```diff
@@ -37,16 +39,21 @@ final private[internal] class FilteredDeltaScanImpl(
...
 
   private val (metadataConjunction, dataConjunction) =
     PartitionUtils.splitMetadataAndDataPredicates(expr, partitionColumns)
 
   override protected def accept(addFile: AddFile): Boolean = {
     if (metadataConjunction.isEmpty) return true
 
+    val cachedResult = evaluationResults.get(addFile.partitionValues)
```
scottsand-db, line 50: can you please wrap this behind a feature flag? you can add a hadoop conf constant to https://github.com/delta-io/connectors/blob/master/standalone/src/main/scala/io/delta/standalone/internal/sources/StandaloneHadoopConf.scala.

`partitionFilterRecordCaching.enabled`. Default to true.

When set to false, we will not use this partition filter record caching, presumably when the user is concerned about memory consumption. 

Please include good scaladoc describing the behaviour and usage of this hadoop conf.

As well, could you please add a microbenchmark? You can add a test set to `ignore` (which won't run the test) ... which creates, say, 100 commits, each with 10000 AddFiles, with, say, 100 unique partitions.

Then compare the runtime of this feature with the feature turned on and off.

scottsand-db (changes requested):
Looking good! Just asked for
a) feature flag
b) micro benchmark with the flag turned on/off

standalone/src/main/scala/io/delta/standalone/internal/scan/FilteredDeltaScanImpl.scala
sonhmai: sure, will do

scottsand-db:
@sonhmai - looking good! Could you add a micro benchmark? with perhaps a few layers of partitioning?

standalone/src/main/scala/io/delta/standalone/internal/scan/FilteredDeltaScanImpl.scala
```diff
@@ -34,19 +39,33 @@ import io.delta.standalone.internal.util.PartitionUtils
 final private[internal] class FilteredDeltaScanImpl(
     replay: MemoryOptimizedLogReplay,
     expr: Expression,
-    partitionSchema: StructType) extends DeltaScanImpl(replay) {
+    partitionSchema: StructType,
+    hadoopConf: Configuration) extends DeltaScanImpl(replay) {
 
   private val partitionColumns = partitionSchema.getFieldNames.toSeq
+  private lazy val evaluationResults = mutable.Map.empty[Map[String, String], Boolean]
```
scottsand-db, line 46: nit we can remove `lazy`

sonhmai: thought about if the client disable the caching then the Map does not have to be initialized at all. It does not matter much though. We surely can remove `lazy` to keep things simple :D

sonhmai:
> @sonhmai - looking good! Could you add a micro benchmark? with perhaps a few layers of partitioning?

sure will do this. just pushed the other first

standalone/src/main/scala/io/delta/standalone/internal/scan/FilteredDeltaScanImpl.scala
sonhmai: removed

standalone/src/main/scala/io/delta/standalone/internal/scan/FilteredDeltaScanImpl.scala
sonhmai: maybe later we can have a separate issue to centralize the benchmarks with say JMH in a separate module? wdyt?

sonhmai: @scottsand-db pls review. thanks

standalone/src/main/scala/io/delta/standalone/internal/scan/FilteredDeltaScanImpl.scala
```diff
@@ -34,19 +39,35 @@ import io.delta.standalone.internal.util.PartitionUtils
...
+
   override protected def accept(addFile: AddFile): Boolean = {
     if (metadataConjunction.isEmpty) return true
 
-    val partitionRowRecord = new PartitionRowRecord(partitionSchema, addFile.partitionValues)
-    val result = metadataConjunction.get.eval(partitionRowRecord)
-    result.asInstanceOf[Boolean]
+    // found in micro-benchmarking that eagerly creating
+    // new PartitionRowRecord can destroy the purpose of caching
+    lazy val partitionRowRecord = new PartitionRowRecord(partitionSchema, addFile.partitionValues)
```
sonhmai, line 59: found thru benchmarking that creating this object can defy the purpose of caching, hence `lazy`

scottsand-db (approved):
LGTM! @sonhmai before I merge, can you please add the results of the micro benchmark to the PR description?

sonhmai:
> LGTM! @sonhmai before I merge, can you please add the results of the micro benchmark to the PR description?

@scottsand-db Added to the first comment. If that is not the right place, pls let me know.

PR #425, opened 2022-08-23
Add DeltaConnectorConfiguration options to flink sink impl

### LINKED ISSUE #423: Passing options to delta flink sink
Hi,
I would recommend to do it via `.option(...)`. And the implementation for `.option(...`) for Delta Sink Builder in my opinion should be implemented in separate branch, reviewed and merged before this one. @scottsand-db sounds good to you?

The hadoop conf is needed/used only by delta standalone. There is no interaction between Flink specific classes from Delta connector with hadoop conf and I think we should need to have a good reason to change this.

@gopik
 you can check  `DeltaSourceBuilderBase.java` to see how `.option(...)` was implemented there and do the same thing for `DeltaSinkBuilder`. Heads up, both builder implementations differ in low level details. 

Also please note that .option(...) is overloaded, it accepts boolean, int, long and String as option values. Yours needs to have the same.

As for the feature flag, you need to have a `DeltaSourceOptions.java` like class for Sink and add the flag there as a `DeltaConfigOption` instance. 

A good example to follow would be IGNORE_DELETES option from `DeltaSourceOptions.java`. Having the API for .option(...) you can use the same definition as IGNORE_DELETES has, use same BooleanOptionTypeConverter and it should all be working.

_Originally posted by @kristoffSC in https://github.com/delta-io/connectors/pull/393#discussion_r947278368_

gopik:
Extract option validator from source impl into a different package and reuse it from source and sink builders.
Closes #423


flink/src/main/java/io/delta/flink/options/BaseOptionTypeConverter.java
```diff
@@ -1,4 +1,4 @@
-package io.delta.flink.source.internal.builder;
+package io.delta.flink.options;
```
kristoffSC, line 1: This has to stay under `internal`. 
Generally all classes that were under `internal` have to stay under that package.

the `internal` package is used to filter out classes for which we do not want to generate java docs via sbt-unidoc plugin.

kristoffSC, line 1: This must stay under the internal package.

Generally all classes that were under internal have to stay under that package.

the internal package is used to filter out classes for which we do not want to generate java docs via sbt-unidoc plugin.

kristoffSC:
Quick, small note

Generally all classes that were under the `internal` have to stay under that package.
The `internal` package is used to filter out classes for which we do not want to generate java docs via sbt-unidoc plugin.

I see that you have moved a lot of classes out from that `internal` packed.  That have to stay like it was.

gopik:
I've move this to an internal package under io.delta.flink. Did you mean this or keeping them under respective source.internal and sink.internal (which implies a copy)?

kristoffSC:
> I've move this to an internal package under io.delta.flink. Did you mean this or keeping them under respective source.internal and sink.internal (which implies a copy)?

`io.delta.flink.internal` is great, thanks.

gopik:
@kristoffSC this is now ready for review.

flink/src/main/java/io/delta/flink/internal/options/OptionValidator.java
```diff
@@ -0,0 +1,85 @@
...
+ * Setting of an option is allowed for known option names. For invalid options, the validation
+ * throws {@link IllegalArgumentException}.
+ */
+public class OptionValidator {
+    private final Map<String, DeltaConfigOption<?>> validOptions;
+    private final DeltaConnectorConfiguration config;
+
+    public OptionValidator(
+        DeltaConnectorConfiguration config,
+        Map<String, DeltaConfigOption<?>> validOptions) {
```
scottsand-db, line 19: in general, we use this indentation:
```java
public OptionValidator(
        DeltaConnectorConfiguration config, // 8 space indent
        Map<String, DeltaConfigOption<?>> validOptions) {
    this.config = config; // 4 space indent
    this.validOptions = validOptions;
}
```

flink/src/main/java/io/delta/flink/internal/options/OptionValidator.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.util.Map;
+
+import io.delta.flink.source.internal.DeltaSourceOptions;
+
+/**
+ * Validator for delta source and sink connector configuration options.
+ *
+ * Setting of an option is allowed for known option names. For invalid options, the validation
+ * throws {@link IllegalArgumentException}.
```
scottsand-db, line 11: let's make this even clearer to the user on how it is used: add a comment (class javadoc) saying that each Source and Sink will pass in their own `validOptions` map.

flink/src/main/java/io/delta/flink/sink/RowDataDeltaSinkBuilder.java
```diff
@@ -158,4 +204,5 @@ private BucketAssigner<RowData, String> resolveBucketAssigner() {
             new DeltaPartitionComputer.DeltaRowDataPartitionComputer(rowType, partitionColumns);
         return new DeltaBucketAssigner<>(partitionComputer);
     }
+
```
scottsand-db, line 207: nit: whitespace noise.

flink/src/main/java/io/delta/flink/sink/internal/DeltaSinkBuilder.java
```diff
@@ -149,6 +155,7 @@ private static String generateNewAppId() {
 
     private final OutputFileConfig outputFileConfig;
 
+
```
scottsand-db, line 158: whitespace noise please delete

scottsand-db (changes requested):
This looks great! Thanks for doing this!

Left a few minor comments.

Could you also update the PR description showing how this new API can be used?

Also update the DeltaSinkBuilder javadoc showing an example.

Thanks!

flink/src/main/java/io/delta/flink/internal/options/OptionValidator.java
```diff
@@ -0,0 +1,90 @@
...
+    private final DeltaConnectorConfiguration config;
+
+    public OptionValidator(
+            DeltaConnectorConfiguration config,
+            Map<String, DeltaConfigOption<?>> validOptions) {
+        this.config = config;
+        this.validOptions = validOptions;
+    }
+
+    public void option(String optionName, String optionValue) {
```
kristoffSC, line 30: nit: missing java doc comment that other overloaded methods have.

flink/src/main/java/io/delta/flink/internal/options/OptionValidator.java
```diff
@@ -0,0 +1,90 @@
...
+     * Sets a configuration option.
+     */
+    public void option(String optionName, long optionValue) {
+        tryToSetOption(() -> {
+            DeltaConfigOption<?> configOption = validateOptionName(optionName);
+            configOption.setOnConfig(config, optionValue);
+        });
+    }
+
+    private void tryToSetOption(Executable argument) {
```
kristoffSC, line 67: The same method and validation can be found in `DeltaSourceBuilderBase::tryToSetOption` method

I think it would nice to have a common code for such validation, specially that contract for .option is same for both Sink and Source

flink/src/main/java/io/delta/flink/internal/options/OptionValidator.java
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * This is an internal class meant for connector implementations only.
+ * Usage example (for sink):
+ * <code>
+ *     OptionValidator validator = new OptionValidator(sinkConfig, validSinkOptions);
+ *     // For any option set on the sink, pass it to validator. If it's successful, sinkConfig
+ *     // will be updated with the corresponding option.
+ * </code>
+ */
+public class OptionValidator {
```
kristoffSC, line 19: Hi,
I think that `OptionValidator`  should not have `.option` methods in his API.
What if, someday we would want to add another version of .option method? We would have to propagate it on Builder and Validator.

I would think that `OptionValidation` only need `tryToSetOption` method.

Also I'm not sure about the name vs what it does. 
In Source builder we have the same kind of "validation" against valid option names, we have the same `tryToSetOption` method but all of that is a part of builder hence the "names" fits a little bit better. 

`OptionValidator` class name seems like a broad thing, validate name, values types?

gopik: @kristoffSC I agree that there's a duplication of option() functionality in builder and validator. The idea to extract this in it's own class was to isolate option processing in one place. Assuming we only keep tryToSetOption there, where should we have validateOptionName?

kristoffSC: The alternative would be to have a new abstract class that will be common base for both builders.
This new class could have `.option`, `validateOptionName` and `tryToSetOption` but I not feeling super comfortable to coupling Sink and Source classes with common class hierarchy. Composition is always favorable ;)

gopik: How about moving DeltaSourceValidationException -> DeltaOptionValidationException? There isn't anything source specific in that.

I've another question: I didn't understand why we need tryToSetOption and pass a runnable. It just seems to add another call that catches exception and wraps it. But the validation is already throwing runtime exceptions, so we don't need this right?

gopik: Ok, I now understand why we needed `tryToSetOption`. I've now created exception type for validation errors and it should cover the usage from source builder as well. Please take a look, if this looks ok, I'll update the source impl to use this as well.

scottsand-db:
@gopik this class hierarchy is getting a little complicated. could you please add a class diagram to the PR description showing the inheritance/composition relationship with the new API(s) you are proposing?

flink/src/main/java/io/delta/flink/internal/options/DeltaOptionValidationException.java
```diff
@@ -0,0 +1,39 @@
...
+        this.validationMessages =
+            (validationMessages == null) ? Collections.emptyList() : validationMessages;
+    }
+
+    @Override
+    public String getMessage() {
+
+        String validationMessages = String.join(System.lineSeparator(), this.validationMessages);
+
+        return "Invalid Delta Source definition detected."
```
scottsand-db, line 33: why does this exception say `Source`? mistake?

flink/src/main/java/io/delta/flink/internal/options/OptionValidator.java
```diff
@@ -0,0 +1,112 @@
...
+ * Validator for delta source and sink connector configuration options.
+ *
+ * Setting of an option is allowed for known option names. For invalid options, the validation
+ * throws {@link DeltaOptionValidationException}.
+ *
+ * This is an internal class meant for connector implementations only.
+ * Usage example (for sink):
+ * <code>
+ *     OptionValidator validator = new OptionValidator(sinkConfig, validSinkOptions);
+ *     // For any option set on the sink, pass it to validator. If it's successful, sinkConfig
```
scottsand-db, line 16: "pass it to validator". should they use `.option` or `tryToSetOption`? You should give an example with each.

flink/src/test/java/io/delta/flink/internal/options/OptionValidatorTest.java
```diff
@@ -0,0 +1,132 @@
...
+
+        DeltaConfigOption<?> boolOption =
+            DeltaConfigOption.of(
+                ConfigOptions.key("bool").booleanType().defaultValue(false),
+                Boolean.class
+            );
+        validOptions.put("bool", boolOption);
+
+        assertThrows(DeltaOptionValidationException.class, () -> {
+            validator.option("string", "'");
```
scottsand-db, line 108: can you use a value better than `'`. why not `"value"`

flink/src/test/java/io/delta/flink/internal/options/OptionValidatorTest.java
```diff
@@ -0,0 +1,132 @@
...
+        DeltaConfigOption<?> boolOption =
+            DeltaConfigOption.of(
+                ConfigOptions.key("bool").booleanType().defaultValue(false),
+                Boolean.class
+            );
+        validOptions.put("bool", boolOption);
+
+        // Trying to set a string value for a bool option type should throw.
+        assertThrows(DeltaOptionValidationException.class, () -> {
+            validator.option("", "'");
```
scottsand-db, line 129: same here. this is just awkward to read. `"'"`

flink/src/main/java/io/delta/flink/sink/RowDataDeltaSinkBuilder.java
```diff
@@ -124,6 +139,36 @@ public RowDataDeltaSinkBuilder withPartitionColumns(String... partitionColumns)
         return this;
     }
 
+
+    public RowDataDeltaSinkBuilder option(String optionName, String optionValue) {
```
scottsand-db, line 143: missing javadoc.

flink/src/main/java/io/delta/flink/sink/RowDataDeltaSinkBuilder.java
```diff
@@ -70,6 +73,16 @@ public class RowDataDeltaSinkBuilder {
      */
     private String[] partitionColumns = {};
 
+    /**
+     * Stores sink configuration options.
+     */
+    private final DeltaConnectorConfiguration sinkConfiguration = new DeltaConnectorConfiguration();
```
scottsand-db, line 79: this is not used in this PR, right? i.e. we are not setting any sink configurations.

can you please add a TODO comment here, saying to add and use sink configurations.

scottsand-db (changes requested):

gopik:
> @gopik this class hierarchy is getting a little complicated. could you please add a class diagram to the PR description showing the inheritance/composition relationship with the new API(s) you are proposing?

There's no class hierarchy here. The proposal is to extract the option validation and exception out of source implementation into a common package and reuse it from source and sink builders.

flink/src/main/java/io/delta/flink/internal/options/OptionValidator.java
gopik: tryToSetOption is private. Added .option in javadoc.

flink/src/main/java/io/delta/flink/internal/options/OptionValidator.java
gopik: Will update source implementation in a followup PR.

flink/src/main/java/io/delta/flink/sink/internal/DeltaSinkOptions.java
```diff
@@ -0,0 +1,26 @@
...
+ * their type and default values.
+ */
+public class DeltaSinkOptions {
+    /**
+     * A map of all valid {@code DeltaSinkOptions}. This map can be used for example by {@code
+     * RowDataDeltaSinkBuilder} to do configuration sanity check.
+     */
+    public static final Map<String, DeltaConfigOption<?>> USER_FACING_SINK_OPTIONS =
+        new HashMap<>();
+
```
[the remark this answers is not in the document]
gopik: How about adding this in a followup PR?

flink/src/main/java/io/delta/flink/sink/RowDataDeltaSinkBuilder.java
gopik: There won't be an changes here when we add any sink configurations. This object stores computed options to be used by the sinkbuilder.

flink/src/main/java/io/delta/flink/internal/options/DeltaOptionValidationException.java
gopik: Yes, it was a copy/paste error. fixed.

flink/src/main/java/io/delta/flink/sink/internal/DeltaSinkOptions.java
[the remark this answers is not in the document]
gopik: Let's add that in a followup PR?

flink/src/main/java/io/delta/flink/internal/options/OptionValidator.java
gopik: Will add source changes in followup PR.

flink/src/test/java/io/delta/flink/internal/options/OptionValidatorTest.java
gopik: That was a mistake, intent was to have an empty string. Fixed.

kristoffSC:
@scottsand-db there were no changes to builder class hierarchy. 
For a moment there was a small discussion about creating new abstract class for builders, but I'm my opinion composition proposed and implemented by @gopik is a good choice.

scottsand-db:
@gopik - can you update the PR description? is this PR still incomplete?
also, the way you have added `OptionValidator` to `RowDataDeltaSinkBuilder` is very clean. Why haven't you also then added it to `DeltaSourceBuilderBase`? Seems like that would clean up the code alot.

`DeltaSourceBuilderBase` is even doing the same `tryToSetOption` calls...

kristoffSC:
> @gopik - can you update the PR description? is this PR still incomplete?
> also, the way you have added OptionValidator to RowDataDeltaSinkBuilder is very clean. Why haven't you also then added it to DeltaSourceBuilderBase? Seems like that would clean up the code alot.
> 
> DeltaSourceBuilderBase is even doing the same tryToSetOption calls...

@gopik  cc @scottsand-db 
Yeah I asked for the same thing in comment:
https://github.com/delta-io/connectors/pull/425#discussion_r957365388

I see you suggested to add it in a separate PR but I would strongly suggest to add it this one here and make it right at the beginning. It would not make this PR to grow significantly.
This PR is a good place since you are adding .option do Sink builder.

flink/src/main/java/io/delta/flink/internal/options/OptionValidator.java
```diff
@@ -0,0 +1,128 @@
...
+            tablePath,
+            Collections.singletonList(
+                String.format("Invalid option [%s] used for Delta Connector.",
+                    invalidOption)));
+    }
+
+    /** Exception to throw when there's an error while setting an option. */
+    private static DeltaOptionValidationException optionValidationException(
+        Path tablePath,
+        Exception e) {
```
scottsand-db, line 117: nit: recall we want 8 space indentation here

flink/src/main/java/io/delta/flink/source/internal/builder/DeltaSourceBuilderBase.java
```diff
@@ -77,6 +77,10 @@
      */
     protected final DeltaConnectorConfiguration sourceConfiguration =
         new DeltaConnectorConfiguration();
+    /**
+     * Validates sink configuration options.
```
scottsand-db, line 81: `source` not `sink`

flink/src/main/java/io/delta/flink/internal/options/OptionValidator.java
gopik: Isn't this 8 space indented, I'm seeing 8 spaces? Any intellij idea setting we can use to check this?

flink/src/main/java/io/delta/flink/source/internal/builder/DeltaSourceBuilderBase.java
gopik: Fixed.

flink/src/main/java/io/delta/flink/internal/options/OptionValidator.java
scottsand-db: Your's is
```java
    private static DeltaOptionValidationException optionValidationException(
        Path tablePath,
        Exception e) {
        return new DeltaOptionValidationException(
            tablePath,
            Collections.singletonList(e.getClass() + " - " + e.getMessage())
        );
    }
```
but it should be
```java
    private static DeltaOptionValidationException optionValidationException(
            Path tablePath,
            Exception e) {
        return new DeltaOptionValidationException(
            tablePath,
            Collections.singletonList(e.getClass() + " - " + e.getMessage())
        );
    }
```

scottsand-db: are you still seeing 8 spaces? that would be `return new` was indented 8 spaces, too?

gopik: You are right, my bad I was looking at a wrong line, didn't check the line numbers. Will fix this, in case you are aware of intellij preference, please let me know.

gopik: You are right, my bad, i was looking at a different line. Will fix. Will check if there's any intellij pref that does this by default.

scottsand-db (approved):
LGTM.

scottsand-db:
@kristoffSC - can you take another look / LGTM this PR?

kristoffSC (approved):
LGTM :)

flink/src/test/java/io/delta/flink/sink/internal/DeltaSinkOptionsTest.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.lang.reflect.Field;
+import java.util.HashSet;
+import java.util.Set;
+import static java.lang.reflect.Modifier.isPublic;
+import static java.lang.reflect.Modifier.isStatic;
+
+import io.delta.flink.internal.options.DeltaConfigOption;
+import org.apache.flink.configuration.ConfigOption;
+import org.junit.Test;
```
kristoffSC, line 11: please use Junit5 Test instead Junit4.

We are migrating all tests slowly to Junit5 so every new test should be using that one.

kristoffSC:
Added one small nit about using junit5 instead junit4 in new test class, other than that 
LGTM :) Thanks for your work @gopik

scottsand-db:
Ill merge once @gopik responds to the comments above.

gopik:
@scottsand-db Fixed the import to use junit5.

kristoffSC (approved):

scottsand-db:
Merged!

Issue #222, opened 2021-12-01
DecimalType comparison expressions should be valid between values with decimal types with different scales/precisions


allisonport-db:
Currently our comparison expressions fail with `IllegalArgumentException` for decimal type arguments that don’t have equal precision/scale since we check for equal data types. This should be valid (and is in Spark SQL). This impacts IN and all the binary comparison operators. 
Ex: ` new EqualTo(Literal.of(BigDecimal.valueOf(2).setScale(1)), Literal.of(BigDecimal.valueOf(2).setScale(2)))` throws an illegal argument exception for mismatched data types.

Fix: add method `DataType::equivalent(DataType other)` that defaults to equals, and override it in `DecimalType`. Update BinaryOperator and IN to use the equivalent method.

kination:
@allisonport-db there are overrided 'equals' class in `DataType`. So you want `equivalent` to be act different with this, right?

allisonport-db:
Hi @kination sorry for the delay.

So basically we want to add `DataType::equivalent(DataType other)` which defaults to `equals` and `DecimalType::equivalent(DataType other)` that overrides it. Then instead of using `equals` in the comparisons in our code, we will use `DataType::equivalent(DataType other)` .

Does that clarify things? If you're interested in working on this let me know and I'll go ahead and assign this issue to you.

kination:
@allisonport-db thanks! Please let me take a look on this issue 🙏

kination:
@allisonport-db thanks. 1 more question...
Could you let me know which build command should I use for testing the result after fix?

kination:
@allisonport-db hello.
For test, when I run
```
Literal.of(BigDecimal.valueOf(2).setScale(1))
```

it shows
```
overloaded method value of with alternatives:
[error]   (x$1: Byte)io.delta.standalone.expressions.Literal <and>
[error]   (x$1: java.sql.Timestamp)io.delta.standalone.expressions.Literal <and>
[error]   (x$1: String)io.delta.standalone.expressions.Literal <and>
[error]   (x$1: Short)io.delta.standalone.expressions.Literal <and>
[error]   (x$1: Long)io.delta.standalone.expressions.Literal <and>
[error]   (x$1: Float)io.delta.standalone.expressions.Literal <and>
[error]   (x$1: Double)io.delta.standalone.expressions.Literal <and>
[error]   (x$1: java.math.BigDecimal)io.delta.standalone.expressions.Literal <and>
[...]

PR #426, opened 2022-08-24
Add 'equivalent' for type comparison

kination:
Currently, comparison expressions fail with `IllegalArgumentException` for decimal type arguments that don’t have equal precision/scale. This should be valid (and is in Spark SQL). This impacts IN and all the binary comparison operators.

This change is to add `equivalent` method, to only be used internally by our code to compare data types.

Refer https://github.com/delta-io/connectors/issues/222 for detail.

Signed-off-by: Kwangin Jung <inylove82@gmail.com>

### DIFF
--- a/standalone/src/main/java/io/delta/standalone/expressions/BinaryOperator.java
+++ b/standalone/src/main/java/io/delta/standalone/expressions/BinaryOperator.java
@@ -13,7 +13,7 @@ public abstract class BinaryOperator extends BinaryExpression {
         super(left, right);
         this.symbol = symbol;
 
-        if (!left.dataType().equals(right.dataType())) {
+        if (!left.dataType().equivalent(right.dataType())) {
             throw new IllegalArgumentException("BinaryOperator left and right DataTypes must be the"
                     + " same, found: " + left.dataType().getTypeName() + " " + symbol + " " +
                     right.dataType().getTypeName());
--- a/standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
+++ b/standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
@@ -19,15 +19,12 @@ package io.delta.standalone.internal
 import java.math.{BigDecimal => BigDecimalJ}
 import java.sql.{Date => DateJ, Timestamp => TimestampJ}
 import java.util.{Arrays => ArraysJ, Objects}
-
 import scala.collection.JavaConverters._
-
 import org.scalatest.FunSuite
 
 import io.delta.standalone.data.RowRecord
 import io.delta.standalone.expressions.{Column, _}
 import io.delta.standalone.types._
-
 import io.delta.standalone.internal.actions.AddFile
 import io.delta.standalone.internal.data.PartitionRowRecord
 import io.delta.standalone.internal.util.PartitionUtils
@@ -185,6 +182,12 @@ class ExpressionSuite extends FunSuite {
         testPredicate(predicateCreator(Literal.of(small), Literal.of(small2)), smallSmall)
       }
     }
+
+    // TODO:
+    val decimalComparison = new EqualTo(
+      Literal.of(BigDecimalJ.valueOf(2).setScale(1)),
+      Literal.of(BigDecimalJ.valueOf(2).setScale(2))
+    )
   }
 
   test("null predicates") {
@@ -235,6 +238,21 @@ class ExpressionSuite extends FunSuite {
       (0 to 10).map{Literal.of}.asJava), false)
     testPredicate( new In(Literal.of(10),
       (0 to 10).map{Literal.of}.asJava), true)
+
+    testPredicate(
+      new In(
+        Literal.of(BigDecimalJ.valueOf(2).setScale(1)),
+        List(
+          Literal.of(BigDecimalJ.valueOf(1).setScale(1)),
+          Literal.of(BigDecimalJ.valueOf(2).setScale(1)),
+          Literal.of(BigDecimalJ.valueOf(3).setScale(1)),
+          Literal.of(BigDecimalJ.valueOf(4).setScale(1)),
+          Literal.of(BigDecimalJ.valueOf(5).setScale(1)),
+        ).asJava
+      ),
+        true
+      )
+
   }
 
   private def testLiteral(literal: Literal, expectedResult: Any) = {
--- a/standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
+++ b/standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
@@ -183,11 +183,10 @@ class ExpressionSuite extends FunSuite {
       }
     }
 
-    // TODO:
-    val decimalComparison = new EqualTo(
-      Literal.of(BigDecimalJ.valueOf(2).setScale(1)),
+    testPredicate(new EqualTo(
+      Literal.of(BigDecimalJ.valueOf(1).setScale(2)),
       Literal.of(BigDecimalJ.valueOf(2).setScale(2))
-    )
+    ), false)
   }
 
   test("null predicates") {
--- a/standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
+++ b/standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
@@ -19,12 +19,15 @@ package io.delta.standalone.internal
 import java.math.{BigDecimal => BigDecimalJ}
 import java.sql.{Date => DateJ, Timestamp => TimestampJ}
 import java.util.{Arrays => ArraysJ, Objects}
+
 import scala.collection.JavaConverters._
+
 import org.scalatest.FunSuite
 
 import io.delta.standalone.data.RowRecord
 import io.delta.standalone.expressions.{Column, _}
 import io.delta.standalone.types._
+
 import io.delta.standalone.internal.actions.AddFile
 import io.delta.standalone.internal.data.PartitionRowRecord
 import io.delta.standalone.internal.util.PartitionUtils
@@ -266,11 +269,7 @@ class ExpressionSuite extends FunSuite {
           Literal.of(BigDecimalJ.valueOf(3).setScale(1)),
           Literal.of(BigDecimalJ.valueOf(4).setScale(1)),
           Literal.of(BigDecimalJ.valueOf(5).setScale(1)),
-        ).asJava
-      ),
-        true
-      )
-
+        ).asJava),true)
   }
 
   private def testLiteral(literal: Literal, expectedResult: Any) = {
--- a/standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
+++ b/standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
@@ -183,10 +183,29 @@ class ExpressionSuite extends FunSuite {
       }
     }
 
+    // test BigDecimal with diff value - same scale
     testPredicate(new EqualTo(
       Literal.of(BigDecimalJ.valueOf(1).setScale(2)),
       Literal.of(BigDecimalJ.valueOf(2).setScale(2))
     ), false)
+
+    // test BigDecimal with diff value - diff scale
+    testPredicate(new EqualTo(
+      Literal.of(BigDecimalJ.valueOf(1).setScale(1)),
+      Literal.of(BigDecimalJ.valueOf(2).setScale(2))
+    ), false)
+
+    // test BigDecimal with same value - diff scale
+    testPredicate(new EqualTo(
+      Literal.of(BigDecimalJ.valueOf(2).setScale(1)),
+      Literal.of(BigDecimalJ.valueOf(2).setScale(3))
+    ), true)
+
+    // test BigDecimal with same value - same scale
+    testPredicate(new EqualTo(
+      Literal.of(BigDecimalJ.valueOf(2).setScale(3)),
+      Literal.of(BigDecimalJ.valueOf(2).setScale(3))
+    ), true)
   }
 
   test("null predicates") {
--- a/standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
+++ b/standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
@@ -119,8 +119,6 @@ class ExpressionSuite extends FunSuite {
       (Literal.of(1.0), Literal.of(2.0), Literal.of(1.0), Literal.ofNull(new DoubleType())),
       (Literal.of(1.toByte), Literal.of(2.toByte), Literal.of(1.toByte),
         Literal.ofNull(new ByteType())),
-      (Literal.of(new BigDecimalJ("123.45")), Literal.of(new BigDecimalJ("887.62")),
-        Literal.of(new BigDecimalJ("123.45")), Literal.ofNull(new DecimalType(5, 2))),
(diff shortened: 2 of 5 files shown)


standalone/src/main/java/io/delta/standalone/expressions/BinaryOperator.java
```diff
@@ -13,7 +13,7 @@ protected BinaryOperator(Expression left, Expression right, String symbol) {
         super(left, right);
         this.symbol = symbol;
 
-        if (!left.dataType().equals(right.dataType())) {
+        if (!left.dataType().equivalent(right.dataType())) {
```
allisonport-db, line 16: yep same thing for `In`

allisonport-db (commented):
Yes right approach, add tests please

tdas:
Please add a description explaining the need for this.

tdas:
Also, is this how Spark compares types?

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
```diff
@@ -185,6 +182,12 @@ class ExpressionSuite extends FunSuite {
         testPredicate(predicateCreator(Literal.of(small), Literal.of(small2)), smallSmall)
       }
     }
+
+    // TODO:
+    val decimalComparison = new EqualTo(
```
kination, line 187: @allisonport-db have some question.
For comparison, result for this will be false. How can I get output that `decimalComparison` is 'false'?

(are there something like 'isFalse', 'status' function inside this class?)

kination:
@tdas @allisonport-db thanks for review. Add some updates.

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
allisonport-db: You can use `.eval()` to evaluate an expression. Look at how we use `testPredicate` to test the other comparison expressions here: https://github.com/delta-io/connectors/blob/master/standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala#L111

kination: @allisonport-db thanks. Could you check the updates?

allisonport-db:
Looks good usage wise. Can you add tests for decimals of the same value with different scales? These should be equal.

Can you also make sure the tests you've added fail without the fix? I think one test is comparing decimals of the same scale I don't think this is testing anything new compared to our existing tests.

kination:
@allisonport-db I've updated few more test. Please check.

Also, could you give me some advice to fix on `Failing because of negative scalastyle result` issue in PR check?

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
```diff
@@ -19,15 +19,12 @@ package io.delta.standalone.internal
 import java.math.{BigDecimal => BigDecimalJ}
 import java.sql.{Date => DateJ, Timestamp => TimestampJ}
 import java.util.{Arrays => ArraysJ, Objects}
-
```
allisonport-db, line 22: without looking at the test logs, removing the blank lines separating import groups is definitely a scalastyle error

if you run `build/sbt standalone/scalastyle` and `build/sbt standalone/testScalastyle` you should be able to debug these errors yourself

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
```diff
@@ -185,6 +182,30 @@ class ExpressionSuite extends FunSuite {
         testPredicate(predicateCreator(Literal.of(small), Literal.of(small2)), smallSmall)
       }
     }
+
+    // test BigDecimal with diff value - same scale
```
allisonport-db, line 186: thanks for adding all the cases.

could you instead test this as part of the tests here https://github.com/delta-io/connectors/blob/master/standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala#L111 ? your change affects all binary operators and we can test all of them by adding it to this code

you should just be adding two lines:
```
// same scales (should already be there actually)
// different scales
```

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
```diff
@@ -235,6 +256,21 @@ class ExpressionSuite extends FunSuite {
       (0 to 10).map{Literal.of}.asJava), false)
     testPredicate( new In(Literal.of(10),
       (0 to 10).map{Literal.of}.asJava), true)
+
+    testPredicate(
```
allisonport-db, line 260: can you add a comment explaining why we are testing the big decimals here?

and can you cover the case with different scale? thanks

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
kination: @allisonport-db it is already included in `test("comparison predicates")`. Do you mean add in https://github.com/delta-io/connectors/blob/master/standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala#L113 ?

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
kination: @allisonport-db I've update this, but `testScalastyle` still shows the error. Could you give some advice about this?

allisonport-db: Have you run those commands? It should output what the scala style error is

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
allisonport-db: https://github.com/delta-io/connectors/blob/master/standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala#L113-L132 is testing all the comparison expressions we support. please test the decimals by adding them to that sequence of literals

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
kination: @allisonport-db currently `build/sbt standalone/scalastyle` works, but `build/sbt standalone/testScalastyle` fails ...

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
kination: @allisonport-db thanks. Updated 🙏

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
```diff
@@ -126,7 +129,13 @@ class ExpressionSuite extends FunSuite {
       (Literal.of("apples"), Literal.of("oranges"), Literal.of("apples"),
         Literal.ofNull(new StringType())),
       (Literal.of("apples".getBytes()), Literal.of("oranges".getBytes()),
-        Literal.of("apples".getBytes()), Literal.ofNull(new BinaryType()))
+        Literal.of("apples".getBytes()), Literal.ofNull(new BinaryType())),
+      // same scales (should already be there actually)
```
allisonport-db, line 133: you can remove the above ones (line 122)

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
```diff
@@ -126,7 +129,13 @@ class ExpressionSuite extends FunSuite {
...
+      (Literal.of(BigDecimalJ.valueOf(1).setScale(2)), Literal.of(BigDecimalJ.valueOf(3).setScale(2)),
+        Literal.of(BigDecimalJ.valueOf(1).setScale(2)), Literal.ofNull(new DecimalType(5, 2))),
+      // different scales
+      (Literal.of(BigDecimalJ.valueOf(1).setScale(1)), Literal.of(BigDecimalJ.valueOf(3).setScale(3)),
+        Literal.of(BigDecimalJ.valueOf(1).setScale(2)), Literal.ofNull(new DecimalType(5, 2))),
```
allisonport-db, line 138: this tailing comma is causing the scalastyle error

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
```diff
@@ -126,7 +129,13 @@ class ExpressionSuite extends FunSuite {
       (Literal.of("apples"), Literal.of("oranges"), Literal.of("apples"),
         Literal.ofNull(new StringType())),
       (Literal.of("apples".getBytes()), Literal.of("oranges".getBytes()),
-        Literal.of("apples".getBytes()), Literal.ofNull(new BinaryType()))
+        Literal.of("apples".getBytes()), Literal.ofNull(new BinaryType())),
+      // same scales (should already be there actually)
+      (Literal.of(BigDecimalJ.valueOf(1).setScale(2)), Literal.of(BigDecimalJ.valueOf(3).setScale(2)),
+        Literal.of(BigDecimalJ.valueOf(1).setScale(2)), Literal.ofNull(new DecimalType(5, 2))),
```
allisonport-db, line 135: I think precision of these will not be 5? the null value should also be of the same precision and scale here

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
```diff
@@ -266,11 +252,19 @@ class ExpressionSuite extends FunSuite {
           Literal.of(BigDecimalJ.valueOf(3).setScale(1)),
           Literal.of(BigDecimalJ.valueOf(4).setScale(1)),
           Literal.of(BigDecimalJ.valueOf(5).setScale(1)),
-        ).asJava
-      ),
-        true
-      )
+        ).asJava),true)
```
allisonport-db, line 255: this missing space will probably cause a scalastyle error as well

allisonport-db: same below

allisonport-db (commented):
almost done I think a few minor comments. please make sure the github tests are passing

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
kination: @allisonport-db `DecimalType` cannot be defined with no parameter. Will it be okay to do it as `new DecimalType(0, 0)`?

allisonport-db: it needs to be the same as the left values. I think it's (1, 2) but can you double check?

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
```diff
@@ -235,6 +239,30 @@ class ExpressionSuite extends FunSuite {
...
+    // BigDecimal value in list, with different scale
+    testPredicate(
+      new In(
+        Literal.of(BigDecimalJ.valueOf(2).setScale(1)),
+        List(
+          Literal.of(BigDecimalJ.valueOf(1).setScale(2)),
+          Literal.of(BigDecimalJ.valueOf(2).setScale(2)),
+          Literal.of(BigDecimalJ.valueOf(3).setScale(2)),
+          Literal.of(BigDecimalJ.valueOf(4).setScale(2)),
+          Literal.of(BigDecimalJ.valueOf(5).setScale(2)),
```
allisonport-db, line 264: the comma's at the ends of lists are causing the scalastyle errors. can you _please_ make sure scalastyle is passing when you make updates to the PR?

kination: @allisonport-db sorry for disturbing, it was my mistake. I've updated to pass scala-style. Could you check again?

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
allisonport-db: nit: remove the comment (should already be there)

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
```diff
@@ -129,7 +127,17 @@ class ExpressionSuite extends FunSuite {
...
+      // same scales (should already be there actually)
+      (Literal.of(BigDecimalJ.valueOf(1).setScale(2)),
+        Literal.of(BigDecimalJ.valueOf(3).setScale(2)),
+        Literal.of(BigDecimalJ.valueOf(1).setScale(2)),
+        Literal.ofNull(new DecimalType(1, 2))),
+      // different scales
+      (Literal.of(BigDecimalJ.valueOf(1).setScale(2)),
+        Literal.of(BigDecimalJ.valueOf(3).setScale(3)),
+        Literal.of(BigDecimalJ.valueOf(1).setScale(4)),
+        Literal.ofNull(new DecimalType(1, 2)))
```
allisonport-db, line 140: can you make the null value also a different scale/precision (it's the same as the first value here)?
otherwise the value is the same test as the above test

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
allisonport-db: Sorry I meant can you add a more specific comment such as...
```
// Here we test In specifically with the BigDecimal data type to make sure we cover the different cases with values and elements of varying precision and scales
```

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
kination: @allisonport-db I've thought it will be difference because input value has all same scale at the first one, and all different scale here. Anyway, make some changes here too.

kination:
@allisonport-db updated, but wonder why build has been failed...

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
```diff
@@ -129,7 +127,16 @@ class ExpressionSuite extends FunSuite {
       (Literal.of("apples"), Literal.of("oranges"), Literal.of("apples"),
         Literal.ofNull(new StringType())),
       (Literal.of("apples".getBytes()), Literal.of("oranges".getBytes()),
-        Literal.of("apples".getBytes()), Literal.ofNull(new BinaryType()))
+        Literal.of("apples".getBytes()), Literal.ofNull(new BinaryType())),
+      (Literal.of(BigDecimalJ.valueOf(1).setScale(2)),
```
allisonport-db, line 131: nit: add comment for `// same scales`

allisonport-db (approved):
LGTM

We can merge this after you add that one comment

standalone/src/test/scala/io/delta/standalone/internal/ExpressionSuite.scala
kination: @allisonport-db done 🙏

allisonport-db:
Can you sign off your latest commit as detailed at the bottom of this page? https://github.com/delta-io/delta/blob/master/CONTRIBUTING.md

kination:
@allisonport-db updated

allisonport-db:
Yes, I also added it to your PR description. Thanks, I'm merging this

PR #436, opened 2022-09-07
[424] Fix delta-flink connector classloader bug

### LINKED ISSUE #424: Flink Source Connector - problem with Delta's shared forkJoinPool when computing Snapshot files
For Flink jobs using Delta-Flink source connector it may happen that Flink will throw exception:

```
Caused by: java.lang.IllegalStateException: Trying to access closed classloader. Please check if you store classloaders directly or indirectly in static fields. If the stacktrace suggests that the leak occurs in a third party library and cannot be fixed immediately, you can disable this check with the configuration 'classloader.check-leaked-classloader'.
    at org.apache.flink.runtime.execution.librarycache.FlinkUserCodeClassLoaders$SafetyNetWrapperClassLoader.ensureInner(FlinkUserCodeClassLoaders.java:164)
    at org.apache.flink.runtime.execution.librarycache.FlinkUserCodeClassLoaders$SafetyNetWrapperClassLoader.getResources(FlinkUserCodeClassLoaders.java:188)
```

**Flink Exception description**
This is a Flink's check to detect if user code (Flink Job code) uses a class loader that was already used by some other job that was executed on this cluster. This is a Flink's safety check to detect case when for any reason not all classes from previous, finished job were unloaded from Flink memory. 

Such object reference leak between jobs is possible whenever we have a static reference to a class loader instance or whenever a shared, often static thread pool is used by Flink Job. In case of delta-standalone we have the latter.

The purpose of this check in Flink is to prevent OOM due to class reference leak.

**Reproduce**
1. comment out `configuration.set(CoreOptions.CHECK_LEAKED_CLASSLOADER, false);` from connectors/flink `DeltaTestUtils.java` `buildCluster` method
2. Run `DeltaSourceBoundedExecutionITCaseTest::shouldReadDeltaTableUsingUserSchema` test with substituting line 
`shouldReadDeltaTable(deltaSource, failoverType)` to `shouldReadDeltaTable(deltaSource, FailoverType.NONE)`

This will execute `shouldReadDeltaTableUsingUserSchema` test 3 times. First run will pass, where second will "stuck". Then you can observe Flink exception in logs.

**Location of the Problem**
The problem is located in `SnapshotImpl.scala` in `loadInMemory` method executed when computing Snapshot's `state`.
The `loadInMemory` uses Scalla's parallel collection -> `ParVector` which uses ForkJoinThreadPool that is shared for entire JVM. It may happen that Flink Job that uses Delta Source connector will compute Delta Snapshot files with thread that was already used by previous Job on the same cluster -> this will trigger exception from aboive.

**Possible Solution**
The solution would be to not use shared Fork Join Thread Pool but custom pool with scope bounded to Snapshot object instance.
In order to substitute thread pool for parallel collection instance you need to:
Define a new Pool:
```
import java.util.concurrent.Executors
import java.util.concurrent.ThreadFactory

  final private class ContextClassloaderFixedThreadFactory extends ThreadFactory {
    override def newThread(runnable: Runnable): Thread = {
      val thread = Executors.defaultThreadFactory.newThread(runnable)
      thread.setContextClassLoader(getClass.getClassLoader)
      thread
    }
  }

  import scala.concurrent._

  implicit val myEc = new ExecutionContext {

    val factory: ThreadFactory = new ThreadFactoryBuilder()
        .setDaemon(true)
        .setNameFormat("delta-worker-pool-%d")
        .setThreadFactory(new ContextClassloaderFixedThreadFactory())
        .build();

    val threadPool: ExecutorService = Executors.newFixedThreadPool(10, factory)

    def execute(runnable: Runnable): Unit = {
      threadPool.submit(runnable)
    }

    def reportFailure(t: Throwable): Unit = {}
  }
```

Use it in the parallel collection:
```
 val pv = new ParVector(paths.map(_.toString).sortWith(_ < _).toVector)
 pv.tasksupport = new ExecutionContextTaskSupport(myEc) <- substituting the default pool and execution context for the collection.
    pv.flatMap { path =>
      if (path.endsWith("json")) {
      ....
      }
     }
```

**Final thoughts**
Thing worth to mention is that described issue will occur only for Flink users. For other users, that are using "vanilla" delta-standalone the forkJoinThreadPool most likely will be just fine. Because of that maybe it will be good to defined an API for Snapshot object that will allow to inject custom thread pool. If no pool was injected that ForkJoinPool will be used.
With this change, we could use the new API in Flink connector and nothing would have change for other users.

scottsand-db:
Resolves #424

### The Problem
#424 describes an issue where our usage of the shared-per-JVM `ForkJoinPool.commonPool()` inside of `SnapshotImpl.loadInMemory` would cause the following error

```
Caused by: java.lang.IllegalStateException: Trying to access closed classloader. Please check if you store classloaders directly or indirectly in static fields. If the stacktrace suggests that the leak occurs in a third party library and cannot be fixed immediately, you can disable this check with the configuration 'classloader.check-leaked-classloader'.
    at org.apache.flink.runtime.execution.librarycache.FlinkUserCodeClassLoaders$SafetyNetWrapperClassLoader.ensureInner(FlinkUserCodeClassLoaders.java:164)
    at org.apache.flink.runtime.execution.librarycache.FlinkUserCodeClassLoaders$SafetyNetWrapperClassLoader.getResources(FlinkUserCodeClassLoaders.java:188)
```

### The Solution
This PR fixes the above error by no longer using the statically-shared common thread pool. Instead, we create a `ForkJoinPool` instance per each `Snapshot::loadInMemory` invocation. This overhead is negligible.

### How was this change tested?
```
build/sbt 'flink/testOnly *DeltaSourceBoundedExecutionITCaseTest'
```

We also explicitly set the flink cluster option `CHECK_LEAKED_CLASSLOADER` to true, so that any class leaks will fail the tests.

### DIFF
--- a/flink/src/test/java/io/delta/flink/source/DeltaSourceBoundedExecutionITCaseTest.java
+++ b/flink/src/test/java/io/delta/flink/source/DeltaSourceBoundedExecutionITCaseTest.java
@@ -200,7 +200,10 @@ public class DeltaSourceBoundedExecutionITCaseTest extends DeltaSourceITBase {
         "2022-06-15 13:24:33.613",
         "2022-06-15 13:25:33.632",
         "2022-06-15 13:26:33.633",
-        "2022-06-15 13:27:33.634"
+
+        // Local filesystem will truncate the logFile last modified timestamps to the nearest
+        // second. So, for example, "2022-06-15 13:27:33.001" would be after last commit.
+        "2022-06-15 13:27:33.000"
     };
 
     /**
--- a/flink/src/test/java/io/delta/flink/utils/DeltaTestUtils.java
+++ b/flink/src/test/java/io/delta/flink/utils/DeltaTestUtils.java
@@ -149,7 +149,9 @@ public class DeltaTestUtils {
 
     public static MiniClusterWithClientResource buildCluster(int parallelismLevel) {
         Configuration configuration = new Configuration();
-        configuration.set(CoreOptions.CHECK_LEAKED_CLASSLOADER, false);
+
+        // By default, let's check for leaked classes in tests.
+        configuration.set(CoreOptions.CHECK_LEAKED_CLASSLOADER, true);
 
         return new MiniClusterWithClientResource(
             new MiniClusterResourceConfiguration.Builder()
--- a/standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
@@ -63,7 +63,7 @@ private[internal] class SnapshotImpl(
   //
   // This default-constructor instance will use a thread pool of size equal to the number of
   // processors available to the JVM.
-  private val forkJoinPool = new java.util.concurrent.ForkJoinPool()
+  private val forkJoinPool = new scala.concurrent.forkjoin.ForkJoinPool()
 
   private val memoryOptimizedLogReplay =
     new MemoryOptimizedLogReplay(files, deltaLog.store, hadoopConf, deltaLog.timezone)
--- a/standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
@@ -177,29 +177,19 @@ private[internal] class SnapshotImpl(
     // instance shared by the entire JVM. This can cause issues for downstream connectors (e.g.
     // the flink-delta connector) that require no object reference leaks between jobs. See #424 for
     // more details. To solve this, we create and use our own ForkJoinPool instance per each method
-    // invocation. It is of type `ExecutorService` since we cannot directly use ForkJoinPool API as
-    // Scala 2.11 uses scala.collection.forkjoin.ForkJoinPool but
-    // Scala 2.12/2.13 uses java.util.concurrent.ForkJoinPool.
+    // invocation.
+    //
+    // Note that we cannot create a ForkJoinPool directly as Scala 2.11 uses
+    // scala.collection.forkjoin.ForkJoinPool but Scala 2.12/2.13 uses
+    // java.util.concurrent.ForkJoinPool.
 
-    // Creates a new ForkJoinPool instance. This instance will use a thread pool of size equal to
-    // the number of processors available to the JVM.
-    val threadPool: ExecutorService = Executors.newWorkStealingPool()
-
-    val executionContext: ExecutionContext = new ExecutionContext {
-      override def execute(runnable: Runnable): Unit = {
-        threadPool.submit(runnable)
-      }
-
-      override def reportFailure(cause: Throwable): Unit = {
-        cause.printStackTrace()
-      }
-    }
-
-    val pv = new ParVector(paths.map(_.toString).sortWith(_ < _).toVector)
-
-    pv.tasksupport = new ExecutionContextTaskSupport(executionContext)
+    // Under the hood, creates a new ForkJoinPool instance. This instance will use a thread pool of
+    // size equal to the number of processors available to the JVM.
+    val execContextService = ExecutionContext.fromExecutorService(null)
 
     try {
+      val pv = new ParVector(paths.map(_.toString).sortWith(_ < _).toVector)
+      pv.tasksupport = new ExecutionContextTaskSupport(execContextService)
       pv.flatMap { path =>
         if (path.endsWith("json")) {
           import io.delta.standalone.internal.util.Implicits._
@@ -222,7 +212,7 @@ private[internal] class SnapshotImpl(
         } else Seq.empty[SingleAction]
       }.toList
     } finally {
-      threadPool.shutdown()
+      execContextService.shutdown()
     }
   }
 
--- a/standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
@@ -55,6 +55,16 @@ private[internal] class SnapshotImpl(
 
   import SnapshotImpl._
 
+  // `loadInMemory` uses a parallel collection which, by default, uses ForkJoinPool.commonPool().
+  // This is a static ForkJoinPool instance shared by the entire JVM. This can cause issues for
+  // downstream connectors (e.g. the flink-delta connector) that require no object reference leaks
+  // between jobs. See #424 for more details. To solve this, we create and use our own ForkJoinPool
+  // instance per each Snapshot.
+  //
+  // This default-constructor instance will use a thread pool of size equal to the number of
+  // processors available to the JVM.
+  private val forkJoinPool = new java.util.concurrent.ForkJoinPool()
+
   private val memoryOptimizedLogReplay =
     new MemoryOptimizedLogReplay(files, deltaLog.store, hadoopConf, deltaLog.timezone)
 
@@ -170,7 +180,9 @@ private[internal] class SnapshotImpl(
   }
 
   private def loadInMemory(paths: Seq[Path]): Seq[SingleAction] = {
-    new ParVector(paths.map(_.toString).sortWith(_ < _).toVector).par.flatMap { path =>
+    val pv = new ParVector(paths.map(_.toString).sortWith(_ < _).toVector)
+    pv.tasksupport = new scala.collection.parallel.ForkJoinTaskSupport(forkJoinPool)
+    pv.flatMap { path =>
       if (path.endsWith("json")) {
         import io.delta.standalone.internal.util.Implicits._
         deltaLog.store
--- a/standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
@@ -17,9 +17,12 @@
 package io.delta.standalone.internal
 
 import java.net.URI
+import java.util.concurrent.{Executors, ExecutorService}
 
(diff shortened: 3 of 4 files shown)


scottsand-db:
Problem:
- scala 2.11 uses a `scala.concurrent.forkjoin.ForkJoinPool` inside of its `ForkJoinTaskSupport` constructor.
- scala 2.12 and 2.13 uses a `java.util.concurrent.ForkJoinPool` inside of its `ForkJoinTaskSupport` constructor.

https://javadoc.io/doc/org.scala-lang/scala-library/2.11.0-M3/scala/collection/parallel/ForkJoinTaskSupport.html
https://javadoc.io/doc/org.scala-lang/scala-library/2.12.0-M3/scala/collection/parallel/ForkJoinTaskSupport.html

standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
```diff
@@ -55,6 +55,16 @@ private[internal] class SnapshotImpl(
...
 
+  // `loadInMemory` uses a parallel collection which, by default, uses ForkJoinPool.commonPool().
+  // This is a static ForkJoinPool instance shared by the entire JVM. This can cause issues for
+  // downstream connectors (e.g. the flink-delta connector) that require no object reference leaks
+  // between jobs. See #424 for more details. To solve this, we create and use our own ForkJoinPool
+  // instance per each Snapshot.
+  //
+  // This default-constructor instance will use a thread pool of size equal to the number of
+  // processors available to the JVM.
+  private val forkJoinPool = new scala.concurrent.forkjoin.ForkJoinPool()
```
zsxwing, line 66: We need to close this otherwise we will leak threads. Could you move this into `loadInMemory` and shut down it before returning.

tdas: Why create new forkjoinpool for every snapshot? why not create a pool for a DeltaLog? Or even have a static forkjoin pool in DeltaLog object for all snapshots every where?

flink/src/test/java/io/delta/flink/source/DeltaSourceBoundedExecutionITCaseTest.java
```diff
@@ -200,7 +200,10 @@ public void shouldReadVersionAsOf(
         "2022-06-15 13:24:33.613",
         "2022-06-15 13:25:33.632",
         "2022-06-15 13:26:33.633",
-        "2022-06-15 13:27:33.634"
+
+        // Local filesystem will truncate the logFile last modified timestamps to the nearest
+        // second. So, for example, "2022-06-15 13:27:33.001" would be after last commit.
+        "2022-06-15 13:27:33.000"
```
tdas, line 206: why is this change needed for this PR?

standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
scottsand-db: @tdas https://docs.scala-lang.org/overviews/parallel-collections/configuration.html
```
Note that if you are making your own ForkJoinPool instance you should call ForkJoinPool.shutdown() when you no longer need the thread pool. If you do not call ForkJoinPool.shutdown() and continue to create new instances of ForkJoinPool the JVM may eventually run out of available threads and throw a java.lang.OutOfMemoryError.
```

scottsand-db: And we cannot have a static reference ... because that was the issue in the first place. Here, in the current code, I am creating my own ForkJoinPool instance **instead** of using `ForkJoinPool.commonPool()` which returns `static final ForkJoinPool common;`

flink/src/test/java/io/delta/flink/source/DeltaSourceBoundedExecutionITCaseTest.java
scottsand-db: good question. my local tests were failing. I decided to fix it now, since it is only a few lines.

I'm happy to put this into a different PR.

scottsand-db:
@kristoffSC take a look?

standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
```diff
@@ -170,27 +173,57 @@ private[internal] class SnapshotImpl(
...
+      }
+
+      override def reportFailure(cause: Throwable): Unit = {
+        cause.printStackTrace()
+      }
+    }
+
+    val pv = new ParVector(paths.map(_.toString).sortWith(_ < _).toVector)
+
+    pv.tasksupport = new ExecutionContextTaskSupport(executionContext)
```
zsxwing, line 200: we can just use `new ExecutionContextTaskSupport(ExecutionContext.fromExecutor(null))`

zsxwing: By the way, it's better to move `val pv = new ParVector(paths.map(_.toString).sortWith(_ < _).toVector)` inside `try` as it could fail.

tdas: we will be creating a new threadpool every time? and when will this threadpool be closed?

tdas:
@scottsand-db i thought static references within classes loaded by the classloader is fine, but I maybe wrong about this.

standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
zsxwing: > we will be creating a new threadpool every time? and when will this threadpool be closed?

Just before `loadInMemory` returns

zsxwing:
> i thought static references within classes loaded by the classloader is fine, but I maybe wrong about this.

IIRC, Flink creates a classloader to run Delta Standalone code. Static references will keep the classloader alive and cannot be GCed.

standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
scottsand-db: @zsxwing - I don't see any docs or code that say the `ForkJoinPool` created by the `ExecutionContext.fromExecutor(null)` call (which then calls `ExecutionContextImpl::createDefaultExecutorService`) is `shutdown`. There's also no way to access the `ForkJoinPool` inside the `ExecutionContextTaskSupport` to shut it down ourselves.

Isn't this vulnerable to thread leaks?

kristoffSC:
@scottsand-db 
Could you mind to include in this PR removing `configuration.set(CoreOptions.CHECK_LEAKED_CLASSLOADER, false);` from connectors/flink `DeltaTestUtils::buildCluster` method.

After this PR we should not need it and this would be an actual proof, that fix is working for flink connector.

standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
zsxwing: Hm, you are right. How about this
```
    val ec = ExecutionContext.fromExecutorService(null)
    try {
      new ExecutionContextTaskSupport(ec)
    } finally {
      ec.shutdownNow()
    }
```

standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
```diff
@@ -170,27 +173,47 @@ private[internal] class SnapshotImpl(
...
-            hadoopConf = hadoopConf)
-        )
-        try {
-          parquetIterable.toArray.map(_.unwrap)
-        } finally {
-          parquetIterable.close()
-        }
-      } else Seq.empty[SingleAction]
-    }.toList
+    // `ParVector` by default, uses ForkJoinPool.commonPool(). This is a static ForkJoinPool
```
scottsand-db, line 176: TODO: explain why this cannot be done per-Snapshot-instance, nor statically

flink/src/test/java/io/delta/flink/utils/DeltaTestUtils.java
```diff
@@ -149,7 +149,9 @@ public static void restartTaskManager(Runnable afterFailAction, MiniCluster mini
 
     public static MiniClusterWithClientResource buildCluster(int parallelismLevel) {
         Configuration configuration = new Configuration();
-        configuration.set(CoreOptions.CHECK_LEAKED_CLASSLOADER, false);
+
+        // By default, let's check for leaked classes in tests.
+        configuration.set(CoreOptions.CHECK_LEAKED_CLASSLOADER, true);
```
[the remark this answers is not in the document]
scottsand-db: I like stating it explicitly.

scottsand-db:
2.11 flink tests passed; 2.12 flink tests failed.
https://github.com/delta-io/connectors/runs/8257560080?check_suite_focus=true
```
[error] Failed: Total 359, Failed 1, Errors 0, Passed 358
...
error] Test io.delta.flink.sink.DeltaSinkStreamingExecutionITCase.testFileSink failed: org.opentest4j.AssertionFailedError: expected: <40000> but was: <32000>, took 5.965s
[error]     at io.delta.flink.sink.DeltaSinkStreamingExecutionITCase.runDeltaSinkTest(DeltaSinkStreamingExecutionITCase.java:236)
[error]     at io.delta.flink.sink.DeltaSinkStreamingExecutionITCase.testFileSink(DeltaSinkStreamingExecutionITCase.java:137)
```

zsxwing (approved):
LGTM

SteNicholas (approved):
LGTM.

kristoffSC:
LGTM :)

kristoffSC (approved):

PR #442, opened 2022-09-12
use complete directory path after filesystem to list files Signed-off…

CabbaggeCollector:
{0, RelativePosition.FromEnd}) is empty when pasting the URL with a '/' in the end (e.g. https://test.dfs.core.windows.net/test/foo/bar.delta/) or else 'bar.delta'. But don't we need the whole path here, like 'foo/bar.delta', including 'foo'?

How is the "directoryCheck" supposed to work/ what should it do? Can't we use directoryCheck directly as directory since we check for "" and return ""? Else Text.AfterDelimiter(Folder, "/", 3) gives the whole directory path which is fine I guess?

Best regards
-by: Jonas vom Stein <jonasvomstein97@gmail.com>

### DIFF
--- a/powerbi/CustomConnector/DeltaLake/DeltaLake.pq
+++ b/powerbi/CustomConnector/DeltaLake/DeltaLake.pq
@@ -109,8 +109,7 @@ DeltaLakeContentBlob = (Folder as text) as table =>
 DeltaLakeContentADLSGen2 = (Folder as text) as table =>
     let
         filesystem = Text.BeforeDelimiter(Folder, "/", 3),
-        directoryCheck = Text.AfterDelimiter(Folder, "/", 3),
-        directory = if directoryCheck = "" then "" else Text.AfterDelimiter(Folder, "/", {0, RelativePosition.FromEnd}),
+        directory = Text.AfterDelimiter(Folder, "/", 3),
         url = filesystem & "?recursive=true&resource=filesystem&directory=" & directory,
         headers = if Extension.CurrentCredential()[AuthenticationKind] = "Implicit" then [] else [Headers=SignRequest(url,directory)],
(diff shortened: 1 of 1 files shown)


scottsand-db:
@gbrueckl can you take a look?

gbrueckl:
Just had a quick look but the custom connector was written by @pichlerpa and not me


Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

CabbaggeCollector:
Maybe it helps: For BLOBs this is already done like in this PR, if you use the connector in Power BI with a blob URL it will list the delta lakes in all subfolders:
https://github.com/delta-io/connectors/blob/2dc8c561998dbb53c1c73fad3fac7f8ce129435e/powerbi/CustomConnector/DeltaLake/DeltaLake.pq#L69
(line 68-69)
For DFS the code goes always through the root folder when the URL ends with '/'. Maybe the DFS code was tested with delta lakes directly in the root folder?
Like https://test.dfs.core.windows.net/test/bar.delta

pichlerpa:
Hi, I will have a look. It could be connected to changes I made to see the folder hierarchy nested in the "Get data" experience  which I finally reverted due to some difficulties a faced. Patrick

pichlerpa:
@CabbaggeCollector with which Power BI Desktop version have you faced this issue?

pichlerpa:
@CabbaggeCollector thanks, I just tested your code, seems fine to me. Can you also please try to update your Power BI Desktop to the latest version and check whether the connector still works? I had problems to reproduce your problem because the connector doesn't seem to work with newer versions, I just wanted to be sure that's not a local problem.

vkorukanti:

PR #451, opened 2022-10-20
Internal API for higher protocol tables + CHECK constraints APIs

allisonport-db:
## Overview

Adds internal APIs for testing purposes to allow creating tables of higher-protocol versions, and allow reading/writing to them. All public APIs should be unaffected and will still only support Protocol(1, 2).

Also adds support for CHECK constraints.

### Reading/writing to higher-protocol tables
We introduce the separate concepts of standalone's supported protocol version vs. the connector's supported protocol version. For testing, we have flag `ignoreConnectorProtocolChecks` which ignores the second check and only checks for standalone's protocol compatibility. This can only be set to true from `DeltaLogImpl.forTable`.

### Creating higher-protocol tables
We also introduce `log.startTransactionWithInitialProtocolVersion` which enables creating a new table with a higher protocol version than (1,2). This is a private API that will only be used for testing. Once we introduce public APIs this operation can be updated to be:
```
txn.upgradeProtocolVersion(1, 3)
```
#### Reasoning

Other possibilities to allow this include:
1. Adding a private API to `OptimisticTransactionImpl` instead, but this would require any tests using it to cast `txn.asInstanceOf[OptimisticTransactionImpl]`. We also would still only want to support version 0 so this would not be the same API that we plan to later make public.
2. Commit the Protocol file action directly. Since we plan to block this later on this makes the later migration more painful.

## CHECK constraints

We introduce a java class `Constraint` that will represent check constraints and column invariants. APIs for retrieving constraints and adding/removing check constraints are added to `Metadata`. We also add a test suite `DeltaConstraintsSuite`

### DIFF
--- a/standalone/src/main/java/io/delta/standalone/Constraint.java
+++ b/standalone/src/main/java/io/delta/standalone/Constraint.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright (2020-present) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.delta.standalone;
+
+import java.util.Objects;
+
+import javax.annotation.Nonnull;
+
+/**
+ * Represents a constraint defined on a Delta table which writers must verify before writing.
+ * Constraints can come in one of two ways:
+ * - A CHECK constraint which is stored in {@code metadata.configuration}
+ * - A column invariant which is stored in a column's metadata
+ */
+// TODO: if we ever plan to support not-null constraints as a java class should we name this
+//  "CheckConstraint" even though we will also be presenting invariants in this form?
+public final class Constraint {
+
+    public static final String CHECK_CONSTRAINT_KEY_PREFIX = "delta.constraints.";
+    @Nonnull private final String name;
+    @Nonnull private final String expression;
+
+    public Constraint(
+            @Nonnull String name,
+            @Nonnull String expression) {
+        this.name = name;
+        this.expression = expression;
+    }
+
+    /**
+     * @return the name of this constraint
+     */
+    @Nonnull
+    public String getName() {
+        return name;
+    }
+
+    /**
+     * @return the expression to enforce of this constraint as a SQL string
+     */
+    @Nonnull
+    public String getExpression() {
+        return expression;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        Constraint constraint = (Constraint) o;
+        return Objects.equals(name, constraint.name) &&
+                Objects.equals(expression, constraint.expression);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(name, expression);
+    }
+}
--- a/standalone/src/main/java/io/delta/standalone/Constraint.java
+++ b/standalone/src/main/java/io/delta/standalone/Constraint.java
@@ -20,14 +20,17 @@ import java.util.Objects;
 
 import javax.annotation.Nonnull;
 
+import io.delta.standalone.actions.Metadata;
+import io.delta.standalone.types.StructField;
+
 // TODO: if we ever plan to support not-null constraints as a java class should we name this
 //  "CheckConstraint" even though we will also be presenting invariants in this form?
 
 /**
  * Represents a constraint defined on a Delta table which writers must verify before writing.
  * Constraints can come in one of two ways:
- * - A CHECK constraint which is stored in {@code metadata.configuration}
- * - A column invariant which is stored in a column's metadata
+ * - A CHECK constraint which is stored in {@link Metadata#getConfiguration()}
+ * - A column invariant which is stored in {@link StructField#getMetadata()}
  */
 public final class Constraint {
 
--- a/standalone/src/main/java/io/delta/standalone/Constraint.java
+++ b/standalone/src/main/java/io/delta/standalone/Constraint.java
@@ -20,14 +20,15 @@ import java.util.Objects;
 
 import javax.annotation.Nonnull;
 
+// TODO: if we ever plan to support not-null constraints as a java class should we name this
+//  "CheckConstraint" even though we will also be presenting invariants in this form?
+
 /**
  * Represents a constraint defined on a Delta table which writers must verify before writing.
  * Constraints can come in one of two ways:
  * - A CHECK constraint which is stored in {@code metadata.configuration}
  * - A column invariant which is stored in a column's metadata
  */
-// TODO: if we ever plan to support not-null constraints as a java class should we name this
-//  "CheckConstraint" even though we will also be presenting invariants in this form?
 public final class Constraint {
 
     public static final String CHECK_CONSTRAINT_KEY_PREFIX = "delta.constraints.";
--- a/standalone/src/main/java/io/delta/standalone/actions/Metadata.java
+++ b/standalone/src/main/java/io/delta/standalone/actions/Metadata.java
@@ -157,7 +157,8 @@ public final class Metadata implements Action {
     }
 
     /**
-     * Add the specified check constraint to the {@code configuration}
+     * Returns a new {@link Metadata} instance with the same properties as this instance but with
+     * the specified check constraint added to the {@code configuration}
      * @param name  the name of the check constraint
      * @param expression  the condition to enforce as a SQL string
      * @return a new {@link Metadata} instance with the same properties as this instance but with
@@ -175,7 +176,8 @@ public final class Metadata implements Action {
     }
 
     /**
-     * Removes the specified check constraint from the {@code configuration}
+     * Returns a new {@link Metadata} instance with the same properties as this instance but with
+     * the specified check constraint removed from the {@code configuration}
      * @param name  the name of the check constraint to remove
      * @return a new {@link Metadata} instance with the same properties as this instance but with
      *         the specified check constraint removed
--- a/standalone/src/main/java/io/delta/standalone/actions/Metadata.java
+++ b/standalone/src/main/java/io/delta/standalone/actions/Metadata.java
@@ -17,15 +17,21 @@
 package io.delta.standalone.actions;
 
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
+import io.delta.standalone.Constraint;
 import io.delta.standalone.types.StructType;
+
+import io.delta.standalone.internal.exception.DeltaErrors;
 
 /**
  * Updates the metadata of the table. The first version of a table must contain
@@ -131,6 +137,58 @@ public final class Metadata implements Action {
     @Nullable
     public StructType getSchema() {
         return schema;
+    }
+
+    /**
+     * @return the CHECK constraints and column invariants defined in this metadata
+     */
+    public List<Constraint> getConstraints() {
+        // todo: get column invariants
+
+        // get check constraints
+        List<Constraint> checkConstraints = configuration.entrySet().stream()
+                .filter(entry -> entry.getKey().startsWith(Constraint.CHECK_CONSTRAINT_KEY_PREFIX))
+                .map(entry -> new Constraint(
+                                entry.getKey().replace(Constraint.CHECK_CONSTRAINT_KEY_PREFIX, ""),
+                                entry.getValue()))
+                .collect(Collectors.toList());
+
+        return checkConstraints;
+    }
+
+    /**
+     * Add the specified check constraint to the {@code configuration}
+     * @param name  the name of the check constraint
+     * @param expression  the condition to enforce as a SQL string
+     * @return a new {@link Metadata} instance with the same properties as this instance but with
+     *         the added check constraint
+     * @throws IllegalArgumentException if a constraint with name already exists
+     */
+    public Metadata addCheckConstraint(String name, String expression) throws Throwable {
+        String fullKey = Constraint.CHECK_CONSTRAINT_KEY_PREFIX + name.toLowerCase(Locale.ROOT);
+        if (configuration.containsKey(fullKey)) {
+            throw DeltaErrors.checkConstraintAlreadyExists(name, configuration.get(fullKey));
+        }
+        Map<String, String> newConfiguration = new HashMap(configuration);
+        newConfiguration.put(fullKey, expression);
+        return copyBuilder().configuration(newConfiguration).build();
+    }
+
+    /**
+     * Removes the specified check constraint from the {@code configuration}
+     * @param name  the name of the check constraint to remove
+     * @return a new {@link Metadata} instance with the same properties as this instance but with
+     *         the specified check constraint removed
+     * @throws IllegalArgumentException if a constraint with name does not exist
+     */
+    public Metadata removeCheckConstraint(String name) throws Throwable {
+        String fullKey = Constraint.CHECK_CONSTRAINT_KEY_PREFIX + name.toLowerCase(Locale.ROOT);
+        if (!configuration.containsKey(fullKey)) {
+            throw DeltaErrors.checkConstraintDoesNotExist(name);
+        }
+        Map<String, String> newConfiguration = new HashMap(configuration);
+        newConfiguration.remove(fullKey);
+        return copyBuilder().configuration(newConfiguration).build();
     }
 
     @Override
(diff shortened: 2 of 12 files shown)


allisonport-db:
Still a few todo's in this code I'd like to resolve before merging (as well as some which are for later). But figured could start reviewing without those resolved

standalone/src/main/scala/io/delta/standalone/internal/exception/DeltaErrors.scala
```diff
@@ -31,6 +31,8 @@ import io.delta.standalone.internal.util.JsonUtils
 /** A holder object for Delta errors. */
 private[internal] object DeltaErrors {
 
+  // TODO: improve this error message to accommodate for standalone vs. connector protocol
+  //  incompatibilities
```
allisonport-db, line 35: I will do this before merge

standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionSuite.scala
```diff
@@ -316,4 +316,10 @@ class OptimisticTransactionSuite extends OptimisticTransactionSuiteBase {
       assert(committedAddFile.getPath === "file:/absolute/path/to/file/test.parquet")
     }
   }
+
+  ///////////////////////////////////////////////////////////////////////////
+  // prepareCommit() protocol checks
+  ///////////////////////////////////////////////////////////////////////////
+
+  // TODO: test appendOnly protocol check
```
allisonport-db, line 324: I will also do this before merge

scottsand-db:
> We introduce the separate concepts of standalone's supported protocol version vs. the connector's supported protocol version. 

The connector's supported protocol version must be a subset of standalone's supported protocol version, right?

standalone/src/main/java/io/delta/standalone/Constraint.java
```diff
@@ -0,0 +1,74 @@
...
+import javax.annotation.Nonnull;
+
+// TODO: if we ever plan to support not-null constraints as a java class should we name this
+//  "CheckConstraint" even though we will also be presenting invariants in this form?
+
+/**
+ * Represents a constraint defined on a Delta table which writers must verify before writing.
+ * Constraints can come in one of two ways:
+ * - A CHECK constraint which is stored in {@code metadata.configuration}
+ * - A column invariant which is stored in a column's metadata
```
scottsand-db, line 30: add the code/class/reference for a "column's metadata"

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -133,6 +139,58 @@ public StructType getSchema() {
         return schema;
     }
 
+    /**
+     * @return the CHECK constraints and column invariants defined in this metadata
+     */
+    public List<Constraint> getConstraints() {
+        // todo: get column invariants
```
scottsand-db, line 146: would be great to keep track of all of these TODOs in the project plan / github issue tracking this project's progress

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -133,6 +139,58 @@ public StructType getSchema() {
...
+     * @param name  the name of the check constraint to remove
+     * @return a new {@link Metadata} instance with the same properties as this instance but with
+     *         the specified check constraint removed
+     * @throws IllegalArgumentException if a constraint with name does not exist
+     */
+    public Metadata removeCheckConstraint(String name) throws Throwable {
+        String fullKey = Constraint.CHECK_CONSTRAINT_KEY_PREFIX + name.toLowerCase(Locale.ROOT);
+        if (!configuration.containsKey(fullKey)) {
+            throw DeltaErrors.checkConstraintDoesNotExist(name);
+        }
```
scottsand-db, line 188: I don't think we need to throw an error here. All the user wants is to ensure that the constraint no longer exists. If it was never there to begin with, that should be fine.

Happy to be convinced otherwise.

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -133,6 +139,58 @@ public StructType getSchema() {
...
+     * @param name  the name of the check constraint
+     * @param expression  the condition to enforce as a SQL string
+     * @return a new {@link Metadata} instance with the same properties as this instance but with
+     *         the added check constraint
+     * @throws IllegalArgumentException if a constraint with name already exists
+     */
+    public Metadata addCheckConstraint(String name, String expression) throws Throwable {
+        String fullKey = Constraint.CHECK_CONSTRAINT_KEY_PREFIX + name.toLowerCase(Locale.ROOT);
+        if (configuration.containsKey(fullKey)) {
+            throw DeltaErrors.checkConstraintAlreadyExists(name, configuration.get(fullKey));
```
scottsand-db, line 170: does delta-on-spark throw an error?

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -133,6 +139,58 @@ public StructType getSchema() {
...
+                .map(entry -> new Constraint(
+                                entry.getKey().replace(Constraint.CHECK_CONSTRAINT_KEY_PREFIX, ""),
+                                entry.getValue()))
+                .collect(Collectors.toList());
+
+        return checkConstraints;
+    }
+
+    /**
+     * Add the specified check constraint to the {@code configuration}
```
scottsand-db, line 160: IMO the `@return` statement seems more accurate+clear than this does.

There's can often be duplication between the method doc and the @return statement. I think it's fine to duplicate it verbatim, as that will make the javadoc most clear.

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
```diff
@@ -92,6 +93,17 @@ private[internal] class DeltaLogImpl private(
...
+  // TODO: eventually these (or some equivalent representation, possibly feature sets) will be set
+  //  via public APIs using DeltaLog.forTable.
+  //  For now we use protocol version numbers to keep the same only supported protocol version at
+  //  Protocol(1, 2).
+
+  /** Max reader protocol version the connector engine can read */
+  private lazy val connectorReaderVersion = 1
+
+  /** Max writer protocol version the connector engine can write to */
+  private lazy val connectorWriterVersion = 2
```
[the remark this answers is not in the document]
scottsand-db: you can put this comment in the code then

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
```diff
@@ -210,19 +242,35 @@ private[internal] class DeltaLogImpl private(
...
    */
   def assertProtocolRead(protocol: Protocol): Unit = {
+    // Check Delta Standalone's compatibility
     if (protocol != null && Action.readerVersion < protocol.minReaderVersion) {
       throw new DeltaErrors.InvalidProtocolVersionException(Action.protocolVersion, protocol)
     }
+    // Check the connector's compatibility
+    if (!ignoreConnectorProtocolChecks) {
+      if (protocol != null && connectorReaderVersion < protocol.minReaderVersion) {
+        throw new DeltaErrors.InvalidProtocolVersionException(
```
scottsand-db, line 252: we should throw a different error. this error references `Delta Standalone`

standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
```diff
@@ -79,6 +82,36 @@ private[internal] object Protocol {
     assert(!metadata.configuration.contains(MIN_WRITER_VERSION_PROP), s"Should not have the " +
       s"protocol version ($MIN_WRITER_VERSION_PROP) as part of table properties")
   }
+
+  /** Check that the protocol is compatible with any features enabled in the table metadata */
+  def checkMetadataProtocolCompatibility(metadata: Metadata, protocol: Protocol): Unit = {
```
scottsand-db, line 87: `checkMetadataProtocolFeatureCompatibility` ?

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
scottsand-db: nvm saw your comment below. can you include a `TODO` here too (about using a different error msg)? So Venki/the next reviewer doesn't repeat what I did

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
```diff
@@ -210,19 +242,35 @@ private[internal] class DeltaLogImpl private(
...
    */
   def assertProtocolWrite(protocol: Protocol): Unit = {
+    // Check Delta Standalone's compatibility
     if (protocol != null && Action.writerVersion < protocol.minWriterVersion) {
       throw new DeltaErrors.InvalidProtocolVersionException(Action.protocolVersion, protocol)
     }
+    // Check the connector's compatibility
+    if (!ignoreConnectorProtocolChecks) {
+      if (protocol != null && connectorWriterVersion < protocol.minWriterVersion) {
+        throw new DeltaErrors.InvalidProtocolVersionException(
```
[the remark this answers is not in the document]
scottsand-db: same as above. whoops. my bad. saw your explanation below.

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
```diff
@@ -210,19 +242,35 @@ private[internal] class DeltaLogImpl private(
    * allowed to read the table that is using the given `protocol`.
    */
   def assertProtocolRead(protocol: Protocol): Unit = {
+    // Check Delta Standalone's compatibility
     if (protocol != null && Action.readerVersion < protocol.minReaderVersion) {
       throw new DeltaErrors.InvalidProtocolVersionException(Action.protocolVersion, protocol)
     }
+    // Check the connector's compatibility
+    if (!ignoreConnectorProtocolChecks) {
```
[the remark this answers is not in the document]
scottsand-db: i'm also not sure if having this flag be part of the deltaLog api is the best route / most scalable. will all our test-only flags be part of the delta log API?

I'm sure there's some scala/sbt way to check for a test flag instead without changing apis.

scottsand-db (changes requested):

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
```diff
@@ -183,6 +196,25 @@ private[internal] class DeltaLogImpl private(
   // Internal Methods
   ///////////////////////////////////////////////////////////////////////////
 
+  // This is a temporary private API to create tables with a higher protocol version to allow
+  // implementing and testing higher-protocol features before we add the public APIs to upgrade
+  // table protocol versions.
+  private[standalone] def startTransactionWithInitialProtocolVersion(
```
scottsand-db, line 202: can you update your PR description to say that `startTransactionWithInitialProtocolVersion` is just for testing?

scottsand-db:
@allisonport-db btw - didn't we decide this was going to target a feature branch?

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
allisonport-db: I just followed what we do in sql when doing with a constraint that DNE
```
alter table table_name drop constraint constraint_name
```

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -133,6 +139,58 @@ public StructType getSchema() {
...
+     * @param name  the name of the check constraint
+     * @param expression  the condition to enforce as a SQL string
+     * @return a new {@link Metadata} instance with the same properties as this instance but with
+     *         the added check constraint
+     * @throws IllegalArgumentException if a constraint with name already exists
+     */
+    public Metadata addCheckConstraint(String name, String expression) throws Throwable {
+        String fullKey = Constraint.CHECK_CONSTRAINT_KEY_PREFIX + name.toLowerCase(Locale.ROOT);
```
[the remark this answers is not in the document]
allisonport-db: This is a good point. On second thought I think we should perform some minor format verification that aligns with what we allow in spark (and I think we would block a name that already has the prefix)

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
allisonport-db: yes

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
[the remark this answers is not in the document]
allisonport-db: So currently this check is what maintains our current behavior when using the public APIs (i.e. a table with Protocol(1, 3) will pass the first check (Standalone's) but will fail this one still) while still being able to upgrade "Delta Standalone's" supported version.

My thinking was that in some tests, we want to only test Standalone's APIs directly without the concept of a "connector". This flag essentially allows us to do that. Likewise, for some tests we do want to test the behavior as a connector so we don't want `ignoreConnectorProtocolChecks=true` for all tests (for example, we have some current tests that make sure we cannot read/write to higher protocol tables through the public APIs.) Also eventually we will want to test these checks once we expand the public APIs.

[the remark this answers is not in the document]
allisonport-db: Depending on how we choose to design it this might end up being very similar to the escape hatch we plan to provide connectors from the feature-protocol compatibility checks.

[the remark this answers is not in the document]
allisonport-db: If you have a different idea I'm open to doing this in a different way (for example, we could set `connectorReaderVersion` and `connectorWriterVersion` to higher values to achieve a similar affect). This was just following the semantics of testing Standalone directly (no connector) vs. Standalone + connector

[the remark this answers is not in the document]
allisonport-db: we can choose a different way to add/set this flag (although I do think it has to be a part of `DeltaLogImpl`)

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
scottsand-db: IMO the decision to or not to throw an error is at a much higher scope than `Metadata.java`. So I think we should not throw an error here. IMO the connector can do that themselves.

Perhaps we can add a `hasCheckConstraint: Boolean` API so they can efficiently look it up?

WDYT @venki?

standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
```diff
@@ -24,14 +24,17 @@ import com.fasterxml.jackson.core.JsonGenerator
...
+import io.delta.standalone.internal.DeltaConfigs
+import io.delta.standalone.internal.exception.DeltaErrors
 import io.delta.standalone.internal.util.{DataTypeParser, JsonUtils}
 
 private[internal] object Action {
   /** The maximum version of the protocol that this version of Delta Standalone understands. */
-  val readerVersion = 1
-  val writerVersion = 2
+  val readerVersion = 2
+  val writerVersion = 5
```
scottsand-db, line 37: can you remind me why this is safe to do? why can't we do it int by int ... feature by feature?

allisonport-db: Do you mean: as we add features, increase this? I considered this, but since we are implementing features non-linearly we would need a way to override this for testing purposes and that would be more complicated.

I think this is safe to do since we've set it up such that all public API behavior is unchanged (there are tests for this) and we don't plan to do a release or add the public APIs to read higher-protocol tables until all the features have been added.

allisonport-db: I think the alternative ways to allow testing a single feature non-linearly would be much more complicated.

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
```diff
@@ -277,8 +298,15 @@ private[internal] class OptimisticTransactionImpl(
 
     val protocolOpt = finalActions.collectFirst{ case p: Protocol => p }
     if (protocolOpt.isDefined) {
-      assert(protocolOpt.get == Protocol(), s"Invalid Protocol ${protocolOpt.get.simpleString}. " +
-        s"Currently only Protocol readerVersion 1 and writerVersion 2 is supported.")
```
allisonport-db, line 281: Should we fail if we see `writerVersion < 2` or will we now allow this?

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
```diff
@@ -224,6 +225,23 @@ private[internal] class OptimisticTransactionImpl(
...
+  // Internal Test-Only Methods
+  ///////////////////////////////////////////////////////////////////////////
+
+  // This is a lightweight version of the eventual public upgrade protocol by version API. When this
+  // is formally added it will be added to OptimisticTransaction and additional checks will be
+  // performed
+  private[standalone] def upgradeProtocolVersion(readerVersion: Int, writerVersion: Int): Unit = {
+      // deltaLog.assertProtocolWrite(Protocol(readerVersion, writerVersion))
+      // deltaLog.assertProtocolRead(Protocol(readerVersion, writerVersion))
+    assert(readerVersion <= Action.readerVersion && writerVersion <= Action.writerVersion,
```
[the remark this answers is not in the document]
allisonport-db: #452

vkorukanti:

PR #454, opened 2022-10-24
Internal API for reading/writing to higher protocol tables and creating higher protocol tables

allisonport-db:
## Overview

Adds internal APIs for testing purposes to allow creating tables of higher-protocol versions, and allow reading/writing to them. All public APIs should be unaffected and will still only support Protocol(1, 2).

### Reading/writing to higher-protocol tables
We introduce the separate concepts of standalone's supported protocol version vs. the connector's supported protocol version. To read/write to a table standalone's versions must be >= the table's protocol version _and_ the connector's versions must be >= the table's protocol version.

For now we add `connectorReaderVersion` and `connectorWriterVersion` to `DeltaLogImpl` and allow setting them internally through `DeltaLogImpl.forTable(...)` (but do NOT add it to the public API.) These are temporary and in the future will be replaced with `connectorSupportedReaderFeatures` and `connectorSupportedWriterFeatures`.

### Creating higher-protocol tables

We introduce `OptimisticTransactionImpl.upgradeProtocolVersion` which enables creating a new table with a higher protocol version than (1,2). This is not added to the public OptimisticTransaction interface--all usages will need to first cast `txn.asInstanceOf[OptimisticTransactionImpl]`

## How will these APIs be used?

We can categorize the two types of tests we will see:
1) Testing the connector protocol checks ([example](https://github.com/delta-io/connectors/pull/454/commits/973e45a5db10fee01128181f8b067a1b6b9373a6#diff-0ea3a9c91f6b9903d13d78f1fb2759d0ff3b43a6a30550547795761624c6af0dL367)). Do we correctly use the provided `supportedReaderFeatures`/`supportedWriterFeatures` to fail? (this can be failing reads/writes, failing `enableFeatureInProtocol` etc)
2) Testing feature APIs ([example](https://github.com/delta-io/connectors/pull/454/files#diff-0ea3a9c91f6b9903d13d78f1fb2759d0ff3b43a6a30550547795761624c6af0dR326)). As an example, do we verify the metadata correctly?

I've added an example of (1) to the PR but will remove before merge.

### Test utils

To facilitate tests of type (2) we add `getDeltaLogWithStandaloneAsConnector` to `TestUtils`.
- Once we switch to `connectorSupportedReaderFeatures` and `connectorSupportedWriterFeatures` we will update `getDeltaLogWithStandaloneAsConnector` to set the connector's supported feature sets to equal standalone's supported feature sets

I've also added `testException` to `TestUtils`.
  - TODO: as a followup refactor PR, update other tests to use `TestUtils`


standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionSuite.scala
```diff
@@ -316,4 +318,88 @@ class OptimisticTransactionSuite extends OptimisticTransactionSuiteBase {
...
+  // prepareCommit() protocol checks
+  ///////////////////////////////////////////////////////////////////////////
+
+  test("appendOnly: metadata-protocol compatibility checks") {
+    val schema = new StructType(Array(new StructField("col1", new IntegerType(), true)))
+
+    // cannot set appendOnly=true with too low a protocol version
+    withTempDir { dir =>
+      val log = getDeltaLogWithStandaloneAsConnector(new Configuration(), dir.getCanonicalPath)
+      val txn = log.startTransactionWithInitialProtocolVersion(1, 1)
```
scottsand-db, line 332: this (1, 1) protocol version isn't a "higher" protocol version, as per the method docs of `startTransactionWithInitialProtocolVersion`.

do we need to use this API here?

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
```diff
@@ -183,6 +191,25 @@ private[internal] class DeltaLogImpl private(
   // Internal Methods
   ///////////////////////////////////////////////////////////////////////////
 
+  // This is a temporary private API to create tables with a higher protocol version to allow
+  // implementing and testing higher-protocol features before we add the public APIs to upgrade
+  // table protocol versions.
+  private[standalone] def startTransactionWithInitialProtocolVersion(
```
scottsand-db, line 197: what will the code look like when this API is removed? what will change?

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
```diff
@@ -183,6 +191,25 @@ private[internal] class DeltaLogImpl private(
   // Internal Methods
   ///////////////////////////////////////////////////////////////////////////
 
+  // This is a temporary private API to create tables with a higher protocol version to allow
```
scottsand-db, line 194: `internal` API

standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
```diff
@@ -26,12 +26,16 @@ import com.fasterxml.jackson.databind.annotation.{JsonDeserialize, JsonSerialize
...
+import io.delta.standalone.internal.DeltaConfigs
+import io.delta.standalone.internal.exception.DeltaErrors
 import io.delta.standalone.internal.util.{DataTypeParser, JsonUtils}
 
 private[internal] object Action {
   /** The maximum version of the protocol that this version of Delta Standalone understands. */
-  val maxSupportedReaderVersion = 1
-  val maxSupportedWriterVersion = 2
+  // Note: features between version (1,2) and (2,5) are in development. We bump to (2,5) now
+  // since the features are not being implemented linearly. This does not affect any public APIs.
```
scottsand-db, line 36: and nothing will be merged to master until all features have been completed/suppported

standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
```diff
@@ -61,8 +65,8 @@ private[internal] sealed trait Action {
  * fields that they do not understand.
  */
 private[internal] case class Protocol(
-    minReaderVersion: Int = Action.maxSupportedReaderVersion,
-    minWriterVersion: Int = Action.maxSupportedWriterVersion) extends Action {
```
scottsand-db, line 65: why this change?

standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionSuite.scala
```diff
@@ -316,4 +318,88 @@ class OptimisticTransactionSuite extends OptimisticTransactionSuiteBase {
...
+
+  ///////////////////////////////////////////////////////////////////////////
+  // TEMPORARY (will be removed before merging)
+  ///////////////////////////////////////////////////////////////////////////
+
+  // This is to demonstrate the two types of tests we will have:
+  // (1) Testing the connector protocol checks. Do we correctly use the provided
+  //     supportedReaderFeatures/supportedWriterFeatures to fail? (this can be failing reads/writes,
+  //     failing enableFeatureInProtocol etc)
+  // (2) Testing feature APIs. As an example, do we verify the metadata correctly?
```
scottsand-db, line 373: can't each of these go into separate test suites? would that be a clean organization?

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
scottsand-db, line 197: what will the public API be and why can't we implement it now?

standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionSuite.scala
allisonport-db: If we don't do this, we create a table with Protocol(1, 2) by default currently. Updating our default protocol will be a part of the later protocol API updates.

allisonport-db: I can add the docs to be more specific. As in, it's a method to specify the initial protocol version for a newly created table.

standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
allisonport-db: All the usages of `Protocol()` in the code were now using `Protocol(2, 5)` when they expected `Protocol(1,2)`. There's also no real reason to justify this default being a specific version now that we are supporting more than a single protocol and it makes the code more confusing / less-self-documenting. I.e. seeing `Protocol()` in the code is confusing and ambiguous

standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionSuite.scala
allisonport-db: I think (1) will likely go into a separate suite. I was thinking (2) will likely be more broken down by feature and go in feature-specific suites. `appendOnly` is a weird case where we already support it and are now adding the check

allisonport-db: If you think the protocol-metadata compatibility checks should go all in the same suite for all the featues--that could be a valid organization and I could move the appendOnly test there

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
allisonport-db: Mentioned in the description above
```
txn.upgradeProtocolVersion(1, 3)
```
Although I'm now wondering whether it should be `setProtocolVersion` instead to account for the first version of the table

allisonport-db: somewhat answered above. we decided to implement the protocol-specific public APIs later. This will involve additional checks (like making sure connectors don't block themselves out) & is just additional work compared to the "lightweight" internal API we provide here

standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
allisonport-db: Is not changing any public APIs not sufficient?

scottsand-db: it may or may not be, but making it clear it won't be merged until finished removes that ambiguity.

e.g. what if it doesn't change any public API, but changes some behind-the-scenes internal implementation?

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
scottsand-db: if the future public API will be `txn.upgradeProtocolVersion(1, 3)`, why don't we just implement it now (public) but with your mock/lightweight impl?

that means all our tests can use the public API, and we can improve it gradually.

allisonport-db: I think we don't want to add public APIs without tests. I would prefer the extra annoyance of refactoring later

We could also leave `startTransactionWithInitialProtocolVersion` as an internal testing method in `DeltaLogImpl` and just update it to do the above snippet.

standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
allisonport-db: Err then if that affects a public API it would be changing the public API? I've updated it either way

standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionLegacySuite.scala
```diff
@@ -250,25 +251,27 @@ class OptimisticTransactionLegacySuite extends FunSuite {
     withTempDir { dir =>
       val log = DeltaLog.forTable(new Configuration(), dir.getCanonicalPath)
 
-      Seq(Protocol(1, 3), Protocol(1, 1), Protocol(2, 2)).foreach { protocol =>
```
allisonport-db, line 253: Here and below: Should we allow tables to be created with Protocol(1,1)? In the current version we have blocked this.

I understand (1, 2) being the default, but I'm unclear on whether we should explicitly block (1,1) (we allow it in delta-spark)

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
```diff
@@ -289,8 +319,6 @@ private[internal] class OptimisticTransactionImpl(
       case _ => // nothing
     }
 
-    deltaLog.assertProtocolWrite(snapshot.protocolScala)
```
allisonport-db, line 292: we do this when starting the transaction instead. we perform the exact same check then anyways we may as well fail earlier

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
vkorukanti: One qn @allisonport-db : when you say later, it will be before we merge the feature branch to master right?

allisonport-db: later = before we do a release with any of this code merged

scottsand-db: `setProtocolVersion` is better

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
```diff
@@ -224,6 +225,25 @@ private[internal] class OptimisticTransactionImpl(
     snapshot.transactions.getOrElse(id, -1L)
   }
 
+  ///////////////////////////////////////////////////////////////////////////
+  // Internal Test-Only Methods
+  ///////////////////////////////////////////////////////////////////////////
+
+  // This is a lightweight version of the eventual public upgrade protocol by version API. When this
+  // is formally added it will be added to OptimisticTransaction and additional checks will be
```
scottsand-db, line 233: what sort of additional checks? e.g. why doesn't this internal API check that the connector is upgrading to a new readerVersion, writerVersion that the connector itself can support?

Overall, it's hard to know if this method is purposefully or accidentally missing a validation/check

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
```diff
@@ -183,6 +191,25 @@ private[internal] class DeltaLogImpl private(
...
+      readerVersion: Int, writerVersion: Int): OptimisticTransaction = {
+
+    update()
+    assertProtocolWrite(snapshot.protocolScala)
+    if (snapshot.version >= 0) {
+      // To keep the scope small, we only support setting the initial protocol version and not
+      // upgrading the version for existing tables
+      throw new UnsupportedOperationException(
+        "Setting the initial protocol is only supported when committing table version 0")
+    }
```
scottsand-db, line 207: why this? the future public API won't have this, right? so why are we doing this now?

standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionLegacySuite.scala
scottsand-db: > In the current version we have blocked this.

then let's keep blocking this. we can make an issue and add it to the project plan to implement this later. will this decision make this PR smaller?

standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionSuite.scala
```diff
@@ -316,4 +318,88 @@ class OptimisticTransactionSuite extends OptimisticTransactionSuiteBase {
...
+
+  // The test below is an example of (1). The test above is an example of (2)
+
+  test("example of type (1) test") {
```
scottsand-db, line 377: IMO the test name should be more specific, and a comment above it can describe that it is an example of a type 1 test

standalone/src/test/scala/io/delta/standalone/internal/util/TestUtils.scala
```diff
@@ -47,4 +50,16 @@ object TestUtils {
...
 
+  def testException[T <: Throwable](f: => Any, messageContains: String)
+    (implicit manifest: Manifest[T]): Unit = {
+    val e = intercept[T]{
+      f
+    }.getMessage
+    assert(e.contains(messageContains))
+  }
+
+  def getDeltaLogWithStandaloneAsConnector(conf: Configuration, path: String) : DeltaLogImpl = {
```
scottsand-db, line 61: is this temporary?

also, method docs since this is an interesting / bold function name

standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionLegacySuite.scala
allisonport-db: It won't really make it smaller but it will reduce the scope. This can be a part of the general protocol changes we will make.

It will also make the `appendOnly` tests obsolete though, since we won't be creating any tables with Protocol(1,1)

standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionSuite.scala
allisonport-db: I was planning on removing this before merging the PR was just for demonstration purposes

standalone/src/test/scala/io/delta/standalone/internal/util/TestUtils.scala
allisonport-db: I think this makes sense to keep as a test util going forward. Once we switch from protocol version --> feature lists, this will set the connector's supported feature sets to equal standalone's supported feature sets. Ack on the method docs

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
vkorukanti: Once `txn.upgradeProtocolVersion(1, 3)` is ready, is this just to set the transaction on new table or can it be used to upgrade the protocol on existing table?

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
```diff
@@ -253,13 +298,27 @@ private[standalone] object DeltaLogImpl {
     apply(hadoopConf, new Path(dataPath, "_delta_log"), clock)
   }
 
+  def forTable(hadoopConf: Configuration, dataPath: String,
```
vkorukanti, line 301: Is this a public API? may need docs if it is.

standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
```diff
@@ -79,6 +84,32 @@ private[internal] object Protocol {
...
+
+    // Append-only
+    if (DeltaConfigs.IS_APPEND_ONLY.fromMetadata(metadata) && protocol.minWriterVersion < 2) {
+      // todo: when we add the feature enums we can use their string here (and all below)
+      throw DeltaErrors.insufficientWriterVersion(protocol, 2, "appendOnly")
+    }
+
+    // Check constraints
+
+    // Generated columns
```
vkorukanti, line 103: are all the features here can be verified using the properties in Metadata?

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
allisonport-db: so we don't have to implement upgrade checks + the tests for it. I can remove the check but initial protocol is sufficient for our tests (we may as well allow only what we need, not more)

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
allisonport-db: not public

standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
allisonport-db: yes, we'll basically do the same thing as in `Protocol.requiredMinimumProtocol` in https://github.com/delta-io/delta (and add each check as we add the feature)

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
allisonport-db: removed this in favor of just directly using `upgradeProtocolVersion`

standalone/src/main/scala/io/delta/standalone/internal/DeltaLogImpl.scala
scottsand-db: @vkorukanti this API is in `DeltaLogImpl.scala` and not `DeltaLog.java`. Internal tests will create `DeltaLogImpl`s.

the future public API will use features, not integer versions.

standalone/src/main/scala/io/delta/standalone/internal/OptimisticTransactionImpl.scala
```diff
@@ -277,8 +293,19 @@ private[internal] class OptimisticTransactionImpl(
 
     val protocolOpt = finalActions.collectFirst{ case p: Protocol => p }
     if (protocolOpt.isDefined) {
```
scottsand-db, line 295: nit: since we are do so many `protocolOpt.get` we may as well do
`protocolOpt.foreach { protocol => ... }`

standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionLegacySuite.scala
```diff
@@ -251,24 +252,26 @@ class OptimisticTransactionLegacySuite extends FunSuite {
       val log = DeltaLog.forTable(new Configuration(), dir.getCanonicalPath)
 
       Seq(Protocol(1, 3), Protocol(1, 1), Protocol(2, 2)).foreach { protocol =>
-        val e = intercept[AssertionError] {
+        val e = intercept[Throwable] {
```
scottsand-db, line 255: can we catch anything more specific?

standalone/src/test/scala/io/delta/standalone/internal/util/TestUtils.scala
scottsand-db: idea: `getDeltaLogWithMaxFeatureSupport` ? im easy.

scottsand-db (approved):
LGTM with one nit and one method rename suggestion.

standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionLegacySuite.scala
allisonport-db: Not really since we're explicitly blocking the writerVersion=1 with an assert. But we see `InvalidProtocolVersionException` for protocols > Protocol(1,2)

Way around this would be to test Protocol(1,3) and Protocol(2,2) separately from Protocol(1,1). I'm good with either here since once we later allow writerVersion=1 we can update this to be more specific

PR #455, opened 2022-10-25
CHECK constraints APIs for setting/removing/retrieving

allisonport-db:
Changes in this PR
- Adds Constraint.java to represent both check constraints and column invariants
- Adds `getConstraints`, `addCheckConstraint` and `removeCheckConstraint` to `Metadata.java`
- Adds tests in `DeltaConstraintsSuite`

### DIFF
--- a/standalone/src/main/java/io/delta/standalone/Constraint.java
+++ b/standalone/src/main/java/io/delta/standalone/Constraint.java
@@ -23,7 +23,7 @@ import io.delta.standalone.types.StructField;
  * Represents a constraint defined on a Delta table which writers must verify before writing.
  * Constraints can come in one of two ways:
  * - A CHECK constraint which is stored in {@link Metadata#getConfiguration()}. CHECK constraints
- *   are stored as the key-value pair "delta.constraints.{constraintName}" --> "{expression}"
+ *   are stored as the key-value pair ("delta.constraints.{constraintName}", "{expression}")
  * - A column invariant which is stored in {@link StructField#getMetadata()}
  *   TODO: provide more details here
  */
--- a/standalone/src/main/java/io/delta/standalone/Constraint.java
+++ b/standalone/src/main/java/io/delta/standalone/Constraint.java
@@ -29,11 +29,17 @@ import io.delta.standalone.types.StructField;
 /**
  * Represents a constraint defined on a Delta table which writers must verify before writing.
  * Constraints can come in one of two ways:
- * - A CHECK constraint which is stored in {@link Metadata#getConfiguration()}
+ * - A CHECK constraint which is stored in {@link Metadata#getConfiguration()}. CHECK constraints
+ *   are stored as the key-value pair "delta.constraints.{constraintName}" --> "{expression}"
  * - A column invariant which is stored in {@link StructField#getMetadata()}
+ *   TODO: provide more details here
  */
 public final class Constraint {
 
+    /**
+     * This is the key-prefix for the check constraint key "delta.constraints.{constraintName}" in
+     * {@link Metadata#getConfiguration()}
+     */
     public static final String CHECK_CONSTRAINT_KEY_PREFIX = "delta.constraints.";
     @Nonnull private final String name;
     @Nonnull private final String expression;
--- a/standalone/src/main/java/io/delta/standalone/Constraint.java
+++ b/standalone/src/main/java/io/delta/standalone/Constraint.java
@@ -62,6 +62,11 @@ public final class Constraint {
     }
 
     @Override
+    public String toString() {
+        return String.format("%s (%s)", name, expression);
+    }
+
+    @Override
     public boolean equals(Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
--- a/standalone/src/main/java/io/delta/standalone/Constraint.java
+++ b/standalone/src/main/java/io/delta/standalone/Constraint.java
@@ -16,15 +16,8 @@
 
 package io.delta.standalone;
 
-import java.util.Objects;
-
-import javax.annotation.Nonnull;
-
 import io.delta.standalone.actions.Metadata;
 import io.delta.standalone.types.StructField;
-
-// TODO: if we ever plan to support not-null constraints as a java class should we name this
-//  "CheckConstraint" even though we will also be presenting invariants in this form?
 
 /**
  * Represents a constraint defined on a Delta table which writers must verify before writing.
@@ -34,55 +27,15 @@ import io.delta.standalone.types.StructField;
  * - A column invariant which is stored in {@link StructField#getMetadata()}
  *   TODO: provide more details here
  */
-public final class Constraint {
-
-    /**
-     * This is the key-prefix for the check constraint key "delta.constraints.{constraintName}" in
-     * {@link Metadata#getConfiguration()}
-     */
-    public static final String CHECK_CONSTRAINT_KEY_PREFIX = "delta.constraints.";
-    @Nonnull private final String name;
-    @Nonnull private final String expression;
-
-    public Constraint(
-            @Nonnull String name,
-            @Nonnull String expression) {
-        this.name = name;
-        this.expression = expression;
-    }
+public interface Constraint {
 
     /**
      * @return the name of this constraint
      */
-    @Nonnull
-    public String getName() {
-        return name;
-    }
+    String getName();
 
     /**
      * @return the expression to enforce of this constraint as a SQL string
      */
-    @Nonnull
-    public String getExpression() {
-        return expression;
-    }
-
-    @Override
-    public String toString() {
-        return String.format("%s (%s)", name, expression);
-    }
-
-    @Override
-    public boolean equals(Object o) {
-        if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
-        Constraint constraint = (Constraint) o;
-        return Objects.equals(name, constraint.name) &&
-                Objects.equals(expression, constraint.expression);
-    }
-
-    @Override
-    public int hashCode() {
-        return Objects.hash(name, expression);
-    }
+    String getExpression();
 }
--- a/standalone/src/main/java/io/delta/standalone/Constraint.java
+++ b/standalone/src/main/java/io/delta/standalone/Constraint.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright (2020-present) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.delta.standalone;
+
+import java.util.Objects;
+
+import javax.annotation.Nonnull;
+
+import io.delta.standalone.actions.Metadata;
+import io.delta.standalone.types.StructField;
+
+// TODO: if we ever plan to support not-null constraints as a java class should we name this
+//  "CheckConstraint" even though we will also be presenting invariants in this form?
+
+/**
+ * Represents a constraint defined on a Delta table which writers must verify before writing.
+ * Constraints can come in one of two ways:
+ * - A CHECK constraint which is stored in {@link Metadata#getConfiguration()}
+ * - A column invariant which is stored in {@link StructField#getMetadata()}
+ */
+public final class Constraint {
+
+    public static final String CHECK_CONSTRAINT_KEY_PREFIX = "delta.constraints.";
+    @Nonnull private final String name;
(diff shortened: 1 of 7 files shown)


standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -133,6 +139,63 @@ public StructType getSchema() {
...
+    /**
+     * Returns a new {@link Metadata} instance with the same properties as this instance but with
+     * the specified check constraint added to the {@code configuration}
+     * @param name  the name of the check constraint
+     * @param expression  the condition to enforce as a SQL string
+     * @return a new {@link Metadata} instance with the same properties as this instance but with
+     *         the added check constraint
+     * @throws IllegalArgumentException if a constraint with name already exists
+     */
+    public Metadata addCheckConstraint(String name, String expression) throws Throwable {
```
allisonport-db, line 169: `name` can be anything if it's escaped
```
\\ ie. this is a valid name:
name = '`delta.constraints`'
```

scottsand-db:
Can you update the target feature branch to be `support_higher_protocol_versions_feature_branch`

standalone/src/main/java/io/delta/standalone/Constraint.java
```diff
@@ -0,0 +1,82 @@
...
+// TODO: if we ever plan to support not-null constraints as a java class should we name this
+//  "CheckConstraint" even though we will also be presenting invariants in this form?
+
+/**
+ * Represents a constraint defined on a Delta table which writers must verify before writing.
+ * Constraints can come in one of two ways:
+ * - A CHECK constraint which is stored in {@link Metadata#getConfiguration()}
+ * - A column invariant which is stored in {@link StructField#getMetadata()}
+ */
+public final class Constraint {
```
scottsand-db, line 35: do we ever expect or want OSS users to create their own constraint classes?

or do we just create a constraint instance .... and give them it... and they use it.

I'm wondering why a class instead of an interface.

standalone/src/test/scala/io/delta/standalone/internal/DeltaConstraintsSuite.scala
```diff
@@ -0,0 +1,189 @@
...
+    // check constraint key format
+    testGetConstraints(
+      Map(
+        // should be retrieved, preserves expression case
+        getCheckConstraintKey("constraints") -> "EXPRESSION",
+        getCheckConstraintKey("delta.constraints") ->
+          "expression0",
+        getCheckConstraintKey(Constraint.CHECK_CONSTRAINT_KEY_PREFIX) ->
+          "expression1",
+        // should not be retrieved
```
scottsand-db, line 71: add a comment here (just one, not one per line) explaining why

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -133,6 +139,63 @@ public StructType getSchema() {
...
+                                entry.getKey().replaceFirst(prefixRegex, ""),
+                                entry.getValue()))
+                .collect(Collectors.toList());
+
+        return checkConstraints;
+    }
+
+    /**
+     * Returns a new {@link Metadata} instance with the same properties as this instance but with
+     * the specified check constraint added to the {@code configuration}
```
scottsand-db, line 162: IMO it would be worth linking to the delta-core docs for check constraints.

Maybe even describe here in 1 line (TLDR) what a check constraint is.

(why? -> because even to me it's not super clear what the difference is)

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -133,6 +139,63 @@ public StructType getSchema() {
...
+     *         the specified check constraint removed
+     * @throws IllegalArgumentException if a constraint with name does not exist
+     */
+    public Metadata removeCheckConstraint(String name) throws Throwable {
+        String fullKey = Constraint.CHECK_CONSTRAINT_KEY_PREFIX + name.toLowerCase(Locale.ROOT);
+        if (!configuration.containsKey(fullKey)) {
+            // todo: decide if we want to throw an error here
+            throw DeltaErrors.checkConstraintDoesNotExist(name);
+        }
+        Map<String, String> newConfiguration = new HashMap(configuration);
```
scottsand-db, line 194: From about 5 mins of searching online ... it seems this doesn't do a deep clone. Let me know if I'm wrong.

https://www.baeldung.com/java-copy-hashmap

Also, please put in the method docs if this is intended to be a deep or shallow clone.

scottsand-db (changes requested):

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -133,6 +139,63 @@ public StructType getSchema() {
...
     }
 
+    /**
+     * @return the CHECK constraints and column invariants defined in this metadata
+     */
+    public List<Constraint> getConstraints() {
+        // todo: get column invariants
+
+        // get check constraints
+        String prefixRegex = "^" + Constraint.CHECK_CONSTRAINT_KEY_PREFIX;
```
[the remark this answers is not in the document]
allisonport-db: yeah good catch

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
allisonport-db: So it doesn't do a deep clone, but since configuration is Map<String, String> and Strings are immutable essentially shallow clone = deep clone (for the purpose of changing one does not modify the other). I also just ran some checks to verify this for sure

allisonport-db: honestly it was probably an initial oversight that `copyBuilder` doesn't do a deep copy of everything... this is something we should address what do you think?

standalone/src/main/java/io/delta/standalone/Constraint.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public static final String CHECK_CONSTRAINT_KEY_PREFIX = "delta.constraints.";
```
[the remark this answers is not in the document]
allisonport-db: yeah sgtm this should be clarified here and in the @param docs for the get/remove methods

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
allisonport-db: oh I remember now...  check constraints are not documented currently in Protocol.md (is this what you were thinking of in terms of docs?). I'll add a short explanation and follow up on adding it to the protocol

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
scottsand-db: >  but since configuration is Map<String, String> and Strings are immutable essentially shallow clone = deep clone

SGTM.

for copybuilder, I wouldn't worry about it now.

standalone/src/main/java/io/delta/standalone/Constraint.java
```diff
@@ -0,0 +1,88 @@
...
+ * - A column invariant which is stored in {@link StructField#getMetadata()}
+ *   TODO: provide more details here
+ */
+public final class Constraint {
+
+    /**
+     * This is the key-prefix for the check constraint key "delta.constraints.{constraintName}" in
+     * {@link Metadata#getConfiguration()}
+     */
+    public static final String CHECK_CONSTRAINT_KEY_PREFIX = "delta.constraints.";
```
allisonport-db, line 43: I'm thinking about moving this to the `Metadata` class what do you think?

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
allisonport-db: Added a short description. Once I update `Protocol.md` make sense to link it somewhere here

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -133,6 +138,59 @@ public StructType getSchema() {
...
+     *              "delta.constraints" prefix)
+     * @param expression  the condition to enforce as a SQL string
+     * @return a new {@link Metadata} instance with the same properties as this instance but with
+     *         the added check constraint
+     * @throws IllegalArgumentException if a constraint with name already exists
+     */
+    public Metadata withCheckConstraint(String name, String expression) throws Throwable {
+        String fullKey = ConstraintImpl.getCheckConstraintKey(name);
+        if (configuration.containsKey(fullKey)) {
+            // todo: decide if we want to throw an error here
```
allisonport-db, line 165: @vkorukanti @scottsand-db should we discuss this? I think @vkorukanti you commented on #456 that this is fine

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -133,6 +138,59 @@ public StructType getSchema() {
...
+     * @param name  the case-insensitive name of the check constraint to remove (without the
+     *              "delta.constraints" prefix)
+     * @return a new {@link Metadata} instance with the same properties as this instance but with
+     *         the specified check constraint removed
+     * @throws IllegalArgumentException if a constraint with name does not exist
+     */
+    public Metadata withoutCheckConstraint(String name) throws Throwable {
+        final String fullKey = ConstraintImpl.getCheckConstraintKey(name);
+        if (!configuration.containsKey(fullKey)) {
+            // todo: decide if we want to throw an error here
```
allisonport-db, line 186: @scottsand-db @vkorukanti same here

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
scottsand-db: what do you think is the most common use case of this API? will they want to override by default? or do they NEVER want to accidentally override? I think answers to those questions will help us get our answer here.

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
scottsand-db: if it doesn't contain it ... then no need to throw an error.

scottsand-db: think of it from an API guarantee perspective: we guarantee that the metadata we return will not contain this check constraint.

the reason `withCheckConstraint` is different is because the user may or may not want to override the existing constraint.

but we don't need to worry about that here.

standalone/src/main/scala/io/delta/standalone/internal/ConstraintImpl.scala
```diff
@@ -0,0 +1,81 @@
...
+
+private[standalone] object ConstraintImpl {
+
+  /**
+   * Extracts constraints from the table properties. These include both CHECK constraints stored in
+   * the [[Metadata#getConfiguration( )]] and column invariants stored in the
+   * [[StructField#getMetadata( )]].
+   */
+  def getConstraints(metadata: Metadata): java.util.List[Constraint] = {
+    // todo: get column invariants
```
scottsand-db, line 47: will this be done in another PR? can you add an "out of scope" section in your PR description listing all of the TODOs in this PR/future work that this pr has introduced?

or, if you think of a better place or way to present that information ^, then that is fine with me too.

standalone/src/main/scala/io/delta/standalone/internal/ConstraintImpl.scala
```diff
@@ -0,0 +1,81 @@
...
+  /**
+   * The [[Metadata#getConfiguration( )]] key for a check constraint with `name`. Keys are always
+   * stored in the metadata in lowercase.
+   */
+  def getCheckConstraintKey(name: String): String = {
+    CHECK_CONSTRAINT_KEY_PREFIX + name.toLowerCase(Locale.ROOT)
+  }
+
+  private[internal] def getCheckConstraints(configuration: Map[String, String])
+      : Seq[Constraint] = {
```
scottsand-db, line 72: nit: please not this syntax it's ugly :D

vkorukanti (approved):
lgtm. Given that this PR is going to be in a feature branch, the pending TODOs can be addressed in separate PRs.

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
allisonport-db: Do we expect our connectors to have the same behavior as the delta-spark SQL APIs for adding/removing check constraints? If we do, I think we should throw errors here (and below) and that error will just be propagated to the end user.

If we think connectors might not necessarily implement the same behavior as the SQL APIs, then this behavior is up for debate (and the `withoutCheckConstraint` should not throw an error) and we should expose an API `containsCheckConstraint` or something along those lines.

standalone/src/main/scala/io/delta/standalone/internal/ConstraintImpl.scala
allisonport-db: this is in #456 and is really just a placeholder

I can create a github issue with a checklist if that will help clarify everything (coming when I get the chance)

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
scottsand-db: Sticking to the SQL API semantics (i.e. throw an error) SGTM

scottsand-db (approved):

PR #456, opened 2022-10-26
Column Invariants

allisonport-db:
Changes in this PR:
- Adds column invariants to `Metadata::getConstraints()`
   - Adds InvariantUtils and adds to SchemaUtils which is 95% copied from Invariants - [here](https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/constraints/Invariants.scala) and SchemaUtils - [here](https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/schema/SchemaUtils.scala#L55)
- Adds tests to `DeltaConstraintsSuite`

*Since column invariants are deprecated we don't provide a public API to set them.

### DIFF
--- a/standalone/src/main/java/io/delta/standalone/actions/Metadata.java
+++ b/standalone/src/main/java/io/delta/standalone/actions/Metadata.java
@@ -31,6 +31,8 @@ import io.delta.standalone.types.StructType;
 
 import io.delta.standalone.internal.ConstraintImpl;
 import io.delta.standalone.internal.exception.DeltaErrors;
+import io.delta.standalone.internal.util.InvariantUtils;
+
 
 /**
  * Updates the metadata of the table. The first version of a table must contain
--- a/standalone/src/main/java/io/delta/standalone/actions/Metadata.java
+++ b/standalone/src/main/java/io/delta/standalone/actions/Metadata.java
@@ -31,7 +31,6 @@ import io.delta.standalone.types.StructType;
 
 import io.delta.standalone.internal.ConstraintImpl;
 import io.delta.standalone.internal.exception.DeltaErrors;
-import io.delta.standalone.internal.util.InvariantUtils;
 
 /**
  * Updates the metadata of the table. The first version of a table must contain
--- a/standalone/src/main/java/io/delta/standalone/actions/Metadata.java
+++ b/standalone/src/main/java/io/delta/standalone/actions/Metadata.java
@@ -33,7 +33,6 @@ import io.delta.standalone.internal.ConstraintImpl;
 import io.delta.standalone.internal.exception.DeltaErrors;
 import io.delta.standalone.internal.util.InvariantUtils;
 
-
 /**
  * Updates the metadata of the table. The first version of a table must contain
  * a {@link Metadata} action. Subsequent {@link Metadata} actions completely
--- a/standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
@@ -28,7 +28,7 @@ import io.delta.standalone.types.StructType
 
 import io.delta.standalone.internal.{ConstraintImpl, DeltaColumnMappingMode, DeltaConfigs}
 import io.delta.standalone.internal.exception.DeltaErrors
-import io.delta.standalone.internal.util.{DataTypeParser, JsonUtils}
+import io.delta.standalone.internal.util.{DataTypeParser, InvariantUtils, JsonUtils}
 
 private[internal] object Action {
   /** The maximum version of the protocol that this version of Delta Standalone understands. */
@@ -100,7 +100,9 @@ private[internal] object Protocol {
     // look at Protocol.requiredMinimumProtocol in Delta
 
     // Column invariants
-    // check for invariants in the schema
+    if (InvariantUtils.getFromSchema(metadata.schema).size > 0 && protocol.minWriterVersion < 2) {
+      throw DeltaErrors.insufficientWriterVersion(protocol, 2, "columnInvariants")
+    }
 
     // Append-only
     val appendOnlyMinProtocol = Protocol(0, 2)
--- a/standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
@@ -89,7 +89,12 @@ private[internal] object Protocol {
       s"protocol version ($MIN_WRITER_VERSION_PROP) as part of table properties")
   }
 
-  /** Check that the protocol is compatible with any features enabled in the table metadata */
+  /**
+   * Checks that the table protocol is sufficient for any features enabled in the table metadata.
+   * For writer features, we check for sufficient writer version. For reader features, we check
+   * for sufficient reader version. And for reader and writer features, we check for sufficient
+   * reader and writer version.
+   */
   def checkMetadataFeatureProtocolCompatibility(metadata: Metadata, protocol: Protocol): Unit = {
 
     def insufficientProtocol(requiredProtocol: Protocol): Boolean = {
--- a/standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
@@ -105,8 +105,11 @@ private[internal] object Protocol {
     // look at Protocol.requiredMinimumProtocol in Delta
 
     // Column invariants
-    if (InvariantUtils.getFromSchema(metadata.schema).size > 0 && protocol.minWriterVersion < 2) {
-      throw DeltaErrors.insufficientWriterVersion(protocol, 2, "columnInvariants")
+    val columnInvariantsMinProtocol = Protocol(0, 2)
+    if (InvariantUtils.getFromSchema(metadata.schema).size > 0 &&
+      insufficientProtocol(columnInvariantsMinProtocol)) {
+      throw DeltaErrors.insufficientTableProtocolVersion(protocol,
+        columnInvariantsMinProtocol, "columnInvariants")
     }
 
     // Append-only
--- a/standalone/src/main/scala/io/delta/standalone/internal/util/InvariantUtils.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/util/InvariantUtils.scala
@@ -21,6 +21,7 @@ import scala.collection.JavaConverters._
 import io.delta.standalone.Constraint
 import io.delta.standalone.types.StructType
 
+import io.delta.standalone.internal.ConstraintImpl
 import io.delta.standalone.internal.exception.DeltaErrors
 
 /**
@@ -60,14 +61,18 @@ private[internal] object InvariantUtils {
     columns.map {
       case (_, field) =>
         val rule = field.getMetadata.get(INVARIANTS_FIELD).asInstanceOf[String]
-        val invariant = Option(JsonUtils.mapper.readValue[PersistedRule](rule).unwrap) match {
-          case Some(PersistedExpression(exprString)) =>
-            exprString
-          case _ =>
-             // todo: clarify when we might hit this?
-             throw DeltaErrors.unrecognizedInvariant()
+        try {
+          val invariant = Option(JsonUtils.mapper.readValue[PersistedRule](rule).unwrap) match {
+            case Some(PersistedExpression(exprString)) =>
+              exprString
+            case _ =>
+              throw DeltaErrors.misformattedInvariant(rule)
+          }
+          ConstraintImpl(s"EXPRESSION($invariant)", invariant)
+        } catch {
+          case e: com.fasterxml.jackson.databind.JsonMappingException =>
+            throw DeltaErrors.misformattedInvariant(rule)
         }
-        new Constraint(s"EXPRESSION($invariant)", invariant)
     }
   }
 
--- a/standalone/src/main/scala/io/delta/standalone/internal/util/InvariantUtils.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/util/InvariantUtils.scala
@@ -1,0 +1,75 @@
+/*
+ * Copyright (2020-present) The Delta Lake Project Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.delta.standalone.internal.util
+
+import scala.collection.JavaConverters._
+
+import io.delta.standalone.Constraint
+import io.delta.standalone.types.StructType
+
+import io.delta.standalone.internal.exception.DeltaErrors
+
+/**
+ * Invariant utils for retrieving column invariants from the schema.
+ */
+private[internal] object InvariantUtils {
+  sealed trait Rule {
+    val name: String
+  }
+
+  sealed trait RulePersistedInMetadata {
+    def wrap: PersistedRule
+    def json: String = JsonUtils.toJson(wrap)
+  }
+
+  /** Rules that are persisted in the metadata field of a schema. */
+  case class PersistedRule(expression: PersistedExpression = null) {
+    def unwrap: RulePersistedInMetadata = {
+      if (expression != null) {
+        expression
+      } else {
+        null
+      }
+    }
+  }
+
+  /** Persisted companion of the ArbitraryExpression rule. */
+  case class PersistedExpression(expression: String) extends RulePersistedInMetadata {
+    override def wrap: PersistedRule = PersistedRule(expression = this)
+  }
+
+  /** Extract column invariants from the given schema */
+  def getFromSchema(schema: StructType): Seq[Constraint] = {
+    val columns = SchemaUtils.filterRecursively(schema, checkComplexTypes = false) { field =>
+      field.getMetadata.contains(INVARIANTS_FIELD)
+    }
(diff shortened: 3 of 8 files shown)


standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionSuite.scala
```diff
@@ -316,4 +318,50 @@ class OptimisticTransactionSuite extends OptimisticTransactionSuiteBase {
       assert(committedAddFile.getPath === "file:/absolute/path/to/file/test.parquet")
     }
   }
+
+  ///////////////////////////////////////////////////////////////////////////
+  // prepareCommit() protocol checks
+  ///////////////////////////////////////////////////////////////////////////
+
+  // TODO: unblock these tests once we allow writerVersion=1
```
allisonport-db, line 326: I commented out the blocking code for writerVersion=1 and ran this test

standalone/src/main/scala/io/delta/standalone/internal/util/InvariantUtils.scala
```diff
@@ -0,0 +1,79 @@
...
+
+import io.delta.standalone.Constraint
+import io.delta.standalone.types.StructType
+
+import io.delta.standalone.internal.exception.DeltaErrors
+
+/**
+ * Invariant utils for retrieving column invariants from the schema.
+ */
+private[internal] object InvariantUtils {
```
allisonport-db, line 29: 95% copied from https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/constraints/Invariants.scala

standalone/src/main/scala/io/delta/standalone/internal/util/SchemaUtils.scala
```diff
@@ -136,6 +136,36 @@ private[standalone] object SchemaUtils {
...
 
+  /**
+   * Finds `StructField`s that match a given check `f`. Returns the path to the column, and the
+   * field.
+   *
+   * @param checkComplexTypes While `StructType` is also a complex type, since we're returning
+   *                          StructFields, we definitely recurse into StructTypes. This flag
+   *                          defines whether we should recurse into ArrayType and MapType.
+   */
+  def filterRecursively(
```
allisonport-db, line 147: Completely copied from https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/schema/SchemaUtils.scala#L55
except for minor API changes like `a.elementType` -> `a.getElementType` due to our Java getters

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -32,6 +33,8 @@
 import io.delta.standalone.types.StructType;
 
 import io.delta.standalone.internal.exception.DeltaErrors;
+import io.delta.standalone.internal.util.InvariantUtils;
+
```
scottsand-db, line 37: nitty nit: noise

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -154,7 +158,8 @@ public List<Constraint> getConstraints() {
                                 entry.getValue()))
                 .collect(Collectors.toList());
 
-        return checkConstraints;
+        return Stream.concat(invariantConstraints.stream(), checkConstraints.stream())
```
scottsand-db, line 161: do we want this return value to be modifiable? else we should probably do the `Collections.unmodifiableList(someList)`

standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
```diff
@@ -91,7 +91,9 @@ private[internal] object Protocol {
     // look at Protocol.requiredMinimumProtocol in Delta
 
     // Column invariants
-    // check for invariants in the schema
+    if (InvariantUtils.getFromSchema(metadata.schema).size > 0 && protocol.minWriterVersion < 2) {
+      throw DeltaErrors.insufficientWriterVersion(protocol, 2, "columnInvariants")
```
scottsand-db, line 95: `Protocol::checkMetadataFeatureProtocolCompatibility` --> is this for writers only? you are throwing `insufficientWriterVersion`?

maybe this is obvious and I am missing it but ... can we update the method doc to make this clearer?

standalone/src/test/scala/io/delta/standalone/internal/DeltaConstraintsSuite.scala
```diff
@@ -186,4 +194,149 @@ class DeltaConstraintsSuite extends FunSuite {
         Seq(new Constraint("test", "col1 < 0")))
     }
   }
+
+  ///////////////////////////////////////////////////////////////////////////
+  // column invariants
+  ///////////////////////////////////////////////////////////////////////////
+
+  def fieldMetadataWithInvariant(expr: String): FieldMetadata = {
```
scottsand-db, line 202: private

standalone/src/test/scala/io/delta/standalone/internal/DeltaConstraintsSuite.scala
```diff
@@ -186,4 +194,149 @@ class DeltaConstraintsSuite extends FunSuite {
...
+        .add("col2", new IntegerType()),
+      expectedConstraints = Seq.empty
+    )
+
+    // top-level column with a column invariant
+    val structField1 = new StructField("col1", new IntegerType(), true,
+      fieldMetadataWithInvariant("col1 < 3"))
+    testGetConstraints(
+      schema = new StructType().add(structField1),
+      expectedConstraints = Seq(new Constraint("EXPRESSION(col1 < 3)", "col1 < 3"))
```
scottsand-db, line 225: so the _name_ of this is `EXPRESSION(col1 < 3)`? is that what Delta-Spark does? I'd expect `name` to be something more human readable.

scottsand-db (changes requested):
Minor comments.

standalone/src/test/scala/io/delta/standalone/internal/DeltaConstraintsSuite.scala
allisonport-db: Yes https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/constraints/Invariants.scala#L57

standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
allisonport-db: So it depends on the feature. For writer-features it will throw `insufficientWriterVersion`. For reader _and_ writer features it can throw `insufficientProtocol`. For reader-features it will throw `insufficientReaderVersion`.

allisonport-db: I can update the method docs

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -133,6 +138,58 @@ public StructType getSchema() {
...
+     * A check constraint's {@code expression} must be enforced for each input row when writing data
+     * and all for existing rows.
+     *
+     * @param name  the name of the check constraint (without the "delta.constraints" prefix)
+     * @param expression  the condition to enforce as a SQL string
+     * @return a new {@link Metadata} instance with the same properties as this instance but with
+     *         the added check constraint
+     * @throws IllegalArgumentException if a constraint with name already exists
+     */
+    public Metadata addCheckConstraint(String name, String expression) throws Throwable {
```
vkorukanti, line 161: nit: The method looks like we are to an existing Metadata. Usually`withCheckConstraint` seems to more appropriate given that it is returning a new Metadata object.

This method and the below one seem like a good candidates on the Builder object itself? Wdyt?
`metatadata.copyBuilder().addCheckConstraint().build()`
`metadata.copyBuilder().removeCheckConstraint().build()`

standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
```diff
@@ -85,12 +85,19 @@ private[internal] object Protocol {
...
+   * For writer features, we check for sufficient writer version. For reader features, we check
+   * for sufficient reader version. And for reader and writer features, we check for sufficient
+   * reader and writer version.
+   */
   def checkMetadataFeatureProtocolCompatibility(metadata: Metadata, protocol: Protocol): Unit = {
     // look at Protocol.requiredMinimumProtocol in Delta
 
     // Column invariants
-    // check for invariants in the schema
+    if (InvariantUtils.getFromSchema(metadata.schema).size > 0 && protocol.minWriterVersion < 2) {
```
vkorukanti, line 98: nit: add constants for required min reader and writer version.

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -133,6 +138,58 @@ public StructType getSchema() {
...
+        String fullKey = ConstraintImpl.getCheckConstraintKey(name);
+        if (configuration.containsKey(fullKey)) {
+            // todo: decide if we want to throw an error here
```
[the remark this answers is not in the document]
allisonport-db: Delta-spark throws an error.

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -133,6 +138,58 @@ public StructType getSchema() {
...
+     * Returns a new {@link Metadata} instance with the same properties as this instance but with
+     * the specified check constraint removed from this instance's {@code configuration}
+     *
+     * @param name  the name of the check constraint to remove (without the "delta.constraints"
+     *              prefix)
+     * @return a new {@link Metadata} instance with the same properties as this instance but with
+     *         the specified check constraint removed
+     * @throws IllegalArgumentException if a constraint with name does not exist
+     */
+    public Metadata removeCheckConstraint(String name) throws Throwable {
```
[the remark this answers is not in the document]
allisonport-db: Not case sensitive, I will add this to the method docs. Thanks :)

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
allisonport-db: I like moving the methods to the builder

allisonport-db: Updated in #455

standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
allisonport-db: #455

standalone/src/main/scala/io/delta/standalone/internal/util/SchemaUtils.scala
```diff
@@ -134,14 +135,47 @@ private[standalone] object SchemaUtils {
             // if existing value is nullable, so should be the new value
             && (!existingField.isNullable || newField.isNullable)
             // and the type of the field must be compatible, too
-            && isDatatypeWriteCompatible(existingField.getDataType, newField.getDataType))
+            && isDatatypeWriteCompatible(existingField.getDataType, newField.getDataType)
+            // and no column invariant is added or altered
+            && Option(newField.getMetadata.get(InvariantUtils.INVARIANTS_FIELD)).forall(
```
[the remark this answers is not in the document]
allisonport-db: To avoid blocking _removing_ a column invariant. We don't provide any explicit APIs to do this, but it seems like something we might not want to explicitly block.

What do you think?

standalone/src/test/scala/io/delta/standalone/internal/DeltaConstraintsSuite.scala
```diff
@@ -190,4 +199,196 @@ class DeltaConstraintsSuite extends FunSuite {
...
+    testGetConstraints(
+      schema = new StructType().add(
+        new StructField(
+          "nested",
+          new StructType().add(nestedStructField),
+          true,
+          fieldMetadataWithInvariant("nested is null")
+        )),
+      expectedConstraints = Seq(ConstraintImpl("EXPRESSION(nested.col1 < 3)", "nested.col1 < 3"),
+        ConstraintImpl("EXPRESSION(nested is null)", "nested is null"))
```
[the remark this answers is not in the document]
allisonport-db: Sure that's more realistic :)

standalone/src/test/scala/io/delta/standalone/internal/DeltaConstraintsSuite.scala
```diff
@@ -190,4 +199,196 @@ class DeltaConstraintsSuite extends FunSuite {
...
+    testGetConstraints(
+      schema = new StructType()
+        .add("col1", new IntegerType())
+        .add("col2", new IntegerType()),
+      expectedConstraints = Seq.empty
+    )
+
+    // top-level column with a column invariant
+    val structField1 = new StructField("col1", new IntegerType(), true,
+      fieldMetadataWithInvariant("col1 < 3"))
```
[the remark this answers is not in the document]
scottsand-db: this wouldn't be part of `StructField` it would be a local private helper method in this test suite. e.g. see Venki's tests, he uses helpers like this to reduce verbosity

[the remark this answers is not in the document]
scottsand-db: but this isn't a big deal the code is still readable

standalone/src/test/scala/io/delta/standalone/internal/DeltaConstraintsSuite.scala
```diff
@@ -190,4 +199,196 @@ class DeltaConstraintsSuite extends FunSuite {
...
+  test("cannot add or change a column invariant in the schema") {
+    val structField = new StructField("col1", new IntegerType())
+    val schemaWithNoInvariant = new StructType().add(structField)
+    val schemaWithInvariant1 = new StructType().add(
+      new StructField(structField.getName, structField.getDataType, structField.isNullable,
+        fieldMetadataWithInvariant("col1 > 1"))
+    )
+    val schemaWithInvariant2 = new StructType().add(
+      new StructField(structField.getName, structField.getDataType, structField.isNullable,
+        fieldMetadataWithInvariant("col1 > 2"))
```
scottsand-db, line 357: how will users be using this new public `Constraint` API?

e.g. here you have created a helper method to create an invariant. is that what we expect users to do?

scottsand-db (approved):
Looks good to me except for 1 remaining comment in SchemaUtils. I do want to keep track of these public API changes we are making, and perhaps starting to add examples so we keep track of the big picture (i.e. the public API, and how it will be used) to make sure we have a solid user journey using these new features.

standalone/src/main/scala/io/delta/standalone/internal/util/SchemaUtils.scala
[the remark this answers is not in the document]
scottsand-db: Don't think I understand your answer.

I'm specifically referring to 

```scala
Option(newField.getMetadata.get(InvariantUtils.INVARIANTS_FIELD)).forall(_ == existingField.getMetadata.get(InvariantUtils.INVARIANTS_FIELD)))
```

To me this is like doing `Option(5).forall(_ == 7)`. Why not just do `5 == 7` ? Why do you need an `Option` ?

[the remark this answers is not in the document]
allisonport-db: If `newField.getMetadata.get(InvariantUtils.INVARIANTS_FIELD)` is `null` (because the invariant has been removed) 

```
// this is true
Option(newField.getMetadata.get(InvariantUtils.INVARIANTS_FIELD))
    .forall(_ == existingField.getMetadata.get(InvariantUtils.INVARIANTS_FIELD)))

 // this is false
newField.getMetadata.get(InvariantUtils.INVARIANTS_FIELD) ==
    existingField.getMetadata.get(InvariantUtils.INVARIANTS_FIELD)
```

I'm open to just blocking removing an invariant though.

standalone/src/test/scala/io/delta/standalone/internal/DeltaConstraintsSuite.scala
allisonport-db: we have added the public APIs for check constraints

standalone/src/main/scala/io/delta/standalone/internal/util/SchemaUtils.scala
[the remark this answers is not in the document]
scottsand-db: what is the relationship between `def isStructWriteCompatible(_existingSchema: StructType, _newSchema: StructType): Boolean = {` and `I'm open to just blocking removing an invariant though.`?

i.e. is this an arbitrary decision? does it impact users?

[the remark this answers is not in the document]
scottsand-db: is there a specific test case for this scenario?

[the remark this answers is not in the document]
allisonport-db: Sorry the overall test name isn't accurate. Test is here https://github.com/delta-io/connectors/pull/456/files#diff-8af5d9c3c7304b98030b6ac30ff9b492293a542f5dadc3e44a7b3aa4ba5284c9R366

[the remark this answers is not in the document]
allisonport-db: We can block any/all column-invariant schema changes for now, and can always unblock them in the future if needed.

We provide no public APIs for removing column invariants, but since they are technically deprecated, we might not want to explicitly block removing them from tables, but we can always change this later if it's necessary. Also removing them isn't really a protocol violation, where as adding or changing them could be since existing data would need to be checked.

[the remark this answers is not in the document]
scottsand-db: is this what delta-spark does?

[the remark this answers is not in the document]
scottsand-db: Feel free to schedule a sync with me so we can just hash this out quickly instead of delaying the review across days :p

scottsand-db (approved):

PR #453, opened 2022-10-21
[WIP][PROTOTYPE] Support column mapping in standalone

vkorukanti:
Prototype code to support column mapping in standalone.

PR #457, opened 2022-10-26
[ColumnMapping] Add utility methods to `SchemaUtils`

vkorukanti:
(This is first PR for column mapping support in Standalone. End-2-end prototype is at #453)

Add/Update following utility methods in `SchemaUtils`
* Make the `prettyFieldName` a public method. This will be used in column mapping metadata population (see PR #453 for end-2-end implementation).
* Add method to search for a (nested) field in a given struct.


scottsand-db (approved):
LGTM

PR #461, opened 2022-10-28
[ColumnMapping] Column mapping metadata verification and physical schema

vkorukanti:
(This is fourth PR for column mapping support in Standalone. End-2-end prototype is at #453)

This PR adds:
 * `checkColumnIdAndPhysicalNameAssignments`: It verifies given a schema, each field has the required column mapping metadata given column mapping mode and the column mapping metadata is unique to each field.
 * `createPhysicalSchema`: Given a schema with column mapping metadata, returns a schema that can be used to read data from or write data into the Parquet files depending upon the column mapping mode
   * NAME mode: returned physical contains only the physical name in the field metadata. This physical name is used to reference a field in the Parquet file
   * ID mode: returned physical schema contains `parquet.field.id` which is matched with field id in column metadata in Parquet file.

### DIFF
--- a/standalone/src/main/scala/io/delta/standalone/internal/DeltaColumnMapping.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/DeltaColumnMapping.scala
@@ -125,7 +125,10 @@ trait DeltaColumnMappingBase extends Logging {
     }
   }
 
-  /** Check the ids and physical names assigned to the columns are unique. */
+  /**
+   * Verify the ids and physical names assigned to the columns are unique and are valid.
+   * If not throw errors.
+   */
   def checkColumnIdAndPhysicalNameAssignments(
       schema: StructType,
       mode: DeltaColumnMappingMode): Unit = {
--- a/standalone/src/main/scala/io/delta/standalone/internal/DeltaColumnMapping.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/DeltaColumnMapping.scala
@@ -18,6 +18,8 @@ package io.delta.standalone.internal
 
 import java.util.{Locale, UUID}
 
+import scala.collection.mutable
+
 import io.delta.standalone.types.{FieldMetadata, StructField, StructType}
 
 import io.delta.standalone.internal.actions.{Metadata, Protocol}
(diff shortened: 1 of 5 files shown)


standalone/src/main/scala/io/delta/standalone/internal/DeltaColumnMapping.scala
```diff
@@ -104,6 +125,96 @@ trait DeltaColumnMappingBase extends Logging {
     }
   }
 
+  /** Check the ids and physical names assigned to the columns are unique. */
```
scottsand-db, line 128: can you clarify that "check" means "assert / throw error"

standalone/src/test/scala/io/delta/standalone/internal/DeltaColumnMappingSuite.scala
```diff
@@ -167,6 +168,141 @@ class DeltaColumnMappingSuite extends FunSuite {
...
+    val physicalSchemaId = createPhysicalSchema(
+      logicalSchema, referenceSchema, IdMapping, checkSupportedMode = false)
+
+    // Make sure every field has the parquet field id. This id is used to read from
+    // or write data into the parquet files.
+    transformColumns(physicalSchemaId)((path, field, _) => {
+      assert(field.getMetadata.contains(COLUMN_MAPPING_METADATA_ID_KEY))
+      assert(field.getMetadata.contains(PARQUET_FIELD_ID_METADATA_KEY))
+
+      // Make sure the parquet filed id is same as the column mapping id
```
scottsand-db, line 253: `field`

standalone/src/test/scala/io/delta/standalone/internal/DeltaColumnMappingSuite.scala
```diff
@@ -167,6 +168,141 @@ class DeltaColumnMappingSuite extends FunSuite {
...
+      assert(
+        field.getMetadata.get(PARQUET_FIELD_ID_METADATA_KEY) ===
+        field.getMetadata.get(PARQUET_FIELD_ID_METADATA_KEY))
```
scottsand-db, line 256: both of these are `PARQUET_FIELD_ID_METADATA_KEY`

scottsand-db (changes requested):
Left minor comments. After this LGTM.

standalone/src/main/scala/io/delta/standalone/internal/DeltaColumnMapping.scala
vkorukanti: Basically validating the column id metadata. Clarified in the comment. Kept the method name to keep it sync with the Delta on Spark.

standalone/src/test/scala/io/delta/standalone/internal/DeltaColumnMappingSuite.scala
vkorukanti: oops, good catch!

scottsand-db (approved):

PR #463, opened 2022-10-31
Remove outdated question in FAQ fixes delta-io/connectors/issues#462

### LINKED ISSUE #462: Flink Connector FAQ says that reading isn't supported even though it is
ref: https://github.com/noelo/connectors/blob/master/flink/README.md#frequently-asked-questions-faq

Can I use this connector to read data from a Delta table?

No, currently we are supporting only writing to a Delta table. A DeltaSource API with the support for reading data from Delta tables will be added in future releases.

But further up in the same README.md

Delta Source
Modes

Delta Source can work in one of two modes, described below.

The DeltaSource class provides factory methods to create sources for both modes. Please see [documentation](https://delta-io.github.io/connectors/latest/delta-flink/api/java/index.html) and examples for details.

noelo:
Resolves #462

### DIFF
--- a/flink/README.md
+++ b/flink/README.md
@@ -408,11 +408,6 @@ The project is compiled using [SBT](https://www.scala-sbt.org/1.x/docs/Command-L
 
 ## Frequently asked questions (FAQ)
 
-#### Can I use this connector to read data from a Delta table?
-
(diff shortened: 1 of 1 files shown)


scottsand-db:
Can you please add a better PR description? It's a trivial change, but we still like having clean PR descriptions :)

scottsand-db (approved):

PR #460, opened 2022-10-27
[ColumnMapping] Add column mapping metadata update functionality

vkorukanti:
(This is third PR for column mapping support in Standalone. End-2-end prototype is at #453)

* Add column mapping modes (`DeltaColumnMappingMode`)
* Add a method to verify and update column mapping metadata (`DeltaColumnMapping.verifyAndUpdateMetadataChange`)

PR #464, opened 2022-11-01
[ColumnMapping] Add support for setting column mapping mode through txn

vkorukanti:
(This is fifth PR for column mapping support in Standalone. End-2-end prototype is at #453)

It integrates the APIs added as part of (#461, #460) in `OptimisticTransaction.updateMetadata` API. Using `OptimisticTransaction.updateMetadata`, clients can update the column mapping mode in a Delta table.

### DIFF
--- a/standalone/src/main/java/io/delta/standalone/actions/Metadata.java
+++ b/standalone/src/main/java/io/delta/standalone/actions/Metadata.java
@@ -19,6 +19,7 @@ package io.delta.standalone.actions;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
@@ -27,9 +28,11 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 
 import io.delta.standalone.Constraint;
+import io.delta.standalone.exceptions.ColumnMappingUnsupportedException;
 import io.delta.standalone.types.StructType;
 
 import io.delta.standalone.internal.ConstraintImpl;
+import io.delta.standalone.internal.DeltaConfigs;
 import io.delta.standalone.internal.exception.DeltaErrors;
 
 /**
@@ -187,6 +190,25 @@ public final class Metadata implements Action {
         Map<String, String> newConfiguration = new HashMap(configuration);
         newConfiguration.remove(fullKey);
         return copyBuilder().configuration(newConfiguration).build();
+    }
+
+    /**
+     * Returns a copy of this {@link Metadata} instance with the given column mapping mode.
+     * Any validation of the given column mapping mode happens when this new metadata is
+     * applied on the table.
+     *
+     * @param mode Column mapping mode. One of the ("none", "id", "none")
+     * @return A copy of this metadata object with the given column mapping mode.
+     * @throws ColumnMappingUnsupportedException if the given mode is not supported.
+     */
+    public Metadata withColumnMappingMode(String mode) throws ColumnMappingUnsupportedException {
+        Map<String, String> newConfiguration = new HashMap(configuration);
+        mode = mode.toLowerCase(Locale.ROOT);
+        if ("name".equals(mode) || "id".equals(mode) || "none".equals(mode)) {
+            newConfiguration.put(DeltaConfigs.COLUMN_MAPPING_MODE().key(), mode);
+            return copyBuilder().configuration(newConfiguration).build();
+        }
+        throw DeltaErrors.unsupportedColumnMappingMode(mode);
     }
 
     @Override
--- a/standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
@@ -125,6 +125,9 @@ private[internal] object Protocol {
     // check config
 
     // Column mapping
+    // The check is actually done as part of `OptimisticTransactionImpl.updateMetadata`
+    // when updating the metadata for column mapping mode. Added the check here for
+    // completeness of all protocol checks in this method.
     metadata.columnMappingMode match {
       case NameMapping | IdMapping =>
         if (protocol.minWriterVersion < DeltaColumnMapping.MIN_WRITER_VERSION ||
--- a/standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
@@ -26,7 +26,7 @@ import com.fasterxml.jackson.databind.annotation.{JsonDeserialize, JsonSerialize
 
 import io.delta.standalone.types.StructType
 
-import io.delta.standalone.internal.{ConstraintImpl, DeltaColumnMappingMode, DeltaConfigs}
+import io.delta.standalone.internal.{ConstraintImpl, DeltaColumnMapping, DeltaColumnMappingMode, DeltaConfigs, IdMapping, NameMapping}
 import io.delta.standalone.internal.exception.DeltaErrors
 import io.delta.standalone.internal.util.{DataTypeParser, JsonUtils}
 
@@ -125,7 +125,14 @@ private[internal] object Protocol {
     // check config
 
     // Column mapping
-    // check column mapping mode
+    metadata.columnMappingMode match {
+      case NameMapping | IdMapping =>
+        if (protocol.minWriterVersion < DeltaColumnMapping.MIN_WRITER_VERSION ||
+          protocol.minReaderVersion < DeltaColumnMapping.MIN_READER_VERSION) {
+          throw DeltaErrors.changeColumnMappingModeOnOldProtocol(protocol)
+        }
+      case _ =>
+    }
 
     // todo: Should we check for any unsupported features? i.e. identity columns
   }
--- a/standalone/src/test/scala/io/delta/standalone/internal/DeltaColumnMappingSuite.scala
+++ b/standalone/src/test/scala/io/delta/standalone/internal/DeltaColumnMappingSuite.scala
@@ -17,16 +17,20 @@ package io.delta.standalone.internal
 
 import scala.collection.mutable
 
+import org.apache.hadoop.conf.Configuration
 import org.scalatest.FunSuite
 
+import io.delta.standalone.{Operation, OptimisticTransaction}
+import io.delta.standalone.actions.{Metadata => MetadataJ}
 import io.delta.standalone.exceptions.{ColumnMappingException, ColumnMappingUnsupportedException}
 import io.delta.standalone.types.{ArrayType, DataType, FieldMetadata, FloatType, IntegerType, LongType, MapType, StringType, StructField, StructType}
 import io.delta.standalone.types.FieldMetadata.builder
 
 import io.delta.standalone.internal.DeltaColumnMapping._
 import io.delta.standalone.internal.actions.{Metadata, Protocol}
-import io.delta.standalone.internal.util.{SchemaMergingUtils, SchemaUtils}
+import io.delta.standalone.internal.util.{ConversionUtils, SchemaMergingUtils}
 import io.delta.standalone.internal.util.SchemaMergingUtils.transformColumns
+import io.delta.standalone.internal.util.TestUtils._
 
 
 class DeltaColumnMappingSuite extends FunSuite {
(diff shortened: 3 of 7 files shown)


standalone/src/test/scala/io/delta/standalone/internal/DeltaColumnMappingSuite.scala
```diff
@@ -303,6 +307,44 @@ class DeltaColumnMappingSuite extends FunSuite {
...
 
+  test("Upgrade to column mapping mode 'name' on the table") {
+    withTempDir { dir =>
+      var log = DeltaLog.forTable(new Configuration(), dir.getCanonicalPath)
+      var txn = log.startTransaction()
+      val addFile = AddFile("/absolute/path/to/file/test.parquet", Map(), 0, 0, true)
+
+      // TODO: until a public API is available, change protocol directly on the txn impl
+      txn.asInstanceOf[OptimisticTransactionImpl]
+          .upgradeProtocolVersion(MIN_READER_VERSION, MIN_WRITER_VERSION)
```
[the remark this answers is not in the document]
allisonport-db: You need to instantiate the `DeltaLog` with [getDeltaLogWithMaxFeatureSupport](https://github.com/delta-io/connectors/blob/support_higher_protocol_versions_feature_branch/standalone/src/test/scala/io/delta/standalone/internal/util/TestUtils.scala#L65)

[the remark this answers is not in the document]
allisonport-db: This is good feedback, I'll add a todo to improve our error messages when we finish the protocol updates. It should be clear why we are failing:
- table version is too high for delta standalone to read/write (error that you are seeing right now)
- upgrading table to too high a version (committing too high a Protocol)
- etc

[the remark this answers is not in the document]
scottsand-db: @allisonport-db 

`getDeltaLogWithMaxFeatureSupport` doesn't seem like the corresponding ... util for `.upgradeProtocolVersion(MIN_READER_VERSION, MIN_WRITER_VERSION)` 

one is max ... the other is min

standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
```diff
@@ -106,7 +106,14 @@ private[internal] object Protocol {
     // check config
 
     // Column mapping
-    // check column mapping mode
+    metadata.columnMappingMode match {
+      case NameMapping | IdMapping =>
```
scottsand-db, line 110: are we supporting `Id` mapping mode right now?

standalone/src/test/scala/io/delta/standalone/internal/DeltaColumnMappingSuite.scala
```diff
@@ -303,6 +307,44 @@ class DeltaColumnMappingSuite extends FunSuite {
...
+      var log = getDeltaLogWithMaxFeatureSupport(new Configuration(), dir.getCanonicalPath)
+      var txn = log.startTransaction()
+      val addFile = AddFile("/absolute/path/to/file/test.parquet", Map(), 0, 0, true)
+
+      // TODO: until a public API is available, change protocol directly on the txn impl
+      txn.asInstanceOf[OptimisticTransactionImpl]
+          .upgradeProtocolVersion(MIN_READER_VERSION, MIN_WRITER_VERSION)
+
+      txn.updateMetadata(metadataJ(simpleSchema, withMode(Map.empty, "name")))
+      txn.commit( addFile :: Nil, OP, "test")
```
scottsand-db, line 321: we shouldn't be able to upgrade to Column Mapping AND commit add files in the same commit .....

that blocking doesn't need to be done in this PR but we shouldn't have a test like this.


you can just commit an empty list of actions

standalone/src/test/scala/io/delta/standalone/internal/DeltaColumnMappingSuite.scala
```diff
@@ -303,6 +307,44 @@ class DeltaColumnMappingSuite extends FunSuite {
...
 
+  test("Upgrade to column mapping mode 'name' on the table") {
+    withTempDir { dir =>
+      var log = getDeltaLogWithMaxFeatureSupport(new Configuration(), dir.getCanonicalPath)
+      var txn = log.startTransaction()
+      val addFile = AddFile("/absolute/path/to/file/test.parquet", Map(), 0, 0, true)
+
+      // TODO: until a public API is available, change protocol directly on the txn impl
+      txn.asInstanceOf[OptimisticTransactionImpl]
+          .upgradeProtocolVersion(MIN_READER_VERSION, MIN_WRITER_VERSION)
```
scottsand-db, line 318: isn't this a protocol downgrade?

```scala
var log = getDeltaLogWithMaxFeatureSupport(new Configuration(), dir.getCanonicalPath)
...
upgradeProtocolVersion(MIN_READER_VERSION, MIN_WRITER_VERSION)
```

how is this allowed?

standalone/src/test/scala/io/delta/standalone/internal/DeltaColumnMappingSuite.scala
```diff
@@ -303,6 +307,44 @@ class DeltaColumnMappingSuite extends FunSuite {
...
+
+      log = getDeltaLogWithMaxFeatureSupport(new Configuration(), dir.getCanonicalPath)
+      txn = log.startTransaction()
+
+      val oldMetadata = txn.metadata()
+      val newConfig = new java.util.HashMap[String, String]()
+      newConfig.putAll(oldMetadata.getConfiguration)
+
+      // Add column mapping related configs
+      newConfig.put("delta.columnMapping.mode", "name")
```
scottsand-db, line 331: What's the point of this? Didn't we already upgrade to column mapping on line 320?

```scala
txn.updateMetadata(metadataJ(simpleSchema, withMode(Map.empty, "name")))
```

standalone/src/test/scala/io/delta/standalone/internal/DeltaColumnMappingSuite.scala
vkorukanti: Its column mapping min required protocol. Given the max support "added" in `getDeltaLogWithMaxFeatureSupport` is same as the Delta column mapping protocol version, this went through. We need to revisit this once we have a proper API to upgrade protocol version

standalone/src/test/scala/io/delta/standalone/internal/DeltaColumnMappingSuite.scala
vkorukanti: I see. I took an example test from here: https://github.com/delta-io/connectors/blame/master/standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionSuite.scala#L308.

vkorukanti: I remove the addFile from this test.

standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
vkorukanti: There isn't much coded needed to support ID mode. I just need to add more tests. For now, making the changes wherever required to avoid forgetting changing them later.

standalone/src/test/scala/io/delta/standalone/internal/DeltaColumnMappingSuite.scala
scottsand-db: I see. so what's the point of calling `.upgradeProtocolVersion(MIN_READER_VERSION, MIN_WRITER_VERSION)` if we are already at that version due to `getDeltaLogWithMaxFeatureSupport` ?

allisonport-db: `getDeltaLogWithMaxFeatureSupport` sets up what protocol tables we are _allowed to read and write_ to (this does not write to the transaction log or have any effect on what we write, only whether we throw an error). `upgradeProtocolVersion` sets the _table's protocol version_ (commits a Protocol file-action)

scottsand-db: That clarifies it. Thanks

standalone/src/test/scala/io/delta/standalone/internal/DeltaColumnMappingSuite.scala
vkorukanti: We shouldn't have updated in line 320. Updated the test and added more.

standalone/src/test/scala/io/delta/standalone/internal/DeltaColumnMappingSuite.scala
```diff
@@ -303,6 +307,80 @@ class DeltaColumnMappingSuite extends FunSuite {
...
+      // TODO: until a public API is available, change protocol directly on the txn impl
+      txn.asInstanceOf[OptimisticTransactionImpl].upgradeProtocolVersion(
+        DeltaColumnMapping.MIN_READER_VERSION,
+        DeltaColumnMapping.MIN_WRITER_VERSION)
+
+      txn.updateMetadata(metadataJ(complexSchema, Map.empty))
+      txn.commit(Seq.empty, OP, "test")
+
+      txn = getTxnWithMaxFeatureSupport(dir.getCanonicalPath)
+      val oldMetadata = txn.metadata()
```
scottsand-db, line 322: @allisonport-db @vkorukanti how do we feel about using this API in practice? Seems like a lot of steps/work just to enable column mapping.

1. get and save current metadata `val oldMetadata = txn.metadata()`
2. get and save current configuration `newConfig.putAll(oldMetadata.getConfiguration)`
3.  update column mapping mode. (this is useful/what we want to do!) `newConfig.put("delta.columnMapping.mode", "name")`
4. create our new latest metadata `val newMetadata = oldMetadata.copyBuilder().configuration(newConfig).build()`
5. update txn metadata `txn.updateMetadata(newMetadata)`

standalone/src/test/scala/io/delta/standalone/internal/DeltaColumnMappingSuite.scala
```diff
@@ -303,6 +307,80 @@ class DeltaColumnMappingSuite extends FunSuite {
...
 
+  test("Upgrade to column mapping mode 'name' on the table") {
+    withTempDir { dir =>
+      var txn = getTxnWithMaxFeatureSupport(dir.getCanonicalPath)
+      // TODO: until a public API is available, change protocol directly on the txn impl
+      txn.asInstanceOf[OptimisticTransactionImpl].upgradeProtocolVersion(
+        DeltaColumnMapping.MIN_READER_VERSION,
+        DeltaColumnMapping.MIN_WRITER_VERSION)
+
+      txn.updateMetadata(metadataJ(complexSchema, Map.empty))
```
scottsand-db, line 318: nit: this test has a lot of ... tiny tests. could we add some comments to make this clear

```
// Step 1. Create table with required protocol version
// Step 2. Create metadata needed to perform column mapping upgrade
// Step 3. Update the table
// Step 4. Verify stuff
```

scottsand-db (approved):
LGTM. However, I'm curious if we want to revisit the API.

https://github.com/delta-io/connectors/pull/464/files#r1012338772

standalone/src/test/scala/io/delta/standalone/internal/DeltaColumnMappingSuite.scala
vkorukanti: Steps 1, 3, 5 can't be avoided, unless we provide a txn.updateColumnMappingMode.

Steps 2, 3 and 4 can be combined into a new API `oldMetadata.withNewColumnMappingMode(String newMode)`. 

@allisonport-db Let me know.

standalone/src/test/scala/io/delta/standalone/internal/DeltaColumnMappingSuite.scala
vkorukanti: added comments

standalone/src/test/scala/io/delta/standalone/internal/DeltaColumnMappingSuite.scala
allisonport-db: I think combining 2, 3, 4 is a good idea and I was thinking we need something like this for `withAppendOnlyMode(mode)` or something along those lines.

That would make the workflow something like this which looks a lot better to me
```
txn.updateMetadata(txn.metadata().withColumnMapppingMode("name"))
```

In general we don't want to add these APIs to `OptimisticTransaction`

vkorukanti: Added `txn.updateMetadata(txn.metadata().withColumnMapppingMode("name"))`

standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
```diff
@@ -125,7 +125,14 @@ private[internal] object Protocol {
     // check config
 
     // Column mapping
-    // check column mapping mode
+    metadata.columnMappingMode match {
```
allisonport-db, line 128: Can we add a test for this check? if this is in a different PR sorry you can resolve this

allisonport-db (commented):
one question about the protocol check

scottsand-db (approved):

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -189,6 +192,25 @@ public Metadata withoutCheckConstraint(String name) throws Throwable {
...
+     * applied on the table.
+     *
+     * @param mode Column mapping mode. One of the ("none", "id", "none")
+     * @return A copy of this metadata object with the given column mapping mode.
+     * @throws ColumnMappingUnsupportedException if the given mode is not supported.
+     */
+    public Metadata withColumnMappingMode(String mode) throws ColumnMappingUnsupportedException {
+        Map<String, String> newConfiguration = new HashMap(configuration);
+        mode = mode.toLowerCase(Locale.ROOT);
+        if ("name".equals(mode) || "id".equals(mode) || "none".equals(mode)) {
```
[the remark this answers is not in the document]
vkorukanti: I agree, I refactored this as part of the PR: https://github.com/delta-io/connectors/pull/466. Will push the changes once this PR is in.

standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
vkorukanti: We do have test for the protocol version check [here](https://github.com/delta-io/connectors/pull/464/files#diff-0262ed8a008c24ecf8b76c114fcb9e6587d2ba8cfd23cb89d8434afb570605bcR47).

One problem is, the check here is never hit. This check happens before this in `OptimisticTransactionImpl.updateMEtadata`. I added the check here for completeness of this method.

allisonport-db: Okay this is fine thanks for clarifying

allisonport-db: Can you just add a comment here saying the check is actually performed in `updateMetadata`?

allisonport-db (approved):
lgtm with the 1 clarifying comment for the protocol check

standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
vkorukanti: added a comment.

PR #465, opened 2022-11-02
 [ColumnMapping] Support `id` column mapping mode

vkorukanti:
(This is sixth PR for column mapping support in Standalone. End-2-end prototype is at #453)

This PR adds support for `id` column mapping mode. There is not much change in the production code as the given code already has support for both `name` and `id` column mapping mode. This change is mostly changing the error checks to consider `id` mode as valid and tests.

### DIFF
--- a/standalone/src/main/java/io/delta/standalone/actions/Metadata.java
+++ b/standalone/src/main/java/io/delta/standalone/actions/Metadata.java
@@ -211,6 +211,20 @@ public final class Metadata implements Action {
         throw DeltaErrors.unsupportedColumnMappingMode(mode);
     }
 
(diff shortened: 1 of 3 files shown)


scottsand-db (approved):

allisonport-db:
If we are supporting `id` and `name` we probably need a `getColumnMappingMode` API right?

vkorukanti:
> If we are supporting `id` and `name` we probably need a `getColumnMappingMode` API right?

Yep, added the API.

allisonport-db (approved):
lgtm. we can decide on string vs enum for the modes later on

PR #466, opened 2022-11-02
[ColumnMapping] Expose the physical schema through `StructType` and `StructField`

vkorukanti:
(This is seventh PR for column mapping support in Standalone. End-2-end prototype is at #453)

(This PR is stacked on top of #464 and #465, look only at the last commit for this PR specific changes).

Expose the physical schema details such as physical name and field id through `StructType` and `StructField`.

Tests are pending.

### DIFF
--- a/standalone/src/main/java/io/delta/standalone/actions/Metadata.java
+++ b/standalone/src/main/java/io/delta/standalone/actions/Metadata.java
@@ -27,6 +27,9 @@ import javax.annotation.Nullable;
 
 import io.delta.standalone.types.StructType;
 
+import io.delta.standalone.internal.DeltaColumnMapping;
+import io.delta.standalone.internal.DeltaColumnMappingMode;
+
 /**
  * Updates the metadata of the table. The first version of a table must contain
  * a {@link Metadata} action. Subsequent {@link Metadata} actions completely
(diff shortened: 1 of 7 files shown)


standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -133,6 +136,17 @@ public StructType getSchema() {
         return schema;
     }
 
+    /**
+     * @return the physical schema used to read or write data in table data files such as Parquet.
```
scottsand-db, line 140: nit: mention that if column mapping is disabled, this will just return the regular schema?

scottsand-db (approved):

standalone/src/main/java/io/delta/standalone/actions/Metadata.java
```diff
@@ -133,6 +136,17 @@ public StructType getSchema() {
...
+     */
+    public StructType getPhysicalSchema() {
```
allisonport-db, line 142: Do we need this? How do we expect a connector to use this exactly?

allisonport-db: Would this be used with the struct type to parquet converter we have?

PR #480, opened 2022-11-29
Flink connector e2e framework upgrade to Flink 1.15

rdziadosz:
Update the feature branch with the current changes from the master branch.
Due to the Flink version upgrade, there has been a change to kubernetes scripts and an upgrade of the libraries used by the end to end framework.


scottsand-db:
Instead of merging master into this branch .. wouldn't it be better to do a rebase? That way all of the code changes in master that are ahead of this branch don't show up as new changes in git?

PR #485, opened 2022-12-06
Update & fix flink integration tests dependencies

scottsand-db:
- upgrade flink version
- add missing necessary dependencies (that are marked as provided in the flink-delta connector dependency list)
- add a shell script to run Maven & SBT flink integration tests

### DIFF
--- a/examples/build.sbt
+++ b/examples/build.sbt
@@ -37,7 +37,7 @@ def getStandaloneVersion(): String = {
     println("Using Delta version " + version)
     version
   } else {
-    "0.5.0"
+    "0.6.0"
   }
 }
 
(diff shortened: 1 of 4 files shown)


kristoffSC:
You refer to flink-example as integration tests? Are there any test that do an actual validation/have assertions? 
I thought what we have in flink-example was simple Flink jobs to show how to use the connector, but those were never used for actual testing the system, sure they should work but they don't test automatically anything.

I was using "integration tests" term to describe an actual tests from flink/test (with IT in the name) that do run flink jobs and checks the results. 

Sorry but I think title and description is misleading, at least for me.

kristoffSC:
I have run `examples/run_flink_examples.sh` and it seems to do the job. 
Builds and run examples. I see job logs on the console.

Other than PR title and description, where it calls flink-example as integration test, which I dont thin they are, all look good to me :)

I would click "approve" but I cant since I'm not assigned as reviewer.

scottsand-db:
These are "integration tests" in that they run against an actual compiled delta-flink JAR.

The IT tests you are referring to are end-to-end tests ... but they just use the local code.

Testing against a real JAR lets us validate that our dependencies are correct and the JAR was generated properly.

PR #497, opened 2023-01-30
Respect writeStatsAsJson and writeStatsAsStruct during checkpoint

allisonport-db:
## Overview
Updates standalone to respect table properties `checkpoint.writeStatsAsJson` and `checkpoint.writeStatsAsStruct`. Also makes a few fixes uncovered from adding the test suite
- Don't write the "cdc" or "commitInfo" fields (delta-spark does [here](https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/Checkpoints.scala#L548))
- Don't write "remove.tags" (delta-spark does [here](https://github.com/delta-io/delta/blob/master/core/src/main/scala/org/apache/spark/sql/delta/Checkpoints.scala#L548))

Note: This PR includes writing the "add.partitionValues_parsed" field, but not "add.stats_parsed". This will be implemented in a later PR.

## Details

To omit fields from the checkpoint we use `SkippingParquetSchemaResolver`. Further detail on when we skip what fields are provided in the class doc for `Checkpoints.fieldsToSkip`.

To write the "add.partitionValues_parsed" field as a struct, we add a `partitionValues_parsed: ParsedPartitionValues ` field to the `AddFile` action. We introduce a custom encoder for `ParsedPartitionValues` that encodes the partition values as a `RowParquetRecord`.

## Testing

Adds a test suite `CheckpointSuite` and adds a test to `OSSCompatibilitySuite`.

I've also added a basic version of the `ActionSerializerSuite` copied from https://github.com/delta-io/delta/blob/master/core/src/test/scala/org/apache/spark/sql/delta/ActionSerializerSuite.scala to make sure we don't serialize the new addfile field. We can add the remaining tests to this suite in a follow-up PR.

### DIFF
--- a/standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
@@ -19,7 +19,7 @@ package io.delta.standalone.internal.actions
 import java.net.URI
 import java.sql.Timestamp
 
-import com.fasterxml.jackson.annotation.{JsonIgnore, JsonInclude, JsonRawValue}
+import com.fasterxml.jackson.annotation.{JsonIgnore, JsonIgnoreType, JsonInclude}
 import com.fasterxml.jackson.core.JsonGenerator
 import com.fasterxml.jackson.databind.{JsonSerializer, SerializerProvider}
 import com.fasterxml.jackson.databind.annotation.{JsonDeserialize, JsonSerialize}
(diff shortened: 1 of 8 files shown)


standalone/src/main/scala/io/delta/standalone/internal/actions/actions.scala
```diff
@@ -183,9 +194,9 @@ private[internal] case class AddFile(
     size: Long,
     modificationTime: Long,
     dataChange: Boolean,
-    @JsonRawValue
```
allisonport-db, line 186: If we don't remove this, we see an error when `stats != null`. According to the protocol we store `stats` as a string so we shouldn't have this anyways (and it's not in delta-io/delta)

standalone/src/test/scala/io/delta/standalone/internal/ActionSerializerSuite.scala
```diff
@@ -0,0 +1,154 @@
...
+
+package io.delta.standalone.internal
+
+import org.scalatest.FunSuite
+
+import io.delta.standalone.types.{IntegerType, StructType}
+
+import io.delta.standalone.internal.actions._
+
+class ActionSerializerSuite extends FunSuite {
```
allisonport-db, line 25: Copied from https://github.com/delta-io/delta/blob/master/core/src/test/scala/org/apache/spark/sql/delta/ActionSerializerSuite.scala (subset of those tests) except for "partitionValues_parsed is not serialized"

vkorukanti:

PR #501, opened 2023-02-08
Avoid null class loader in LogStoreProvider

snoe925:
Class.forName passed a null class loader will use the bootstrap loader and not the system class loader. This happens when using the Armeria server for example.
Signed-off-by: Shannon Noe <snoe925@gmail.com>

### DIFF
--- a/standalone/src/main/scala/io/delta/standalone/internal/storage/LogStoreProvider.scala
+++ b/standalone/src/main/scala/io/delta/standalone/internal/storage/LogStoreProvider.scala
@@ -55,8 +55,15 @@ private[internal] trait LogStoreProvider {
       new DelegatingLogStore(hadoopConf)
     } else {
       // scalastyle:off classforname
+      // Do not pass a null class loader
+      // - https://github.com/netty/netty/issues/7290
+      // - https://bugs.openjdk.java.net/browse/JDK-7008595
+      var loader = Thread.currentThread().getContextClassLoader
+      if (loader == null) {
+        loader = this.getClass().getClassLoader
+      }
       val logStoreClass =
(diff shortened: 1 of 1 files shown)


standalone/src/main/scala/io/delta/standalone/internal/storage/LogStoreProvider.scala
```diff
@@ -55,8 +55,15 @@ private[internal] trait LogStoreProvider {
...
+      // Do not pass a null class loader
+      // - https://github.com/netty/netty/issues/7290
+      // - https://bugs.openjdk.java.net/browse/JDK-7008595
+      var loader = Thread.currentThread().getContextClassLoader
+      if (loader == null) {
+        loader = this.getClass().getClassLoader
+      }
       val logStoreClass =
-        Class.forName(className, true, Thread.currentThread().getContextClassLoader)
+        Class.forName(className, true, loader)
```
scottsand-db, line 66: what about
```scala
// comments here
val classLoader = Option(Thread.currentThread().getContextClassLoader)
    .getOrElse(this.getClass().getClassLoader)
val logStoreClass = Class.forName(className, true, classLoader)
```
this is a little more scala-like

scottsand-db: but otherwise LGTM

snoe925: The outer comments about Java influenced me.  Certainly can be a one liner with Option().  😄

scottsand-db:
@snoe925 any reason you closed this? The change LGTM

PR #525, opened 2023-03-20
Avoid null LogStoreProvider class

snoe925:
The thread context class loader can be null.  When null use the class loader of the calling class.


snoe925:
The PR was old so I reopened a fresh one.

PR #527, opened 2023-03-20
Avoid null classloader

snoe925:
Netty likes to keep thread class loaders set to null.  If the thread class loader is null, then use the calling class.


snoe925:
I had to do this PR N times as I don't know how my work connection was destroying the code.

snoe925:
Can this PR merge?  Anything blocking it?

vkorukanti:

PR #532, opened 2023-04-10
[Delta-Flink Sink] Compute `txnVersion` lazily to reduce CPU utilization on commit

scottsand-db:
Previously, we were calling `deltaLog.startTransaction.txnVersion(appId)` every flink checkpoint (i.e. commit into the delta log). If a full log replay had not already been triggered, then `txnVesion` would trigger a full log replay - an expensive operation.

This is unnecessary as the result of `txnVersion` is only used during the first commit.

This change should reduce the CPU overhead each commit.

### DIFF
--- a/flink/src/main/java/io/delta/flink/internal/lang/Lazy.java
+++ b/flink/src/main/java/io/delta/flink/internal/lang/Lazy.java
@@ -29,11 +29,17 @@ public class Lazy<T> {
         this.supplier = supplier;
     }
 
-    public synchronized T get() {
+    public synchronized T threadSafeGet() {
         if (!instance.isPresent()) {
             instance = Optional.of(supplier.get());
         }
+        return instance.get();
+    }
 
+    public T threadUnsafeGet() {
+        if (!instance.isPresent()) {
+            instance = Optional.of(supplier.get());
+        }
         return instance.get();
     }
 }
--- a/flink/src/main/java/io/delta/flink/internal/lang/Lazy.java
+++ b/flink/src/main/java/io/delta/flink/internal/lang/Lazy.java
@@ -29,14 +29,8 @@ public class Lazy<T> {
         this.supplier = supplier;
     }
 
-    public synchronized T threadSafeGet() {
-        if (!instance.isPresent()) {
-            instance = Optional.of(supplier.get());
-        }
-        return instance.get();
-    }
-
-    public T threadUnsafeGet() {
+    /** Not thread safe. */
+    public T get() {
         if (!instance.isPresent()) {
             instance = Optional.of(supplier.get());
         }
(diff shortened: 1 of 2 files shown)


flink/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
```diff
@@ -271,16 +272,20 @@ private SortedMap<Long, List<CheckpointData>> getCommittablesPerCheckpoint(
             List<DeltaGlobalCommittable> globalCommittables,
             DeltaLog deltaLog) {
 
-        OptimisticTransaction transaction = deltaLog.startTransaction();
-        long tableVersion = transaction.txnVersion(appId);
+        // The last committed table version by THIS flink application.
+        // Even though `Lazy` is thread-unsafe, we know that the GlobalCommitter runs with
+        // parallelism equal to 1, and we are only accessing this local variable in this single
```
kristoffSC, line 277: Even if Flink would allow for more than one Global Committer (parallelism > 1) this code would be executed only by one thread.

So this is "safe" not because `parallelism equal to 1` but because Flink's threading model.

flink/src/main/java/io/delta/flink/internal/lang/Lazy.java
```diff
@@ -0,0 +1,40 @@
...
+public class Lazy<T> {
+    private final Supplier<T> supplier;
+    private Optional<T> instance = Optional.empty();
+
+    public Lazy(Supplier<T> supplier) {
+        this.supplier = supplier;
+    }
+
+    /** Not thread-safe (i.e. may perform duplicate `supplier.get()` calls) */
+    public T get() {
```
vkorukanti, line 33: we could add synchronized keyword to the method and get rid of thread-unsafe comments here and below.

flink/src/main/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitter.java
```diff
@@ -271,16 +272,20 @@ private SortedMap<Long, List<CheckpointData>> getCommittablesPerCheckpoint(
...
+        // method.
+        Lazy<Long> lastCommittedTableVersion =
+            new Lazy<>(() -> deltaLog.startTransaction().txnVersion(appId));
 
-        if (!this.firstCommit || tableVersion < 0) {
+        if (!this.firstCommit || lastCommittedTableVersion.get() < 0) {
```
vkorukanti, line 282: Add a comment saying that keep the `lastCommittedTableVersion.get()` as last call, given it is expensive.

vkorukanti (approved):
lgtm

flink/src/main/java/io/delta/flink/internal/lang/Lazy.java
kristoffSC: > we could add synchronized keyword

I would actually advice to not do that. This code will never be called from more than one thread due to Flink's threading model. Its a common practice that whenever you wrote Flink's processing functions etc you tend not to use synchronization or any thread blocking things unless your logic executes your business logic on multiple threads.

If this class would be use in the future on a HOT PATH, than `synchronized` would kill its performance.

PR #533, opened 2023-04-10
[Delta-Standalone] Compute Snapshot protocol and metadata more efficiently

scottsand-db:
## The problem

`loadProtocolAndMetadata` can be an expensive operation when initializing a snapshot.

image

## The solution

This PR passes in a `previousSnapshotHint` into the `Snapshot` constructor, to be used during reverse log replay (when we calculate the latest protocol and metadata for the table).

At a high level, here is the logic that this PR implements:

Suppose the latest version of the table is V24 and we perform our first snapshot creation, i.e. `DeltaLog.forTable(...).snapshot()`. We will, as usual, scan files `0024.json`, `0023.json`, `0022.json`, `0021.json`, until we get to `0020.checkpoint.parquet` where we will find the protocol and metadata. (assuming no recent protocol and metadata updates in the json files).

Suppose some time later we call `log.update()`, and that versions 25-29 have been added to the table.

This PR changes the reverse replay logic so that we ONLY read files 29-25 to look for new protocol or metadata updates.

If we don't find any, then we just use the already-saved protocol and metadata values from the current log's snapshot at V25. That is, we do not read versions 24-20.

Thus, this PR reduces the amount of file scanning and processing we have to do during snapshot initialization.

image

### DIFF
--- a/standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
+++ b/standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
@@ -31,7 +31,7 @@ import org.apache.hadoop.fs.Path
 import org.scalatest.FunSuite
 
 import io.delta.standalone.{DeltaLog, Operation, Snapshot}
-import io.delta.standalone.actions.{AddFile => AddFileJ, JobInfo => JobInfoJ, Metadata => MetadataJ, NotebookInfo => NotebookInfoJ, RemoveFile => RemoveFileJ}
+import io.delta.standalone.actions.{AddFile => AddFileJ, JobInfo => JobInfoJ, Metadata => MetadataJ, NotebookInfo => NotebookInfoJ, Protocol => ProtocolJ, RemoveFile => RemoveFileJ}
 import io.delta.standalone.exceptions.DeltaStandaloneException
 import io.delta.standalone.types.{BooleanType, IntegerType, LongType, StringType, StructType}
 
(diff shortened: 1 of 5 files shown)


standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
```diff
@@ -678,6 +678,95 @@ abstract class DeltaLogSuiteBase extends FunSuite {
...
+      val protocolV0 = new ProtocolJ(1, 2)
+      writerLog.startTransaction()
+        .commit((metadataV0 :: protocolV0 :: Nil).asJava, manualUpdate, engineInfo)
+
+      // V1-V8: The latest protocol and metadata is still in V0
+      for (i <- 1 to 8) { commit(i, writerLog) }
+
+      val readerLog = DeltaLog.forTable(conf, dir.getCanonicalPath)
+      val metrics1 = readerLog.snapshot().asInstanceOf[SnapshotImpl].protocolMetadataLoadMetrics
+      assert(metrics1.fileVersions.toList.sorted === (0 to 8).toList)
```
vkorukanti, line 707: do you want to verify that we have the correct metadata and protocol? Here and below?

standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
```diff
@@ -678,6 +678,95 @@ abstract class DeltaLogSuiteBase extends FunSuite {
...
+      //       delta-standalone protocol support.
+      val protocolV28 = new ProtocolJ(1, 2)
+      writerLog.startTransaction()
+        .commit((metadataV28 :: protocolV28 :: Nil).asJava, manualUpdate, engineInfo)
+      commit(29, writerLog)
+
+      // The current snapshot is still at V26. We should read V29, V28, see the newest protocol and
+      // metadata, and stop early
+      val metrics6 = readerLog.update().asInstanceOf[SnapshotImpl].protocolMetadataLoadMetrics
+      assert(metrics6.fileVersions.toList.sorted === (28 to 29).toList)
```
vkorukanti, line 767: If there aren't any existing tests, please also add the case where the DeltaLog is created from scratch and the scanning happens as exepected.

vkorukanti (approved):
The changes lgtm (% couple of comments about tests).

Do you have the full flamegraph? Its not clear from the attached one how load P&M method is taking time compared to the other operations.

standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
scottsand-db: Yup, that case is already covered. I'll add better test documentation to make this clearer.

standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
scottsand-db: In this specific case, there's only one possible metadata and protocol.
For cases where there are different possible answers (though, only one correct answer) I have asserted that we have the correct answer.

scottsand-db:
@vkorukanti to me, the flame graph shows that it is this load operation that is taking up the most CPU (width).

Don't think I can upload the full flame graph without the user's consent.

image

kristoffSC:
Hi @vkorukanti 
is this what you are asking for? 
https://user-images.githubusercontent.com/7932805/223778596-6b0b74e1-c636-480a-a35c-db5f8571daf5.png

This is a flame graph that I've ceated for my Demo Flink job that shows exact same problem.
This was for issue https://github.com/delta-io/connectors/pull/515. In that ticket I've identified 4 items to improve. Ticket 515 fixed first two items and now @scottsand-db fixed item 4 in this ticket.

PR #545, opened 2023-05-08
Delta/Flink Sink and Delta Standalone - allow disabling of delta checkpointing

scottsand-db:
There may be some cases when the delta checkpointing (every 10th delta commit, where we write the latest active files and un-expired tombstones into a parquet file(s)) is a bottleneck for applications.

e.g. One scenario might be that a user wants to ingest using the Flink/Delta Sink, and use Delta Lake + Apache Spark to create the delta checkpoints in a distributed manner, thus not bottlenecking the flink application.

This PR enables setting the hadoopConf flag (used when you instantiate delta-standalone and the delta-flink sink) `io.delta.standalone.checkpointing.enabled` to `false` to disable checkpointing.

### DIFF
--- a/flink/src/test/java/io/delta/flink/sink/DeltaSinkStreamingExecutionITCase.java
+++ b/flink/src/test/java/io/delta/flink/sink/DeltaSinkStreamingExecutionITCase.java
@@ -220,6 +220,31 @@ public class DeltaSinkStreamingExecutionITCase extends DeltaSinkExecutionITCaseB
         );
     }
 
+    @Test
+    void can_disable_delta_checkpointing() throws Exception {
+        final org.apache.hadoop.conf.Configuration hadoopConf =
+            new org.apache.hadoop.conf.Configuration();
+        hadoopConf.set("io.delta.standalone.checkpointing.enabled", "false");
+
+        DeltaTestUtils.initTestForNonPartitionedTable(deltaTablePath);
+
+        StreamExecutionEnvironment env = getTestStreamEnv(false); // no failover
+        env.addSource(new CheckpointCountingSource(1_000, 12))
+            .setParallelism(1)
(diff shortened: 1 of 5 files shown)


flink/src/test/java/io/delta/flink/sink/DeltaSinkStreamingExecutionITCase.java
```diff
@@ -220,6 +220,31 @@ public void testFileSinkWithGlobalCommitterFailover(
         );
     }
 
+    @Test
+    void can_disable_delta_checkpointing() throws Exception {
```
vkorukanti, line 224: the test doesn't need to be public?

Also change the naming to camel case?

flink/src/test/java/io/delta/flink/sink/DeltaSinkStreamingExecutionITCase.java
```diff
@@ -220,6 +220,31 @@ public void testFileSinkWithGlobalCommitterFailover(
...
+        final org.apache.hadoop.conf.Configuration hadoopConf =
+            new org.apache.hadoop.conf.Configuration();
+        hadoopConf.set("io.delta.standalone.checkpointing.enabled", "false");
```
vkorukanti, line 227: can use `StandaloneHadoopConf.CHECKPOINTING_ENABLED` or not available in Flink module?

vkorukanti (approved):
lgtm.

flink/src/test/java/io/delta/flink/sink/DeltaSinkStreamingExecutionITCase.java
scottsand-db: `StandaloneHadoopConf` is an internal scala object, can't use it here

PR #546, opened 2023-05-09
Block adding a new column with nullable=false for existing tables

allisonport-db:
### Changes proposed
For a non-empty table we should block adding a new non-null column (when the current commit does NOT remove all existing data).

Adding a non-null column corrupts the table since any existing rows will have null values for the new column.

### Testing

Updates test in delta standalone and adds a test for the flink sink.

### DIFF
--- a/flink/src/test/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitterTest.java
+++ b/flink/src/test/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitterTest.java
@@ -199,19 +199,33 @@ public class DeltaGlobalCommitterTest {
     @Test
     public void testMergeIncompatibleSchema() throws Exception {
         //GIVEN
-        DeltaTestUtils.initTestForPartitionedTable(tablePath.getPath());
+        DeltaTestUtils.initTestForNonPartitionedTable(tablePath.getPath());
         List<DeltaGlobalCommittable> globalCommittables =
             DeltaSinkTestUtils.getListOfDeltaGlobalCommittables(
-                3, DeltaSinkTestUtils.getTestPartitionSpec());
+                3, new LinkedHashMap<>());
+
 
         // new schema drops one of the previous columns
-        RowType updatedSchema =
-            DeltaSinkTestUtils.dropOneColumnFromSchema(DeltaSinkTestUtils.TEST_ROW_TYPE);
+        RowType updatedSchema1 = DeltaSinkTestUtils
+                .dropOneColumnFromSchema(DeltaSinkTestUtils.TEST_ROW_TYPE);
+        // new schema adds a non-null column
+        RowType updatedSchema2 = DeltaSinkTestUtils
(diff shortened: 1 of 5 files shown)


flink/src/test/java/io/delta/flink/sink/internal/committer/DeltaGlobalCommitterTest.java
```diff
@@ -199,19 +199,33 @@ public void testMergeSchemaSetToFalse() throws Exception {
     @Test
     public void testMergeIncompatibleSchema() throws Exception {
```
allisonport-db, line 200: This test was broken already for a few different reasons and it wasn't caught because of the very general `assertThrows(RuntimeException ...)`. I've fixed this test and added error message checking but only for this test.
- Used `initTestForPartitionedTable` with `TEST_ROW_TYPE` (wrong schema for that init table).
- Needed to switch to the non-partitioned table because `.dropOneColumnFromSchema` will remove the last column. This was one of the partition columns in `TEST_PARTITIONED_ROW_TYPE` and we would instead catch a `RuntimeException(Partition column(s) col1,col2 not found in schema:...)`
- Was not instantiating a GlobalCommiter with `mergeSchema=true`. We would instead be catching `RuntimeException(DataStream's schema is different from current table's schema..)` assuming the rest of the test was correct, instead of the Standalone's exception for incompatible schema

standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionSuite.scala
```diff
@@ -252,6 +248,10 @@ class OptimisticTransactionSuite extends OptimisticTransactionSuiteBase {
     // change of datatype
     val schema4 = new StructType(Array(new StructField("a", new StringType(), true)))
     testSchemaChange(schema2, schema4, shouldThrow = true)
+
+    // add non-nullable field
+    val schema5 = schema1.add(new StructField("c", new IntegerType(), false))
```
scottsand-db, line 253: nit: why does this have to be `schema5` instead of the previous `schema3`?

scottsand-db (approved):
LGTM

standalone/src/test/scala/io/delta/standalone/internal/OptimisticTransactionSuite.scala
allisonport-db: I mean it can be schema3 if I move it up and bump all the other schema numbers in this test by 1... (this is a different test there are already schema 1-4)

scottsand-db: OH the github diff made me think that this was the same method as the above method, where you deleted `schema3`

So, nvm :)

allisonport-db: Cool merging

PR #553, opened 2023-05-25
[WIP] [Standalone] Load the app transaction versions using memory-efficient log replay

tdas:
TODO

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -472,6 +472,7 @@ lazy val standalone = (project in file("standalone"))
       "com.github.mjakubowski84" %% "parquet4s-core" % parquet4sVersion excludeAll (
         ExclusionRule("org.slf4j", "slf4j-api")
       ),
+      "org.xerial.snappy" % "snappy-java" % "1.1.8.4",
       "com.fasterxml.jackson.module" %% "jackson-module-scala" % "2.12.3",
       "org.json4s" %% "json4s-jackson" % "3.7.0-M11" excludeAll (
         ExclusionRule("com.fasterxml.jackson.core"),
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -472,6 +472,7 @@ lazy val standalone = (project in file("standalone"))
       "com.github.mjakubowski84" %% "parquet4s-core" % parquet4sVersion excludeAll (
         ExclusionRule("org.slf4j", "slf4j-api")
       ),
+      "org.xerial.snappy" % "snappy-java" % "1.1.8.4",
```
tdas, line 475: needed for running tests on mac M1/M2

standalone/src/main/scala/io/delta/standalone/internal/SnapshotImpl.scala
```diff
@@ -216,18 +226,11 @@ private[internal] class SnapshotImpl(
...
-            }
           case m: Metadata if null == metadata =>
-            metadata = m
-
-            if (protocol != null && metadata != null) {
-              // Stop since we have found the latest Protocol and metadata.
-              return (protocol, metadata, createMetrics)
-            }
+            // We only need the latest metadata
+          metadata = m
```
tdas, line 231: this reduces the scope of quitting early.. but its always very likely that the metadata and protocol is in the checkpoint which will be loaded last.

standalone/src/test/scala/io/delta/standalone/internal/DeltaLogSuite.scala
```diff
@@ -770,9 +770,10 @@ abstract class DeltaLogSuiteBase extends FunSuite {
       commit(29, writerLog)
       // The current snapshot is still at V26. We should read V29, V28, see the newest protocol and
       // metadata, and stop early
+      // update this
```
tdas, line 773: TODO

